PR #53, opened 2024-09-19
chore: clean up PromotionStrategy reconcile function add add comments

zachaller:
Just pull create pull request into it's own function and add comments on what code is doing.

### DIFF
--- a/internal/controller/promotionstrategy_controller.go
+++ b/internal/controller/promotionstrategy_controller.go
@@ -75,7 +75,7 @@ func (r *PromotionStrategyReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 
-	// If a ProposedCommit does not exist, create it othwise get it and store the ProposedCommit in a map with the branch as the key.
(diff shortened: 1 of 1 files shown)


internal/controller/promotionstrategy_controller.go
```diff
@@ -477,3 +406,98 @@ func (r *PromotionStrategyReconciler) copyCommitStatuses(ctx context.Context, cs
 
 	return nil
 }
+
+func (r *PromotionStrategyReconciler) tryToMergePullRequest(ctx context.Context, ps *promoterv1alpha1.PromotionStrategy, proposedCommitMap map[string]*promoterv1alpha1.ProposedCommit) error {
```
leoluz: Why prefix with `try` and not just `mergePullRequest`? The caller should be the one with the responsibility to know how to react on errors.

leoluz (approved):
LGTM

internal/controller/promotionstrategy_controller.go
zachaller: Yea the function is not actually retrying, the function name was in regards to trying to describe the action of trying to merge pull requests if the commit status checks where green.

PR #60, opened 2024-10-02
chore: more tests and fixes

zachaller:


Makefile
```diff
@@ -72,6 +72,14 @@ test: manifests generate fmt vet envtest ## Run tests.
 test-e2e:
 	go test ./test/e2e/ -v
 
+.PHONY: test-parallel
+test-parallel: ginkgo manifests generate fmt vet envtest ## Run tests in parallel
```
crenshaw-dev: Should `ginkgo manifests generate fmt vet envtest` be pulled into its own target?

If so, is it really necessary to have it before every test run? Or could we just document that `x-new-target` needs to be run before tests if there are "x, y, or z types of code changes"?

internal/controller/commitstatus_controller.go
```diff
@@ -65,36 +66,56 @@ func (r *CommitStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		logger.Info("No need to reconcile")
 		return ctrl.Result{}, nil
 	}
 
-	commitStatusProvider, err := r.getCommitStatusProvider(ctx, cs)
-	if err != nil {
-		return ctrl.Result{}, err
+	if cs.Spec.Sha == "" || cs.Spec.Phase == "" {
+		logger.Info("Skip setting commit status, missing sha or phase")
+		return ctrl.Result{}, nil
```
crenshaw-dev, line 87: Do we already have apply-time validation for these?

internal/controller/commitstatus_controller.go
```diff
@@ -65,36 +66,56 @@ func (r *CommitStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		if err != nil {
+			return err
+		}
+
+		_, err = commitStatusProvider.Set(ctx, &cs)
+		if err != nil {
+			return err
+		}
+
+		newCs.Status.ObservedGeneration = newCs.Generation
```
crenshaw-dev: Not sure I follow why `newCs` exists, it doesn't look like we're actually using it.

internal/controller/promotionstrategy_controller.go
```diff
@@ -133,60 +133,44 @@ func (r *PromotionStrategyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	pc := promoterv1alpha1.ProposedCommit{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            pcName,
+			Namespace:       ps.Namespace,
+			OwnerReferences: []metav1.OwnerReference{*controllerRef},
+			Labels: map[string]string{
+				"promoter.argoproj.io/promotion-strategy": utils.KubeSafeLabel(ctx, ps.Name),
+				"promoter.argoproj.io/proposed-commit":    utils.KubeSafeLabel(ctx, pcName),
+				"promoter.argoproj.io/environment":        utils.KubeSafeLabel(ctx, environment.Branch),
```
crenshaw-dev: Wanna go ahead and pull these label keys into constants?

internal/controller/promotionstrategy_controller.go
```diff
@@ -133,60 +133,44 @@ func (r *PromotionStrategyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Spec: promoterv1alpha1.ProposedCommitSpec{
+			RepositoryReference:    ps.Spec.RepositoryReference,
+			ProposedBranch:         fmt.Sprintf("%s-%s", environment.Branch, "next"),
+			ActiveBranch:           environment.Branch,
+			ActiveCommitStatuses:   append(environment.ActiveCommitStatuses, ps.Spec.ActiveCommitStatuses...),
+			ProposedCommitStatuses: append(environment.ProposedCommitStatuses, ps.Spec.ProposedCommitStatuses...),
+		},
+	}
 
+	err := r.Patch(ctx, &pc, client.MergeFrom(&promoterv1alpha1.ProposedCommit{}))
```
crenshaw-dev, line 162: Why patch rather than update? Since we own the resource, update seems safer.

internal/controller/promotionstrategy_controller.go
```diff
@@ -274,6 +260,7 @@ func (r *PromotionStrategyReconciler) calculateStatus(ctx context.Context, ps *p
 			//logger.Info("No proposed commit statuses configured, assuming success", "branch", environment.Branch)
 		} else {
 			ps.Status.Environments[i].Proposed.CommitStatus.Phase = string(promoterv1alpha1.CommitPhasePending)
+			ps.Status.Environments[i].Proposed.CommitStatus.Sha = pcMap[environment.Branch].Status.Proposed.Hydrated.Sha
```
crenshaw-dev: This loop is getting pretty gnarly. Can we break out some functions?

internal/controller/promotionstrategy_controller.go
```diff
@@ -303,62 +290,44 @@ func (r *PromotionStrategyReconciler) copyCommitStatuses(ctx context.Context, cs
...
 
-			if cs.Labels == nil {
-				cs.Labels = make(map[string]string)
+			copiedCommitStatus = &promoterv1alpha1.CommitStatus{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        proposedCSObjectKey.Name,
+					Annotations: commitStatus.Annotations,
+					Labels:      commitStatus.Labels,
+					Namespace:   commitStatus.Namespace,
+				},
```
crenshaw-dev, line 332: Should we own this CS?

internal/controller/promotionstrategy_controller.go
```diff
@@ -303,62 +290,44 @@ func (r *PromotionStrategyReconciler) copyCommitStatuses(ctx context.Context, cs
...
-			cs.Spec.Sha = copyToProposedHydratedSha
-			cs.Spec.Name = branch + " - " + commitStatus.Spec.Name
-
-			//TODO: This should be a function that gets the URL from the provider in a the providers required format
-			cs.Spec.Url = "https://github.com/" + cs.Spec.RepositoryReference.Owner + "/" + cs.Spec.RepositoryReference.Name + "/commit/" + copyFromActiveHydratedSha
-			err = r.Update(ctx, &cs)
+			copiedCommitStatus.Labels[promoterv1alpha1.CommitStatusLabelCopy] = "true"
+			copiedCommitStatus.Labels["promoter.argoproj.io/commit-status-copy-from"] = utils.KubeSafeLabel(ctx, commitStatus.Spec.Name)
+			copiedCommitStatus.Labels["promoter.argoproj.io/commit-status-copy-from-sha"] = utils.KubeSafeLabel(ctx, copyFromActiveHydratedSha)
+			copiedCommitStatus.Labels["promoter.argoproj.io/commit-status-copy-from-branch"] = utils.KubeSafeLabel(ctx, branch)
```
crenshaw-dev: Probably should go ahead and make constants for these labels.

internal/controller/promotionstrategy_controller.go
```diff
@@ -303,62 +290,44 @@ func (r *PromotionStrategyReconciler) copyCommitStatuses(ctx context.Context, cs
...
+
+			err = r.Patch(ctx, copiedCommitStatus, client.MergeFrom(&promoterv1alpha1.CommitStatus{}))
```
crenshaw-dev, line 350: Ditto from above, patch rather than Update?

internal/controller/promotionstrategy_controller.go
```diff
@@ -382,7 +351,7 @@ func (r *PromotionStrategyReconciler) mergePullRequests(ctx context.Context, ps
 
 		if previousEnvironmentStatus != nil {
 			// There is no previous environment to compare to, so we can't copy the commit statuses.
```
crenshaw-dev: This comment seems backwards given the line above

crenshaw-dev (changes requested):

internal/controller/commitstatus_controller.go
zachaller: So phase defaults to pending which is find so it should not ever be an empty string because it is also an enum.

Sha was a bit trickier because it is required but defaults to empty string idk enough about what end users would do when creating a commits status. I actually at one point figured out to regex it to a sha but then I figured if people might not know the shaw at creation. So I just went with this. Could probably remove Phase check it should be impossible.

internal/controller/commitstatus_controller.go
zachaller: newCs gets updated from the Get request

internal/controller/promotionstrategy_controller.go
zachaller: Yea I planned on it, can do it now

internal/controller/promotionstrategy_controller.go
zachaller: To avoid conflicts with the proposed commit controllers status updates etc

internal/controller/promotionstrategy_controller.go
zachaller: I don't think so, it would be weird in ArgoCD to have this copied one show up but not the parent.

internal/controller/promotionstrategy_controller.go
zachaller: Same logic we own this so we can patch to avoid conflicts with commitstatus controller.

internal/controller/commitstatus_controller.go
crenshaw-dev: No worries, honestly a comment saying `empty phase should be impossible, checking anyway` might be fine.

internal/controller/commitstatus_controller.go
crenshaw-dev: Right, but then as far as I can tell it never gets used beyond this field mutation.

internal/controller/promotionstrategy_controller.go
crenshaw-dev: Maybe the parent is the owner? Haven't thought it through, but maybe.

crenshaw-dev: Or maybe the ProposedCommit is the owner, same as the parent CommitStatus?

internal/controller/commitstatus_controller.go
zachaller: Your right that should be used in the update

internal/controller/promotionstrategy_controller.go
zachaller: Yea the CommitStatus maybe makes sense, idk depends on if we want to expose that detail to end users that use ArgoCD and have commit statuses checked in 🤷

crenshaw-dev: Fine to leave it for later

crenshaw-dev (approved):

PR #120, opened 2024-11-15
chore: wrap controller errors

crenshaw-dev:
Also did small refactors and added TODOs where I thought something might need to be done.


internal/controller/changetransferpolicy_controller.go
```diff
@@ -360,20 +366,21 @@ func (r *ChangeTransferPolicyReconciler) creatOrUpdatePullRequest(ctx context.Co
...
+			return fmt.Errorf("failed to get PR %q: %w", prName, err)
 		}
 	} else {
-		// Pull request already exists, update it.
-		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		// Pull Request already exists, update it.
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			prUpdated := promoterv1alpha1.PullRequest{}
-			err := r.Get(ctx, client.ObjectKey{Namespace: pr.Namespace, Name: pr.Name}, &prUpdated)
+			// TODO: consider skipping this Get on the first attempt, the object we already got might be up to date.
```
zachaller, line 380: Not sure the extra code complexity to do this is worth the cost of hitting the cache again

internal/controller/commitstatus_controller.go
```diff
@@ -89,22 +90,24 @@ func (r *CommitStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 		newCs := promoterv1alpha1.CommitStatus{}
-		err := r.Client.Get(ctx, req.NamespacedName, &newCs, &client.GetOptions{})
+		// TODO: consider skipping Get on the initial attempt. The object we already got might be up to date.
+		err = r.Client.Get(ctx, req.NamespacedName, &newCs, &client.GetOptions{})
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to get CommitStatus %q: %w", req.Name, err)
 		}
 
+		// TODO: consider pulling this outside the retry loop and instead using the reconcile requeue to handle SCM errors.
```
zachaller, line 107: I did this in the loop so because commit status are generally modified by other controllers and the status updates conflict so much that I wanted to not have to re-reconcile all the time. I originally had it outside this loop and it just ended up making noisy logs etc.

zachaller: This is one of those maybe document reasons in code things

internal/controller/pullrequest_controller.go
```diff
@@ -139,29 +139,32 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		// TODO: is there more work avoidance we can do here? Just because the generation has changed, do we know that
+		//       we need to hit the SCM? For example, an annotation update probably shouldn't cause an API call.
+		err = r.updatePullRequest(ctx, pr, pullRequestProvider)
 		if err != nil {
-			return ctrl.Result{}, err
+			return ctrl.Result{}, fmt.Errorf("failed to update pull request: %w", err)
 		}
 	}
 
+	// TODO: can we just skip this status update if the generation hasn't changed?
```
zachaller, line 163: We do there is code above that will return if there is no generation change

zachaller (approved):

internal/controller/changetransferpolicy_controller.go
crenshaw-dev: Yeah and are we even hitting the API or just the informer cache?

PR #127, opened 2024-11-20
fix: flakey tests

zachaller:


internal/controller/changetransferpolicy_controller_test.go
```diff
@@ -96,6 +100,23 @@ var _ = Describe("ChangeTransferPolicy Controller", func() {
 				g.Expect(pr.Status.State).To(Equal(promoterv1alpha1.PullRequestOpen))
 				g.Expect(pr.Name).To(Equal(utils.KubeSafeUniqueName(ctx, prName)))
 			}, EventuallyTimeout).Should(Succeed())
+
+			Eventually(func(g Gomega) {
+				err = k8sClient.Get(ctx, typeNamespacedName, changeTransferPolicy)
+				Expect(err).To(Succeed())
+				changeTransferPolicy.Spec.AutoMerge = ptr.To(true)
```
crenshaw-dev, line 109: Setting here just keeps the PR from being cleaned up before we enter this block? If so, probably want a comment to explain that.

crenshaw-dev (changes requested):

crenshaw-dev (approved):

Issue #113, opened 2024-11-15
feat: webhook trigger reconcile


zachaller:
We need to reconcile the CTP and possibly the PromotionStrategy on push from a webhook configured via the application. In order to do this we need to create a webhook receiver service that modifies an annotation on the respective resources.

PR #130, opened 2024-11-21
feat: webhook support and no timed reconcile

zachaller:
fixes: https://github.com/argoproj-labs/gitops-promoter/issues/113


internal/utils/utils.go
```diff
@@ -207,3 +207,23 @@ func UpsertEnvironmentStatus(slice []promoterv1alpha1.EnvironmentStatus, i promo
...
 }
+
+func UpsertChangeTransferPolicyList(slice []promoterv1alpha1.ChangeTransferPolicy, insertList []promoterv1alpha1.ChangeTransferPolicy) []promoterv1alpha1.ChangeTransferPolicy {
+	for _, i := range insertList {
+		slice = UpsertChangeTransferPolicy(slice, i)
+	}
+	return slice
+}
+
+func UpsertChangeTransferPolicy(slice []promoterv1alpha1.ChangeTransferPolicy, i promoterv1alpha1.ChangeTransferPolicy) []promoterv1alpha1.ChangeTransferPolicy {
```
crenshaw-dev: Should rename the `i` variable

cmd/main.go
```diff
@@ -225,9 +223,14 @@ func main() {
...
 
-	// setupLog.Info("starting informer factory")
-	// informerFactory.Start(processSignals.Done())
-	// informerFactory.WaitForCacheSync(processSignals.Done())
+	whr := webhookreceiver.NewWebhookReceiver(mgr)
+	go func() {
+		err = whr.Start(processSignals, ":3333")
+		if err != nil {
+			setupLog.Error(err, "unable to start webhook receiver")
+			os.Exit(1)
```
crenshaw-dev: Is this gonna muck with graceful shutdown of other stuff? Should we panic instead? Not sure if panics propagate up from goroutines.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -132,8 +138,25 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+		return fmt.Errorf("failed to set field index for .status.proposed.hydrated.sha: %w", err)
+	}
+
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &promoterv1alpha1.ChangeTransferPolicy{}, ".status.active.hydrated.sha", func(rawObj client.Object) []string {
+		//nolint:forcetypeassert
+		ctp := rawObj.(*promoterv1alpha1.ChangeTransferPolicy)
+		return []string{ctp.Status.Active.Hydrated.Sha}
+	}); err != nil {
+		return fmt.Errorf("failed to set field index for .status.active.hydrated.sha: %w", err)
+	}
```
crenshaw-dev: Shouldn't the indexes be set up in the CommitStatus controller, since that's what's retrieving the CTPs on these fields?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -250,6 +276,7 @@ func (r *ChangeTransferPolicyReconciler) setCommitStatusState(ctx context.Contex
 				Key:   status.Key,
 				Phase: string(promoterv1alpha1.CommitPhasePending),
 			})
+			logger.Info("No CommitStatus found", "key", status.Key, "sha", targetCommitBranchState.Hydrated.Sha)
```
crenshaw-dev: Maybe just one log statement outside all the ifs? Log key, sha, phase, tooManyMatchingShas, and len(csList.Items).

internal/controller/changetransferpolicy_controller_test.go
```diff
@@ -202,7 +206,80 @@ var _ = Describe("ChangeTransferPolicy Controller", func() {
...
+			changeTransferPolicy.Spec.ActiveBranch = "environment/development"
+			// We set auto merge too false to avoid the PR being merged automatically so we can run checks on it
+			changeTransferPolicy.Spec.AutoMerge = ptr.To(false)
+
+			Expect(k8sClient.Create(ctx, scmSecret)).To(Succeed())
+			Expect(k8sClient.Create(ctx, scmProvider)).To(Succeed())
+			Expect(k8sClient.Create(ctx, gitRepo)).To(Succeed())
+			Expect(k8sClient.Create(ctx, changeTransferPolicy)).To(Succeed())
+
+			gitPath, err := os.MkdirTemp("", "*")
```
crenshaw-dev, line 233: Need cleanup?

internal/controller/commitstatus_controller.go
```diff
@@ -120,13 +127,21 @@ func (r *CommitStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			//nolint: wrapcheck
 			return err
 		}
+
+		err = r.triggerReconcileChangeTransferPolicy(ctx, newCs, oldSha, cs.Spec.Sha)
+		if err != nil {
+			logger.Error(err, "failed to trigger reconcile of ChangeTransferPolicy via CommitStatus")
```
crenshaw-dev: Do we need to log here, or will it be logged higher up?

internal/controller/commitstatus_controller.go
```diff
@@ -162,3 +177,79 @@ func (r *CommitStatusReconciler) getCommitStatusProvider(ctx context.Context, co
...
+	if err != nil {
+		return fmt.Errorf("failed to list ChangeTransferPolicy with new sha %q: %w", newSha, err)
+	}
+
+	ctpList := utils.UpsertChangeTransferPolicyList(ctpListActiveOldSha.Items, ctpListActiveNewSha.Items)
+	ctpList = utils.UpsertChangeTransferPolicyList(ctpList, ctpListProposedOldSha.Items)
+	ctpList = utils.UpsertChangeTransferPolicyList(ctpList, ctpListProposedNewSha.Items)
+
+	logger.Info("ChangeTransferPolicy list", "count", len(ctpList), "oldSha", oldSha, "newSha", newSha)
+	for _, ctp := range ctpList {
```
crenshaw-dev, line 229:
```suggestion
	// TODO: parallelize this loop since it contains network calls.
	for _, ctp := range ctpList {
```

internal/controller/commitstatus_controller.go
```diff
@@ -162,3 +177,79 @@ func (r *CommitStatusReconciler) getCommitStatusProvider(ctx context.Context, co
...
+			".status.proposed.hydrated.sha": newSha,
+		}),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list ChangeTransferPolicy with new sha %q: %w", newSha, err)
+	}
+
+	ctpList := utils.UpsertChangeTransferPolicyList(ctpListActiveOldSha.Items, ctpListActiveNewSha.Items)
+	ctpList = utils.UpsertChangeTransferPolicyList(ctpList, ctpListProposedOldSha.Items)
+	ctpList = utils.UpsertChangeTransferPolicyList(ctpList, ctpListProposedNewSha.Items)
```
crenshaw-dev: Can this function signature be modified to handle all four lists at once?

internal/controller/suite_test.go
```diff
@@ -194,6 +202,15 @@ var _ = BeforeSuite(func() {
 	}).SetupWithManager(k8sManager)
 	Expect(err).ToNot(HaveOccurred())
 
+	webhookReceiverPort := WebhookReceiverPort + GinkgoParallelProcess()
+	whr := webhookreceiver.NewWebhookReceiver(k8sManager)
+	go func() {
+		err = whr.Start(ctx, fmt.Sprintf(":%d", webhookReceiverPort))
+		if err != nil {
+			os.Exit(1)
```
crenshaw-dev: Maybe use a t.Fatal or something like that instead?

internal/webhookreceiver/server.go
```diff
@@ -0,0 +1,132 @@
...
+
+	return nil
+}
+
+func (wr *webhookReceiver) postRoot(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "must be a POST request", http.StatusMethodNotAllowed)
+		return
+	}
+	jsonBytes, err := io.ReadAll(r.Body)
```
crenshaw-dev, line 71:
```suggestion
	// TODO: add a configurable payload max side for DoS protection.
	jsonBytes, err := io.ReadAll(r.Body)
```

internal/webhookreceiver/server.go
```diff
@@ -0,0 +1,132 @@
...
+	logger.Info("Triggered reconcile via webhook", "ChangeTransferPolicy", ctp.Namespace+"/"+ctp.Name)
+}
+
+func (wr *webhookReceiver) findChangeTransferPolicy(ctx context.Context, jsonBytes []byte) (*promoterv1alpha1.ChangeTransferPolicy, error) {
+	var beforeSha string
+	var ref string
+	ctpLists := promoterv1alpha1.ChangeTransferPolicyList{}
+
+	// TODO: probably move to own function once we start adding providers because rules might be more complex
+	if gjson.GetBytes(jsonBytes, "before").Exists() && gjson.GetBytes(jsonBytes, "pusher").Exists() {
```
crenshaw-dev, line 105: Can we read directly off the response body instead of having to unmarshal into a new bytes array?

crenshaw-dev (changes requested):

internal/controller/changetransferpolicy_controller.go
zachaller: I want to setup all indexes for resources in the controller that manages the resource because it centralizes all that code and you can't have the same index field setup twice. 

So the rule is whatever resource you want an index on should be setup in the controller that manages that resource. Then every controller that uses the same client under the manager can use the index.

internal/controller/changetransferpolicy_controller_test.go
zachaller: eh... maybe its in a temp dir will cleanup on reboots :P

There are other test based temp dirs not cleaned up so will do that in different PR

internal/webhookreceiver/server.go
zachaller: No Body is a `io.ReadCloser` which is not generally accepted anywhere and  gjson does not support that as a field anywhere they only take strings and bytes

internal/controller/changetransferpolicy_controller.go
crenshaw-dev: Gotcha, then probably just need to add a comment to explain why the indexes exist.

internal/utils/utils.go
```diff
@@ -207,3 +207,25 @@ func UpsertEnvironmentStatus(slice []promoterv1alpha1.EnvironmentStatus, i promo
 	}
 	return append(slice, i)
 }
+
+func UpsertChangeTransferPolicyList(slice []promoterv1alpha1.ChangeTransferPolicy, insertList ...[]promoterv1alpha1.ChangeTransferPolicy) []promoterv1alpha1.ChangeTransferPolicy {
+	for _, policies := range insertList {
+		for _, i := range policies {
+			slice = UpsertChangeTransferPolicy(slice, i)
```
crenshaw-dev:
```suggestion
		for _, p := range policies {
			slice = UpsertChangeTransferPolicy(slice, p)
```

crenshaw-dev (approved):

PR #139, opened 2024-12-03
feat(ci): add release workflow

crenshaw-dev:
This is basically just copy/paste of @leoluz's work on https://github.com/argoproj-labs/argocd-ephemeral-access/

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,104 @@
+# Based on: https://github.com/argoproj-labs/gitops-promoter/blob/main/.goreleaser.yaml
+
+# The lines below are called `modelines`. See `:help modeline`
+# yaml-language-server: $schema=https://raw.githubusercontent.com/goreleaser/goreleaser/v2.3.2/www/docs/static/schema.json
+# vim: set ts=2 sw=2 tw=0 fo=cnqoj
+
+project_name: gitops-promoter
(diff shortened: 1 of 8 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,104 @@
...
+      {{- else }}{{ .Arch }}{{ end }}
+      {{- if .Arm }}v{{ .Arm }}{{ end }}
+    # use zip for windows archives
+    format_overrides:
+      - goos: windows
+        format: zip
+
+release:
+  prerelease: auto
+  draft: false
```
zachaller: Can we set this to true

config/default/kustomization.yaml
```diff
@@ -26,117 +26,14 @@ resources:
...
 #- ../prometheus
 
-patches:
 # Protect the /metrics endpoint by putting it behind auth.
 # If you want your controller-manager to expose the /metrics
 # endpoint w/o any authn/z, please comment the following line.
+patches:
 - path: manager_auth_proxy_patch.yaml
-
-# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
```
zachaller, line 35: We can't remove all these comments it breaks kube builder

zachaller (approved):

PR #142, opened 2024-12-03
fix: codegen checks

zachaller:
The release pipeline ended up breaking the codegen checks this enables it and ignores the install.yaml file when running ./hack/manifest_release.sh locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,4 +33,4 @@ secret.yaml
 
 junit.xml
 
-dist
+./install.yaml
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -33,4 +33,4 @@ secret.yaml
 
 junit.xml
 
-dist
+./install.yaml
```
zachaller, line 36: This gets created at root when running `make manifests-releases`

crenshaw-dev (approved):

PR #146, opened 2024-12-04
fix: drop windows support

zachaller:


crenshaw-dev (approved):
win-doze, am i right

crenshaw-dev (approved):
winner of the darwin award

PR #153, opened 2024-12-05
fix: don't create previous cs check when no actives checks configured

zachaller:
If there are not configured active checks globally or on the previous environment do not create, a previous environment managed commit status.

Also do not create lowest level environments CommitStatus

### DIFF
--- a/internal/controller/promotionstrategy_controller.go
+++ b/internal/controller/promotionstrategy_controller.go
@@ -376,7 +376,8 @@ func (r *PromotionStrategyReconciler) updatePreviousEnvironmentCommitStatus(ctx 
 		}
 
 		if previousEnvironmentIndex != -1 && (len(ps.Spec.ActiveCommitStatuses) != 0 || len(ps.Spec.Environments[previousEnvironmentIndex].ActiveCommitStatuses) != 0) {
-			// If there is no configured active checks we should not create a commit status for the previous environment.
+			// Since there is at least one configured active check, and since this is not the first environment,
+			// we should not create a commit status for the previous environment.
 			err := r.createOrUpdatePreviousEnvironmentCommitStatus(ctx, ctpMap[environment.Branch], commitStatusPhase, previousEnvironmentStatus)
 			if err != nil {
 				return fmt.Errorf("failed to create or update previous environment commit status for branch %s: %w", environment.Branch, err)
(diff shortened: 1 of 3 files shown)


internal/controller/promotionstrategy_controller.go
```diff
@@ -164,11 +164,14 @@ func (r *PromotionStrategyReconciler) upsertChangeTransferPolicy(ctx context.Con
 		},
 	}
 
-	previousEnvironmentCommitStatusSelector := promoterv1alpha1.CommitStatusSelector{
-		Key: promoterv1alpha1.PreviousEnvironmentCommitStatusKey,
-	}
-	if !slices.Contains(pcNew.Spec.ProposedCommitStatuses, previousEnvironmentCommitStatusSelector) {
-		pcNew.Spec.ProposedCommitStatuses = append(pcNew.Spec.ProposedCommitStatuses, previousEnvironmentCommitStatusSelector)
+	previousEnvironmentIndex, _ := utils.GetPreviousEnvironmentStatusByBranch(*ps, environment.Branch)
+	if previousEnvironmentIndex > 0 && (len(ps.Spec.ActiveCommitStatuses) > 0 || len(ps.Spec.Environments[previousEnvironmentIndex].ActiveCommitStatuses) > 0) {
```
crenshaw-dev: Should this be `previousEnvironmentIndex >= 0`?

internal/controller/promotionstrategy_controller.go
```diff
@@ -372,9 +375,12 @@ func (r *PromotionStrategyReconciler) updatePreviousEnvironmentCommitStatus(ctx
 			commitStatusPhase = promoterv1alpha1.CommitPhaseSuccess
 		}
 
-		err := r.createOrUpdatePreviousEnvironmentCommitStatus(ctx, ctpMap[environment.Branch], commitStatusPhase, previousEnvironmentStatus)
-		if err != nil {
-			return fmt.Errorf("failed to create or update previous environment commit status for branch %s: %w", environment.Branch, err)
+		if previousEnvironmentIndex != -1 && (len(ps.Spec.ActiveCommitStatuses) != 0 || len(ps.Spec.Environments[previousEnvironmentIndex].ActiveCommitStatuses) != 0) {
+			// If there is no configured active checks we should not create a commit status for the previous environment.
```
crenshaw-dev:
```suggestion
			// Since there is at least one configured active check, and since this is not the first environment,
			// we should not create a commit status for the previous environment.
```

crenshaw-dev (changes requested):

internal/controller/promotionstrategy_controller.go
zachaller: No, because we don't want to create it on the lowest level environment either

crenshaw-dev: But if the previous environment's index is `0`, doesn't that mean we're currently working on environment `1`?

zachaller: Yea fixed the logic

crenshaw-dev (approved):

PR #161, opened 2024-12-17
chore: add tests for utility functions for upsert env & tuncate strings

zhaque44:


zachaller:
Will need to fix the lint issues, and then also add a new Make target that runs go test but excludes the `TestControllers` function which could actually probably be renamed to `TestControllersGinkgo`. Then we would probably need to add the new Make target to the github ci action.

zhaque44:
> and then also add a new Make target that runs go test but excludes the `TestControllers` function which could actually probably be renamed to `TestControllersGinkgo`. Then we would probably need to add the new Make target to the github ci action.

@zachaller I dont know my way around this codebase well, can you show me an example of that, I know gingko does that BDD style of tests

zhaque44:
@zachaller ive updated the func name & created the make target, before adding it to CI, I want to decide on what to name the make target, do you have any suggestions?

zachaller:
go-unit-test for the make target I think makes sense

zhaque44:
@zachaller its ready for review

zachaller:
The coverage is still broken need to merge reports not sure how because ginkgo doesn't seem to support the new go binary format

crenshaw-dev:
I wonder if you could do two uploads under different codecov tags, or whatever they call them.

zachaller:
hm... I would have to look into that guessing its more than just https://github.com/argoproj-labs/gitops-promoter/pull/161/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR53

emirot:
@zachaller I believe this PR is too old and should be closed

crenshaw-dev:
Closing. Will accept if tests are rewritten for Ginkgo

PR #167, opened 2024-12-27
feat: argo cd commitstatus manager

zachaller:


zachaller:
Going to turn this into a CRD instead

PR #170, opened 2025-01-03
feat: ArgoCDCommitStatus Kind

zachaller:


api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -0,0 +1,72 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ArgoCDCommitStatusSpec defines the desired state of ArgoCDCommitStatus.
+type ArgoCDCommitStatusSpec struct {
+	// +kubebuilder:validation:Required
+	PromotionStrategyRef ObjectReference `json:"promotionStrategyRef,omitempty"`
+
+	// +kubebuilder:validation:Required
+	ApplicationSelector *metav1.LabelSelector `json:"applicationSelector,omitempty"`
```
crenshaw-dev, line 32: Should we leave this field open for an additional expressions matching field?

api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// ArgoCDCommitStatusStatus defines the observed state of ArgoCDCommitStatus.
+type ArgoCDCommitStatusStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+	ApplicationsSelected []SelectedApplications `json:"applicationsSelected,omitempty"`
```
crenshaw-dev: I'd rename the field `SelectedApplications` to match the struct name. Just reads easier to me.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,371 @@
...
+			continue
+		}
+
+		aggregateItem := &Aggregate{
+			application: &application,
+		}
+
+		key := objKey{
+			repo:         strings.TrimRight(application.Spec.SourceHydrator.DrySource.RepoURL, ".git"),
+			targetBranch: application.Spec.SourceHydrator.SyncSource.TargetBranch,
```
crenshaw-dev: Will eventually probably need URL normalization.

api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+type SelectedApplications struct {
+	Namespace string            `json:"namespace"`
+	Name      string            `json:"name"`
+	Sate      CommitStatusPhase `json:"state"`
```
crenshaw-dev:
```suggestion
	State      CommitStatusPhase `json:"state"`
```

Should we actually just rename it phase?

api/v1alpha1/argocdcommitstatus_types.go
zachaller: That contains that already pretty sure if you mean this.

```
type LabelSelector struct {
	// matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels
	// map is equivalent to an element of matchExpressions, whose key field is "key", the
	// operator is "In", and the values array contains only "value". The requirements are ANDed.
	// +optional
	MatchLabels map[string]string `json:"matchLabels,omitempty" protobuf:"bytes,1,rep,name=matchLabels"`
	// matchExpressions is a list of label selector requirements. The requirements are ANDed.
	// +optional
	// +listType=atomic
	MatchExpressions []LabelSelectorRequirement `json:"matchExpressions,omitempty" protobuf:"bytes,2,rep,name=matchExpressions"`
}
```

api/v1alpha1/argocdcommitstatus_types.go
zachaller: done

api/v1alpha1/argocdcommitstatus_types.go
zachaller: done

internal/controller/argocdcommitstatus_controller.go
zachaller: Yea for sure

zachaller: Promotor always needs the https url pretty sure so we need to get it into that form from say `git@*` etc etc

docs/crd-specs.md
```diff
@@ -61,3 +61,12 @@ auth mechanism.
 ```yaml
 {!docs/example-resources/ScmProvider.yaml!}
 ```
+
+## ArgoCDCommitStatus
+
+An ArgoCDCommitStatus is used as a way to aggregate all the ArgoCD Application's that are being used in the promotion strategy. It is used
+to check the status of the ArgoCD Application's that are being used in the promotion strategy.
```
crenshaw-dev:
```suggestion
An ArgoCDCommitStatus is used as a way to aggregate all the Argo CD Application's that are being used in the promotion strategy. It is used
to check the status of the Argo CD Application's that are being used in the promotion strategy.
```

for goodness and truth

internal/controller/argocdcommitstatus_controller.go
crenshaw-dev: Yeah, I think Argo's normalization function always goes to https

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+	for _, obj := range ul.Items {
+		var application argocd.ArgoCDApplication
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &application)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to cast unstructured object to typed object: %w", err)
+		}
+
+		if application.Spec.SourceHydrator == nil {
+			continue
+		}
```
crenshaw-dev: We should document that this currently only supports apps using the hydrator, even though you could theoretically use promoter without Argo CD's hydrator.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+				Sha:   application.Status.Sync.Revision,
+				Phase: phase,
+			},
+		}
+		aggregateItem.selectedApplication = &promoterv1alpha1.ApplicationsSelected{}
+		aggregateItem.selectedApplication.Phase = phase
+		aggregateItem.selectedApplication.Sha = application.Status.Sync.Revision
+		aggregateItem.selectedApplication.Name = application.GetName()
+		aggregateItem.selectedApplication.Namespace = application.GetNamespace()
+		aggregateItem.selectedApplication.LastTransitionTime = application.Status.Health.LastTransitionTime
```
crenshaw-dev: Initialize selectedApplication inline? Would be easier to read.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+			} else {
+				pending++
+			}
+
+			// Find the most recent last transition time
+			if s.selectedApplication.LastTransitionTime != nil {
+				if mostRecentLastTransitionTime == nil || s.selectedApplication.LastTransitionTime.Time.After(mostRecentLastTransitionTime.Time) {
+					mostRecentLastTransitionTime = s.selectedApplication.LastTransitionTime
+				}
+			}
```
crenshaw-dev: `&&` instead of nesting?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+		}
+
+		// Did the mostRecentLastTransitionTime occur more than 5 seconds ago
+		if mostRecentLastTransitionTime != nil && time.Since(mostRecentLastTransitionTime.Time) < 5*time.Second {
```
crenshaw-dev: I'm a little worried about this check. Seems like if we have one flappy app, we never update the CommitStatus. Maybe we should update everything except the Phase?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+			return ctrl.Result{RequeueAfter: (5*time.Second - time.Since(mostRecentLastTransitionTime.Time)) + 1*time.Second}, nil
+		}
+
+		// Resolve state
+		if healthy == len(aggregateItem) {
+			resolvedPhase = promoterv1alpha1.CommitPhaseSuccess
+			desc = fmt.Sprintf("%d/%d apps are healthy", healthy, len(aggregateItem))
+		} else if degraded == len(aggregateItem) {
+			resolvedPhase = promoterv1alpha1.CommitPhaseFailure
+			desc = fmt.Sprintf("%d/%d apps are degraded", healthy, len(aggregateItem))
```
```suggestion
			desc = fmt.Sprintf("%d/%d apps are degraded", degraded, len(aggregateItem))
```

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+		} else {
+			desc = fmt.Sprintf("Waiting for apps to be healthy (%d/%d healthy, %d/%d degraded, %d/%d pending)", healthy, len(aggregateItem), degraded, len(aggregateItem), pending, len(aggregateItem))
```
```suggestion
			desc = fmt.Sprintf("Waiting for apps to be healthy (%d healthy, %d degraded, %d pending)", healthy, degraded, pending)
```

Maybe cleaner?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+		return fmt.Errorf("failed to create controller: %w", err)
+	}
+	return nil
+}
+
+func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.Context, argoCDCommitStatus promoterv1alpha1.ArgoCDCommitStatus, targetBranch string, repo string, sha string, phase promoterv1alpha1.CommitStatusPhase, desc string) error {
+	logger := log.FromContext(ctx)
+
+	commitStatusName := targetBranch + "/health"
+	resourceName := strings.ReplaceAll(commitStatusName, "/", "-") + "-" + hash([]byte(repo))
```
crenshaw-dev: We should document the limitation that the repo URL has to exactly match

crenshaw-dev (changes requested):

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+	promoterv1alpha1 "github.com/argoproj-labs/gitops-promoter/api/v1alpha1"
+)
+
+var gvk = schema.GroupVersionKind{
+	Group:   "argoproj.io",
+	Version: "v1alpha1",
+	Kind:    "Application",
+}
+
+type Aggregate struct {
```
crenshaw-dev: Could/should this struct be private?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+	application         *argocd.ArgoCDApplication
+	commitStatus        *promoterv1alpha1.CommitStatus
+	selectedApplication *promoterv1alpha1.ApplicationsSelected
```
crenshaw-dev: Field name should be plural, right?

I wonder about the word "selected"... Maybe "targeted would be clearer? Not sure.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+}
+
+type objKey struct {
+	repo         string
+	targetBranch string
+}
```
```suggestion
// objKey describes a unique grouping of applications based on their repo URL and their target sync branch.
// This is effectively the key for the aggregate commit status representing a group of applications
// syncing from the same branch.
type objKey struct {
	repo         string
	targetBranch string
}
```

crenshaw-dev: Maybe also rename the struct? objKey is a bit vague.

internal/controller/argocdcommitstatus_controller.go
crenshaw-dev: Actually maybe initialize the whole struct here? Doesn't look like it needs to be initialized above.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+			Name:               application.GetName(),
+			Phase:              phase,
+			Sha:                application.Status.Sync.Revision,
+			LastTransitionTime: application.Status.Health.LastTransitionTime,
+		})
+
+		aggregates[key] = append(aggregates[key], aggregateItem)
+	}
+
+	for key, aggregateItem := range aggregates {
```
crenshaw-dev: Maybe aggregateStatus instead of aggregateItem?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+		var resolvedSha string
```
crenshaw-dev: Initialize below instead of here?

internal/controller/argocdcommitstatus_controller.go
crenshaw-dev: I think this would be more readable as two utility functions: one to get counts, the other to get the most recent transition time.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,398 @@
...
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to update ArgoCDCommitStatus status during requeuing: %w", err)
+			}
+			logger.Info("Most recent last transition time is less than 5 seconds old, requeuing")
+			// Requeue with a delay of the time until the mostRecentLastTransitionTime occurred + 1 second
+			return ctrl.Result{RequeueAfter: (5*time.Second - time.Since(mostRecentLastTransitionTime.Time)) + 1*time.Second}, nil
+		}
+
+		// Resolve state
+		if healthy == len(aggregateItem) {
```
crenshaw-dev: Maybe another utility function to get the phase and description? I'm basically trying to get the contents of this loop nice and short, and the utilities testable.

crenshaw-dev (changes requested):

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,420 @@
...
+		resolvedPhase, desc := r.calculateAggregatedPhaseAdnDescription(appsInEnvironment, resolvedSha)
+
+		mostRecentLastTransitionTime := r.getMostRecentLastTransitionTime(appsInEnvironment)
+
+		// Did the mostRecentLastTransitionTime occur more than 5 seconds ago
+		if mostRecentLastTransitionTime != nil && time.Since(mostRecentLastTransitionTime.Time) < 5*time.Second {
+			err = r.updateAggregatedCommitStatus(ctx, argoCDCommitStatus, targetBranch, repo, resolvedSha, promoterv1alpha1.CommitPhasePending, desc)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
```
crenshaw-dev: I think you could pass `mostRecentLastTransitionTime` to `calculateAggregatedPhaseAdnDescription` and have it handle setting `resolvedPhase` to `Pending` if the transition is too recent.

That would let you get rid of the `if` block and just call updateAggregatedCommitStatus` in one place.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,420 @@
...
+		}
+		argoCDCommitStatus.Status.ApplicationsSelected = append(argoCDCommitStatus.Status.ApplicationsSelected, *aggregateItem.selectedApplications)
+
+		aggregates[key] = append(aggregates[key], aggregateItem)
+	}
+
+	return aggregates, nil
+}
+
+func (r *ArgoCDCommitStatusReconciler) calculateAggregatedPhaseAdnDescription(appsInEnvironment []*aggregate, resolvedSha string) (promoterv1alpha1.CommitStatusPhase, string) {
```
crenshaw-dev: Typo in the function name

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,401 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the ArgoCDCommitStatus object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.1/pkg/reconcile
```
crenshaw-dev, line 79: Remove boilerplate?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	for targetBranch, appsInEnvironment := range groupedArgoCDApps {
+		gitOperation, err := git.NewGitOperations(ctx, r.Client, gitAuthProvider, r.PathLookup, repositoryRef, &argoCDCommitStatus, targetBranch)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to initialize git client: %w", err)
+		}
+
+		var resolvedSha string
+		resolvedSha, err = gitOperation.LsRemote(ctx, targetBranch)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to ls-remote sha: %w", err)
```
```suggestion
			return ctrl.Result{}, fmt.Errorf("failed to ls-remote sha for branch %q: %w", targetBranch, err)
```

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	for targetBranch, appsInEnvironment := range groupedArgoCDApps {
+		gitOperation, err := git.NewGitOperations(ctx, r.Client, gitAuthProvider, r.PathLookup, repositoryRef, &argoCDCommitStatus, targetBranch)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to initialize git client: %w", err)
+		}
+
+		var resolvedSha string
+		resolvedSha, err = gitOperation.LsRemote(ctx, targetBranch)
```
```suggestion
		resolvedSha, err := gitOperation.LsRemote(ctx, targetBranch)
```

Equivalent?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,401 @@
...
+
+	err = r.Status().Update(ctx, &argoCDCommitStatus)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update ArgoCDCommitStatus status: %w", err)
+	}
+
+	return ctrl.Result{RequeueAfter: 15 * time.Second}, nil // Timer for now :(
+}
+
+func (r *ArgoCDCommitStatusReconciler) groupArgoCDApplicationsWithPhase(argoCDCommitStatus *promoterv1alpha1.ArgoCDCommitStatus, ulAppList unstructured.UnstructuredList) (map[string][]*aggregate, error) {
```
crenshaw-dev, line 151:
```suggestion
// groupArgoCDApplicationsWithPhase returns a map. The key is a branch name. The value is a list of apps configured for that target branch, along with the commit status for that one app.
// As a side-effect, this function updates argoCDCommitStatus to represent the aggregate status
// of all matching apps.
func (r *ArgoCDCommitStatusReconciler) groupArgoCDApplicationsWithPhase(argoCDCommitStatus *promoterv1alpha1.ArgoCDCommitStatus, ulAppList unstructured.UnstructuredList) (map[string][]*aggregate, error) {
```

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,401 @@
...
+
+	for _, ulApp := range ulAppList.Items {
+		var application argocd.ArgoCDApplication
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(ulApp.Object, &application)
+		if err != nil {
+			return map[string][]*aggregate{}, fmt.Errorf("failed to cast unstructured object to typed object: %w", err)
+		}
+
+		if application.Spec.SourceHydrator == nil {
+			continue
```
crenshaw-dev: Maybe just go ahead and throw an error? Would be consistent with how we handle non-matching repo URLs.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	} else if degraded == len(appsInEnvironment) {
+		resolvedPhase = promoterv1alpha1.CommitPhaseFailure
+		desc = fmt.Sprintf("%d/%d apps are degraded", degraded, len(appsInEnvironment))
+	} else {
+		desc = fmt.Sprintf("Waiting for apps to be healthy (%d healthy, %d degraded, %d pending)", healthy, degraded, pending)
+	}
+
+	if mostRecentLastTransitionTime != nil && time.Since(mostRecentLastTransitionTime.Time) < 5*time.Second {
+		return promoterv1alpha1.CommitPhasePending, desc
+	}
```
crenshaw-dev, line 240:
```suggestion
	// Don't consider the aggregate status healthy until 5s after the most recent transition.
	// This helps avoid prematurely accepting a transitive healthy state.
	if mostRecentLastTransitionTime != nil && time.Since(mostRecentLastTransitionTime.Time) < 5*time.Second {
		return promoterv1alpha1.CommitPhasePending, desc
	}
```

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		if client.IgnoreNotFound(err) != nil {
+			return fmt.Errorf("failed to get CommitStatus object: %w", err)
+		}
+		// Create
+		err = r.Client.Create(ctx, &desiredCommitStatus)
+		logger.Info("Created ArgoCDCommitStatus", "name", desiredCommitStatus.Name)
+		if err != nil {
+			return fmt.Errorf("failed to create CommitStatus object: %w", err)
+		}
+		currentCommitStatus = desiredCommitStatus
```
crenshaw-dev: This looks like it goes unused.

docs/crd-specs.md
```diff
@@ -61,3 +61,12 @@ auth mechanism.
 ```yaml
 {!docs/example-resources/ScmProvider.yaml!}
 ```
+
+## ArgoCDCommitStatus
+
+An ArgoCDCommitStatus is used as a way to aggregate all the Argo CD Application's that are being used in the promotion strategy. It is used
+to check the status of the Argo CD Application's that are being used in the promotion strategy.
```
crenshaw-dev:
```suggestion
An ArgoCDCommitStatus is used as a way to aggregate all the Argo CD Applications that are being used in the promotion strategy. It is used
to check the status of the Argo CD Applications that are being used in the promotion strategy.
```

crenshaw-dev (approved):

crenshaw-dev (approved):

PR #199, opened 2025-02-17
feat: add filter path support

zachaller:


api/v1alpha1/changetransferpolicy_types.go
```diff
@@ -51,6 +51,9 @@ type ChangeTransferPolicySpec struct {
 	// ProposedCommitStatuses lists the statuses to be monitored on the proposed branch
 	// +kubebuilder:validation:Optional
 	ProposedCommitStatuses []CommitStatusSelector `json:"proposedCommitStatuses"`
+
+	// +kubebuilder:validation:Optional
+	OpenPullerRequestFilter *OpenPullerRequestFilter `json:"openPullRequestFilter,omitempty"`
```
crenshaw-dev: s/puller/pull

crenshaw-dev: Maybe something like "relevantFilesFilter"?

internal/git/git.go
```diff
@@ -265,12 +266,37 @@ func (g *GitOperations) IsPullRequestRequired(ctx context.Context, environmentNe
...
+				logger.V(4).Info("YAML file changed", "file", file)
+				return true, nil
+			}
+		}
+		return false, nil
+	}
+
+	for _, file := range diffOutput {
+		for _, pathPattern := range filter.Paths {
+			matched, err := filepath.Match(pathPattern, file)
```
crenshaw-dev: I'd suggest a different path match library, like maybe doublestar. The filepath glob language is pretty limited.

crenshaw-dev: We should probably think about how we can mitigate directory traversal risks when applying glob patterns. Since we return a boolean, I think the only risk is path enumeration.

crenshaw-dev (changes requested):

api/v1alpha1/changetransferpolicy_types.go
zachaller: I think under this top level key, we would also include things like commitMessageFilters, commitUsers, etc etc

Issue #141, opened 2024-12-03
feat: add gitlab support


zachaller:

robinlieb:
@zachaller 
Working in that feature, created a draft pull request in #205 for that. 
I have two questions that I would like to clarify:
1. GitLab has a slightly different approach when it comes to orgs, which is also the possibility to have subgroups, so a repo could have the slug (.../group/subgroup/project). To cover that (or make it more clear), I have created dedicated objects for the git repository CR for each SCM. This would mean that it breaks the current approach. Technically this could also be covered with the current implementation, to just put the GitLab namespace in the owner property. Is there any preference on one approach?
2. For the commit phase there are currently three phases. From the comments if the Phase property I guess the intention was to have one project generic specification and then have each SCM provider implement its own logic the map these. Is that right?

crenshaw-dev:
On 1, I like the spec being super explicit, I think it's a better user experience. At this early stage, the breaking change is probably fine. But curious if @zachaller thinks this would be a good opportunity to practice our API versioning skills.

On 2, that's exactly correct. The phases are meant to generally map to their SCM-specific equivalents.

zachaller:
@robinlieb thanks for the contributions

On 1, we did talk about the breaking change in the API and with how early the project is we are good with that change for now until we setup proper conversion webhook infrastructure.

robinlieb:
Thanks for the clarification and the initial review!
For the git username I created a dedicated pr (#206).
I try to resolve the comments in #205 by the end of this week, but will update the PR once done.

crenshaw-dev:
Done here: https://github.com/argoproj-labs/gitops-promoter/pull/205

Thanks again @robinlieb!

PR #205, opened 2025-03-02
feat: add gitlab support

robinlieb:
Implements #141


api/v1alpha1/gitrepository_types.go
```diff
@@ -25,10 +25,9 @@ import (
 
 // GitRepositorySpec defines the desired state of GitRepository
 type GitRepositorySpec struct {
-	// +kubebuilder:validation:Required
-	Owner string `json:"owner"`
-	// +kubebuilder:validation:Required
-	Name string `json:"name"`
+	GitHub GitHubRepo `json:"github,omitempty"`
+	GitLab GitLabRepo `json:"gitlab,omitempty"`
```
crenshaw-dev: Maybe pointers so it's super easy to tell if one's meant to be used?

api/v1alpha1/gitrepository_types.go
```diff
@@ -25,10 +25,9 @@ import (
...
+	Fake   FakeRepo   `json:"fake,omitempty"`
 	// +kubebuilder:validation:Required
 	ScmProviderRef ObjectReference `json:"scmProviderRef"`
```
crenshaw-dev, line 32: We'll need new validation to throw an error if someone uses a repo from one SCM with the SCMProvider of another SCM.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -340,7 +344,16 @@ func (r *ChangeTransferPolicyReconciler) creatOrUpdatePullRequest(ctx context.Co
...
+
+	var prName string
+	switch {
+	case gitRepo.Spec.GitHub.Name != "":
+		prName = utils.KubeSafeUniqueName(ctx, utils.GetPullRequestName(ctx, gitRepo.Spec.GitHub.Owner, gitRepo.Spec.GitHub.Name, ctp.Spec.ProposedBranch, ctp.Spec.ActiveBranch))
+	case gitRepo.Spec.GitLab.Name != "":
+		prName = utils.KubeSafeUniqueName(ctx, utils.GetPullRequestName(ctx, gitRepo.Spec.GitLab.Namespace, gitRepo.Spec.GitLab.Name, ctp.Spec.ProposedBranch, ctp.Spec.ActiveBranch))
+	case gitRepo.Spec.Fake.Name != "":
+		prName = utils.KubeSafeUniqueName(ctx, utils.GetPullRequestName(ctx, gitRepo.Spec.Fake.Owner, gitRepo.Spec.Fake.Name, ctp.Spec.ProposedBranch, ctp.Spec.ActiveBranch))
+	}
```
crenshaw-dev: Could probably extract the KubeSafeUniqueName call instead of duplicating.

internal/git/git.go
```diff
@@ -310,7 +310,7 @@ func (g *GitOperations) runCmd(ctx context.Context, directory string, args ...st
 	cmd := exec.Command("git", args...)
 	cmd.Env = []string{
 		"GIT_ASKPASS=promoter_askpass.sh", // Needs to be on path
-		fmt.Sprintf("GIT_USER=%s", user),
+		fmt.Sprintf("GIT_USERNAME=%s", user),
```
crenshaw-dev: Was this previously just broken? If so, might justify its own PR/test.

internal/scms/gitlab/commit_status.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return &CommitStatus{client: client, k8sClient: k8sClient}, nil
+}
+
+func (cs *CommitStatus) Set(ctx context.Context, commitStatus *v1alpha1.CommitStatus) (*v1alpha1.CommitStatus, error) {
+	logger := log.FromContext(ctx)
+	logger.Info("Setting Commit Phase")
+
+	repo, err := utils.GetGitRepositorytFromObjectKey(ctx, cs.k8sClient, client.ObjectKey{
```
crenshaw-dev:
```suggestion
	repo, err := utils.GetGitRepositoryFromObjectKey(ctx, cs.k8sClient, client.ObjectKey{
```

Not sure if this is a new typo or was already there

internal/scms/gitlab/commit_status.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create status: %w", err)
+	}
+
+	logger.Info("GitLab rate limits",
+		"limit", resp.Header.Get("Ratelimit-Limit"),
+		"remaining", resp.Header.Get("Ratelimit-Remaining"),
+		"reset", resp.Header.Get("Ratelimit-Reset"),
+		"url", resp.Request.URL,
+	)
```
crenshaw-dev: Is rate limiting done on a per-token basis? If so, can we make sure we log something like the ScmProvider's name so we can correlate those two?

internal/scms/gitlab/git_operations.go
```diff
@@ -0,0 +1,63 @@
...
+type GitAuthenticationProvider struct {
+	scmProvider *v1alpha1.ScmProvider
+	secret      *v1.Secret
+	client      *gitlab.Client
+}
+
+func NewGitlabGitAuthenticationProvider(scmProvider *v1alpha1.ScmProvider, secret *v1.Secret) GitAuthenticationProvider {
+	client, err := GetClient(*secret, scmProvider.Spec.GitLab.Domain)
+	if err != nil {
+		panic(err)
```
crenshaw-dev: Could we just return up the error?

internal/scms/gitlab/git_operations.go
```diff
@@ -0,0 +1,63 @@
...
+		scmProvider: scmProvider,
+		secret:      secret,
+		client:      client,
+	}
+}
+
+func (gl GitAuthenticationProvider) GetGitHttpsRepoUrl(repo v1alpha1.GitRepository) string {
+	if gl.scmProvider.Spec.GitLab != nil && gl.scmProvider.Spec.GitLab.Domain != "" {
+		return fmt.Sprintf("https://%s/%s/%s.git", gl.scmProvider.Spec.GitLab.Domain, repo.Spec.GitLab.Namespace, repo.Spec.GitLab.Name)
+	}
```
crenshaw-dev: We should probably do some validation on these URL parts.

internal/scms/gitlab/pullrequest.go
```diff
@@ -0,0 +1,236 @@
...
+	)
+	if err != nil {
+		return "", fmt.Errorf("failed to create pull request: %w", err)
+	}
+
+	logger.Info("GitLab rate limits",
+		"limit", resp.Header.Get("Ratelimit-Limit"),
+		"remaining", resp.Header.Get("Ratelimit-Remaining"),
+		"reset", resp.Header.Get("Ratelimit-Reset"),
+		"url", resp.Request.URL,
```
crenshaw-dev: Ditto re: logging a field to identify the token.

internal/scms/gitlab/pullrequest.go
```diff
@@ -0,0 +1,236 @@
...
+	logger.V(4).Info("github response status",
+		"status", resp.Status)
+
+	return strconv.Itoa(mr.IID), nil
+}
+
+func (pr *PullRequest) Update(ctx context.Context, title, description string, prObj *v1alpha1.PullRequest) error {
+	logger := log.FromContext(ctx)
+
+	mrIID, err := strconv.Atoi(prObj.Status.ID)
```
crenshaw-dev, line 76: I wonder if we should standardize on "Merge Request" within the context of the gitlab-specific code for variable names, etc.

internal/scms/gitlab/pullrequest.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return nil
+}
+
+func (pr *PullRequest) Close(ctx context.Context, prObj *v1alpha1.PullRequest) error {
+	logger := log.FromContext(ctx)
+
+	mrIID, err := strconv.Atoi(prObj.Status.ID)
+	if err != nil {
+		return fmt.Errorf("invalid merge request ID: %w", err)
```
crenshaw-dev:
```suggestion
		return fmt.Errorf("invalid merge request ID: %q", err)
```

Maybe? Just in case it's an empty string

internal/scms/gitlab/pullrequest.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to close merge request: %w", err)
+	}
+
+	logger.Info("GitLab rate limits",
+		"limit", resp.Header.Get("Ratelimit-Limit"),
+		"remaining", resp.Header.Get("Ratelimit-Remaining"),
+		"reset", resp.Header.Get("Ratelimit-Reset"),
+		"url", resp.Request.URL,
+	)
```
crenshaw-dev: Will stop +1'ing the extra field request, but will say here that it probably makes sense to pull a utility function for adding rate limiting fields.

internal/scms/gitlab/pullrequest.go
```diff
@@ -0,0 +1,236 @@
...
+	logger.V(4).Info("github response status",
+		"status", resp.Status)
+
+	if len(mrs) > 0 {
+		prObj.Status.ID = strconv.Itoa(mrs[0].IID)
+		prObj.Status.State = v1alpha1.PullRequestState(strings.TrimSuffix(mrs[0].State, "ed"))
```
crenshaw-dev: Might be worth documenting why the "ed" strip works. Or possibly switching to an explicit switch statement, if that's more readable.

crenshaw-dev (changes requested):
Looking fantastic! Did a quick first round review.

internal/git/git.go
zachaller: The askpass script wants GIT_USERNAME https://github.com/argoproj-labs/gitops-promoter/blob/main/hack/git/promoter_askpass.sh#L5

crenshaw-dev: Fixed in a different PR

zachaller:
One thing to note and this is more information than anything that needs to be done in this PR, there is also webhook support that is somewhat required at lest for faster reconciliation triggering than the default I think 10 minutes. There is not much support for multiple providers but you can see that [here](https://github.com/argoproj-labs/gitops-promoter/blob/main/internal/webhookreceiver/server.go)

api/v1alpha1/gitrepository_types.go
robinlieb: Updated, thanks!

internal/controller/changetransferpolicy_controller.go
robinlieb: Updated, thanks!

internal/scms/gitlab/commit_status.go
robinlieb: Updated, thanks!

internal/scms/gitlab/pullrequest.go
robinlieb: Created util function to make it more explicit, thanks!

internal/scms/gitlab/git_operations.go
robinlieb: Done, thanks!

api/v1alpha1/gitrepository_types.go
robinlieb: Added a check [here](https://github.com/argoproj-labs/gitops-promoter/pull/205/files#diff-a7cadd8b7616e655fd258138320b33eb96f7cce8e8663ec92651f49ce07df000R40), is this the way you had in mind?

internal/scms/gitlab/git_operations.go
robinlieb: Just generally if its a valid url or do you have something specific in mind?

crenshaw-dev: I think each part should be independently validated. The domain can probably be validated with the `url` package. The other two parts probably have to match these rules: https://docs.gitlab.com/user/reserved_names/

And then the whole URL should be assembled using the `url` package instead of just string formatting.

api/v1alpha1/gitrepository_types.go
crenshaw-dev: Yep, that looks good!

internal/scms/gitlab/git_operations.go
robinlieb: Added pattern validation for the `Namespace` and `Name` properties based on the provided GitLab documentation .
Also added a check if parsing of the constructed url is successful. If not, the `GetGitHttpsRepoUrl` returns empty string. Would probably better to return an error for proper error handling, but would go beyond the scope of this PR in my perspective. Would work on that on a dedicated PR if possible.

internal/scms/gitlab/pullrequest.go
robinlieb: Do you mean that it's currently e.g. `repo` vs. `gitRepo`?

robinlieb:
@zachaller good catch, created #222 to keep track of that.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -394,6 +395,13 @@ func (r *ArgoCDCommitStatusReconciler) getGitAuthProvider(ctx context.Context, a
 	case scmProvider.Spec.GitHub != nil:
 		logger.V(4).Info("Creating GitHub git authentication provider")
 		return github.NewGithubGitAuthenticationProvider(scmProvider, secret), ps.Spec.RepositoryReference, nil
+	case scmProvider.Spec.GitLab != nil:
+		logger.V(4).Info("Creating GitLab git authentication provider")
+		gitlabClient, err := gitlab.NewGitlabGitAuthenticationProvider(scmProvider, secret)
+		if err != nil {
+			return nil, ps.Spec.RepositoryReference, fmt.Errorf("failed to create Git Client: %w", err)
```
```suggestion
			return nil, ps.Spec.RepositoryReference, fmt.Errorf("failed to create GitLab client: %w", err)
```

internal/controller/changetransferpolicy_controller.go
```diff
@@ -175,6 +176,13 @@ func (r *ChangeTransferPolicyReconciler) getGitAuthProvider(ctx context.Context,
 	case scmProvider.Spec.GitHub != nil:
 		logger.V(4).Info("Creating GitHub git authentication provider")
 		return github.NewGithubGitAuthenticationProvider(scmProvider, secret), nil
+	case scmProvider.Spec.GitLab != nil:
+		logger.V(4).Info("Creating GitLab git authentication provider")
+		provider, err := gitlab.NewGitlabGitAuthenticationProvider(scmProvider, secret)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create Auth Provider: %w", err)
```
```suggestion
			return nil, fmt.Errorf("failed to create GitLab Auth Provider: %w", err)
```

internal/scms/gitlab/git_operations.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (gl GitAuthenticationProvider) GetUser(ctx context.Context) (string, error) {
+	return "oauth2", nil
+}
+
+func GetClient(secret v1.Secret, domain string) (*gitlab.Client, error) {
+	token := string(secret.Data["token"])
+	if token == "" {
+		return nil, fmt.Errorf("GitLab token is required")
```
crenshaw-dev:
```suggestion
		return nil, fmt.Errorf("secret %q is missing required data key 'token'", secret.Name)
```

internal/scms/gitlab/pullrequest.go
crenshaw-dev: nah, it just that most stuff in the new gitlab code refers to "pull requests" even though gitlab calls them "merge requests." Not a big deal, but maybe better to stick with the SCM's convention.

internal/scms/gitlab/pullrequest.go
```diff
@@ -0,0 +1,230 @@
...
+
+	return strconv.Itoa(mr.IID), nil
+}
+
+func (pr *PullRequest) Update(ctx context.Context, title, description string, prObj *v1alpha1.PullRequest) error {
+	logger := log.FromContext(ctx)
+
+	mrIID, err := strconv.Atoi(prObj.Status.ID)
+	if err != nil {
+		return fmt.Errorf("failed to convert MR number to int: %w", err)
```
crenshaw-dev:
```suggestion
		return fmt.Errorf("failed to convert MR ID %q to int: %w", prObj.Status.ID, err)
```

internal/scms/gitlab/pullrequest.go
```diff
@@ -0,0 +1,230 @@
...
+
+	return nil
+}
+
+func (pr *PullRequest) Close(ctx context.Context, prObj *v1alpha1.PullRequest) error {
+	logger := log.FromContext(ctx)
+
+	mrIID, err := strconv.Atoi(prObj.Status.ID)
+	if err != nil {
+		return fmt.Errorf("failed to convert MR number to int: %w", err)
```
crenshaw-dev:
```suggestion
		return fmt.Errorf("failed to convert MR ID %q to int: %w", prObj.Status.ID, err)
```

docs/getting-started.md
```diff
@@ -97,6 +98,46 @@ spec:
     The GitRepository and ScmProvider also need to be installed to the same namespace that you plan on creating PromotionStrategy 
     resources in, and it also needs to be in the same namespace of the secret it references.
 
+## GitLab Configuration
+
+To configure the GitOps Promoter with GitLab, you will need to create a GitLab Access Token with `api` and `write_repository` scopes and configure the necessary resources to allow the promoter to interact with your repository. This Access Token should be used in a secret as follows:
```
crenshaw-dev: I had to use the Developer role. That seems to be the lowest role which allows write access to MRs and commit statuses.

```suggestion
To configure the GitOps Promoter with GitLab, you will need to create a GitLab Access Token under the "Developer" role with `api` and `write_repository` scopes and configure the necessary resources to allow the promoter to interact with your repository. This Access Token should be used in a secret as follows:
```

crenshaw-dev (changes requested):
Let's heckin' go: https://gitlab.com/mac9416/gitops-promoter-tests/-/merge_requests

crenshaw-dev:
I'm not getting anything for rate limits:

```
2025-03-24T10:15:27.383156-04:00        INFO    GitLab rate limits      {"controller": "pullrequest", "controllerGroup": "promoter.argoproj.io", "controllerKind": "PullRequest", "PullRequest": {"name":"mac9416-gitops-promoter-tests-environment-dev-next-environment-dev-acfd9df0","namespace":"promoter-system"}, "namespace": "promoter-system", "name": "mac9416-gitops-promoter-tests-environment-dev-next-environment-dev-acfd9df0", "reconcileID": "d28b1c59-8dcf-4fcf-9a60-c935b5d1cdc7", "scmProvider": "gitlab", "limit": "", "remaining": "", "reset": "", "url": "https://gitlab.com/api/v4/merge_requests?source_branch=environment%2Fdev-next&state=opened&target_branch=environment%2Fdev"}
```

robinlieb:
> I'm not getting anything for rate limits:
> 
> ```
> 2025-03-24T10:15:27.383156-04:00        INFO    GitLab rate limits      {"controller": "pullrequest", "controllerGroup": "promoter.argoproj.io", "controllerKind": "PullRequest", "PullRequest": {"name":"mac9416-gitops-promoter-tests-environment-dev-next-environment-dev-acfd9df0","namespace":"promoter-system"}, "namespace": "promoter-system", "name": "mac9416-gitops-promoter-tests-environment-dev-next-environment-dev-acfd9df0", "reconcileID": "d28b1c59-8dcf-4fcf-9a60-c935b5d1cdc7", "scmProvider": "gitlab", "limit": "", "remaining": "", "reset": "", "url": "https://gitlab.com/api/v4/merge_requests?source_branch=environment%2Fdev-next&state=opened&target_branch=environment%2Fdev"}
> ```

As far as I understand they are only provided on GitLab Self-Managed and also only once the limit is reached. There is no information provided to get the current status of the rate limit. I changed the implementation to only log if they are provided in the response.

crenshaw-dev (approved):

zachaller (approved):

PR #216, opened 2025-03-18
feat: add annotation for on copied active commit status

zachaller:
Adds an annotation called `promoter.argoproj.io/previous-environment-statuses` to the copied previous environments aggregated commit status, with a breakdown of all the active commit statuses configured and their state.

```
kind: CommitStatus
metadata:
  annotations:
    promoter.argoproj.io/previous-environment-statuses: |
      automated-testing: pending
      performance-test: pending
```

### DIFF
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -32,3 +32,6 @@ const PreviousEnvironmentCommitStatusKey = "promoter-previous-environment"
 
 // ReconcileAtAnnotation is the annotation used to indicate when the webhook triggered a reconcile
 const ReconcileAtAnnotation = "promoter.argoproj.io/reconcile-at"
+
+// CommitStatusDeAggregationLabel is the label used to identify commit statuses that make up the aggregated active commit status
+const CommitStatusDeAggregationLabel = "promoter.argoproj.io/commit-status-de-aggregation"
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -33,5 +33,5 @@ const PreviousEnvironmentCommitStatusKey = "promoter-previous-environment"
 // ReconcileAtAnnotation is the annotation used to indicate when the webhook triggered a reconcile
 const ReconcileAtAnnotation = "promoter.argoproj.io/reconcile-at"
 
-// CommitStatusDeAggregationAnnotation is the label used to identify commit statuses that make up the aggregated active commit status
-const CommitStatusDeAggregationAnnotation = "promoter.argoproj.io/previous-environment-statuses"
+// CommitStatusPreviousEnvironmentStatusesAnnotation is the label used to identify commit statuses that make up the aggregated active commit status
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/tidwall/gjson v1.18.0
 	go.uber.org/zap v1.27.0
+	gopkg.in/yaml.v2 v2.4.0
```
crenshaw-dev: Maybe go ahead with v3 instead?

internal/controller/promotionstrategy_controller.go
```diff
@@ -316,19 +330,30 @@ func (r *PromotionStrategyReconciler) createOrUpdatePreviousEnvironmentCommitSta
...
 				return fmt.Errorf("failed to create previous environments CommitStatus: %w", err)
 			}
+			return nil
 		}
-	} else if updatedCS.Spec.Phase != phase || updatedCS.Spec.Sha != ctp.Status.Proposed.Hydrated.Sha {
+		return fmt.Errorf("failed to get previous environments CommitStatus: %w", err)
+	}
+
+	updatedYamlStatusMap := make(map[string]string)
+	err = yaml.Unmarshal([]byte(updatedCS.Annotations[promoterv1alpha1.CommitStatusDeAggregationAnnotation]), &updatedYamlStatusMap)
```
crenshaw-dev: Is it safe to assume that this annotation is present?

api/v1alpha1/constants.go
```diff
@@ -32,3 +32,6 @@ const PreviousEnvironmentCommitStatusKey = "promoter-previous-environment"
 
 // ReconcileAtAnnotation is the annotation used to indicate when the webhook triggered a reconcile
 const ReconcileAtAnnotation = "promoter.argoproj.io/reconcile-at"
+
+// CommitStatusDeAggregationAnnotation is the label used to identify commit statuses that make up the aggregated active commit status
+const CommitStatusDeAggregationAnnotation = "promoter.argoproj.io/commit-status-de-aggregation"
```
crenshaw-dev:
```suggestion
const CommitStatusDeAggregationAnnotation = "promoter.argoproj.io/previous-environment-statuses"
```

Maybe something like that?

crenshaw-dev (changes requested):
Overall lgtm. Probably needs some docs. Maybe in the example CS manifest?

internal/controller/promotionstrategy_controller.go
zachaller: Yes we create this just above so it should always exist, unless some user messes with it :P

go.mod
zachaller: interesting not sure why it auto imported that, change

zachaller:
> Overall lgtm. Probably needs some docs. Maybe in the example CS manifest?

Was not sure

internal/controller/promotionstrategy_controller.go
zachaller: added a check still

crenshaw-dev (approved):

PR #217, opened 2025-03-19
fix: wrap pull request creation error

alexymantha:
Very minor fix, but I was trying to figure out why it was failing to open the request but the underlying error was hidden. I'm assuming this is just an oversight, I don't see a reason not to forward the error.


zachaller (approved):

zachaller:
was just missed, thanks for the contribution.

PR #224, opened 2025-03-20
feat: template pull requests title and body #119

### LINKED ISSUE #119: feat: allow templating of the PR title and description
We need to add support to template the PR title and description this should include access to the metadata file as well.

alexymantha:
Allows for templating the pull request title and body.

~~I added the templates to the GitRepository CRD because I feel like it makes sense to allow each repository to configure how they want the pull requests to be titled. I think this allows the most flexibility. I'm open to suggestion if you think that should live elsewhere.~~ See discussion below for more details

Here is an example of a PR that was created using a custom title template: https://github.com/alexymantha/gitops-test/pull/3.

Closes #119


crenshaw-dev:
Gonna chat w/ Zach and team about config location. But this is an exciting feature, thanks for the PR!

agaudreault:
Maybe it is missing a configMap to define the default in it. So the override per repo could be by the `GitRepository`, but it feels like it should be more part of the PromotionStrategy or another CRD because 2 different microservices might use the same repo, but want to have different PR titles. Another CRD specific for user config ("PromotionConfig") would offer that possible granularity.

alexymantha:
> Maybe it is missing a configMap to define the default in it.

I like that idea. I don't think there is any code currently to handle configmaps, I will create a separate issue to set that up. That will be needed for other configuration sooner or later.

> PromotionStrategy or another CRD because 2 different microservices might use the same repo, but want to have different PR titles. Another CRD specific for user config ("PromotionConfig") would offer that possible granularity.

That crossed my mind, but I'm not sure promotion strategies and pull requests are related enough for that to make sense in the `PromotionStrategy` CRD. `PromotionConfig` could make sense, as an in-between. I just think pull request titles are the kind of thing that are more likely to be enforced across an entire repo than at the promotion level if that makes sense. Or at least being able to configure a default template for a repository would be nice.

Let's say we create a new `PromotionConfig` resource, maybe a `GitRepository` could define a default `PromotionConfig`, which a `PromotionStrategy` can override by configuring its own `PromotionConfig`? Combined with a global ConfigMap, you could then configure the templates at 3 levels:
- Global: in the ConfigMap
- Repository: as a default PromotionConfig
- PromotionStrategy: as a PromotionConfig override

zachaller:
We talked a bit about this I think I personally like this seperation

`PromotionConfiguration` for end user controlled fields such as these templates. How this would work is if this is in the same namespace as the controller it would act as a global setting. I then think we allow end users to also install this into the same namespace as PromotionStrategy this acts as an override of the global configurations. I think we might want to add a reference to the PromotionStrategy spec (or GitRepo/Scm) so as to allow multiple PromotionStratgegies (etc) within the same namespace to have different configuration just like we would allow multiple PromotionStratagies within the same namespace to reference different Repos and hence Scm's.

I think we will then need another CRD for admin only controller configs for things like number or reconcilation threads and any form of security setting etc. Let's call this something like `ControllerConfiguration`. This would be a hard coded lookup by the controller for a specific named `ControllerConfiguration` probably controllable via a CLI flag with a default. We don't need this today because we currently have no usages.

We do have quite a few usages for the `PromotionConfiguration` however so that would be the priority.

crenshaw-dev:
My recommendation would be:

1) Use a new CRD for global config
2) Make all config hot-reloaded
3) Don't implement any non-global config yet - wait until we've collected enough use cases to design the config referential structure and precedence rules

zachaller:
If your unfamiliar with kubebuilder here is the command to add a new CRD. I think it's probably worth adding the controller even if it does nothing.

```
kubebuilder create api --group promoter --version v1alpha1 --kind PromotionConfiguration
INFO Create Resource [y/n]                        
y
INFO Create Controller [y/n]                      
y
```

alexymantha:
Thanks everyone for the suggestions! So I created the `PromotionConfiguration` CR but only implemented the global version that will live in the controller's namespace and will apply to all promotions. As Michael suggested, I think we can expand to more specific configurations as in the future when it will be needed.

I can also create an issue for the admin-only `ControllerConfiguration` that was suggested by Zach

config/config/global_promotionconfiguration.yaml
```diff
@@ -0,0 +1,13 @@
...
+  labels:
+    app.kubernetes.io/name: promoter
+    app.kubernetes.io/part-of: promoter
+    app.kubernetes.io/managed-by: kustomize
+  name: global
+spec:
+  pullRequest:
+    template:
+      title: "Promote {{ trunc 5 .Status.Proposed.Dry.Sha }} to `{{ .Spec.ActiveBranch }}`"
+      body: "This PR is promoting the environment branch `{{ .Spec.ActiveBranch }}` which is currently on dry sha {{ .Status.Active.Dry.Sha }} to dry sha {{ .Status.Proposed.Dry.Sha }}."
```
agaudreault: I would go for `description` instead of `body`. 

```suggestion
      description: "This PR is promoting the environment branch `{{ .Spec.ActiveBranch }}` which is currently on dry sha {{ .Status.Active.Dry.Sha }} to dry sha {{ .Status.Proposed.Dry.Sha }}."
```

internal/controller/changetransferpolicy_controller.go
```diff
@@ -342,6 +344,16 @@ func (r *ChangeTransferPolicyReconciler) creatOrUpdatePullRequest(ctx context.Co
...
 
+	promotionConfig := &promoterv1alpha1.PromotionConfiguration{}
+	if err := r.Get(ctx, client.ObjectKey{Name: r.Config.GlobalPromotionConfigurationName, Namespace: r.Namespace}, promotionConfig); err != nil {
+		return fmt.Errorf("failed to get global promotion configuration: %w", err)
+	}
+
+	title, body, err := TemplatePullRequest(&promotionConfig.Spec.PullRequest, ctp)
+	if err != nil {
+		return fmt.Errorf("failed to template pull request: %w", err)
+	}
```
agaudreault: Can we isolate that in a new package even if it is the only usage for now. We kind of know that the logic to resolve the PromotionConfiguration object will get more complex in the future.
```
settingsMgr := settings.NewSettingsManager(globalPromotionConfigurationName, namespace)
promotionConfig, err := settingsMgr.GetPromotionConfiguration()
```

alexymantha: Done!

internal/controller/changetransferpolicy_controller.go
```diff
@@ -473,3 +485,17 @@ func (r *ChangeTransferPolicyReconciler) mergePullRequests(ctx context.Context,
 
 	return nil
 }
+
+func TemplatePullRequest(prc *promoterv1alpha1.PullRequestConfiguration, ctp *promoterv1alpha1.ChangeTransferPolicy) (string, string, error) {
+	title, err := utils.RenderStringTemplate(prc.Template.Title, ctp)
```
crenshaw-dev: Maybe we should nest the CTP in case we want to later add other top-level config keys? @zachaller 

```suggestion
	title, err := utils.RenderStringTemplate(prc.Template.Title, map[string]any{"ctp": ctp})
```

internal/settings/manager.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+	"fmt"
+
+	promoterv1alpha1 "github.com/argoproj-labs/gitops-promoter/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ManagerConfig struct {
+	GlobalNamespace                  string
+	GlobalPromotionConfigurationName string
```
crenshaw-dev: Maybe we should just hard-code a constant? I can't think of a reason someone would need to change this.

crenshaw-dev (changes requested):

internal/controller/changetransferpolicy_controller.go
zachaller: Agree, because we will want to add the `hydrator.metadata` as context I think eventually as well. But I would add that support in a follow up PR

zachaller: is it not already nested under `ctp`? Or are you think more than that.

internal/settings/manager.go
zachaller: I want to in the future add instance-id support (for many reasons) which means you might be running multiple controllers in one namespace.

crenshaw-dev: Constant for now, configurable when we add instance ID? Seems better to code it all at once.

zachaller: That's fair

internal/controller/changetransferpolicy_controller.go
alexymantha: Good idea, it's updated

internal/settings/manager.go
alexymantha: Moved it to a constant

alexymantha:
I think I've addressed all comments, anything else I should update?

internal/settings/manager.go
```diff
@@ -0,0 +1,38 @@
+package settings
```
crenshaw-dev, line 1: Do we even need a "Manager" abstraction? The config is just a k8s object, what if we just lean into that and do k8s client "gets" where appropriate to pull the current config?

cc: @zachaller

crenshaw-dev (changes requested):
Not to bikeshed, it just happens that your feature introduces really interesting design questions @alexymantha  :-D

internal/settings/manager.go
alexymantha: Originally I was just doing a Get but @agaudreault brought a good point (https://github.com/argoproj-labs/gitops-promoter/pull/224#discussion_r2007998761) that we will probably be doing some processing on the config such as merging it with a more specific config. But I do agree that it's an unnecessary abstraction until then.

crenshaw-dev: Yeah I feel like I'd rather remove the abstraction until we have the use case to design it around. @agaudreault is that alright?

agaudreault: I think the struct allows to centralize the code on how to get the settings. This is very likely to be refactored in the future, so having the code centralized will make it easier to refactor, and allows to mock/abstract the config logic in the unit tests. 

In doesn't matter if the config is one k8s object, multiple ones, or anything more complex using informers behind the scene. In the end, the promotion config is a `PromotionConfiguration` object and this class abstracts it.

agaudreault (approved):
LGTM

internal/settings/manager.go
crenshaw-dev: > allows to mock/abstract the config logic in the unit tests

Mock APIs are available out of the box for client-go, so I think this would actually make mocks more difficult.

I'm worried that future-us will look at the already half-built abstraction and simply choose to operate within it rather than being open to other ways of handling config. In other words, I'd rather avoid building the design until we have the use case to design around.

crenshaw-dev: Chatted, we can worry about it later. Let's ship it

crenshaw-dev (approved):

PR #243, opened 2025-04-03
feat: remove active and proposed commit from spec

zachaller:


README.md
```diff
@@ -25,18 +25,17 @@ metadata:
...
-  activeCommitStatuses:
+  checks:
     - key: argocd-app-health
-  proposedCommitStatuses:
     - key: security-scan
   environments:
     - branch: environment/dev
     - branch: environment/test
+      checks:
+        - key: performance-test
```
crenshaw-dev, line 37: Should we take the opportunity to standardize on "checks that block this environment" instead of "checks evaluated on this environment"? i.e. move performance-test to prod. 

I'm not sold on the change, but now would be the time to consider it.

api/v1alpha1/controllerconfiguration_types.go
```diff
@@ -26,6 +26,14 @@ import (
 // ControllerConfigurationSpec defines the desired state of ControllerConfiguration.
 type ControllerConfigurationSpec struct {
 	PullRequest PullRequestConfiguration `json:"pullRequest,omitempty"`
+	// ActiveCommitStatuses lists the statuses to be monitored on the active branch
```
crenshaw-dev, line 29:
```suggestion
	// ActiveCommitStatuses lists the statuses to be monitored on active branches
```

README.md
zachaller: I have been talking about that with Alex and Leo, it's really awkward to think about with it just being `checks`.  I think by requiring a end users to have to know if they are active vs proposed and making them place active on the gated environment is awkward without a more friendly spec to help guide.

We either need to say put checks on the environment they are ran against and don't worry about how they are used in promotion or we need a spec that leaks that they are different.

```
kind: PromotionStrategy
metadata:
  name: argocon-demo
spec:
  checks:
    - key: argocd-health
    - key: yaml-lint
  environments:
    - branch: environments/development
    - branch: environments/staging
      checks:
        - key: performance-test
    - branch: environments/production
      checks:
        - key: deployment-freeze
---
kind: ControllerConfiguration
metadata:
  name: promoter-controller-configuration
spec:
  activeCommitStatuses:
  - key: performance-test
  - key: argocd-health
```

vs

```
kind: PromotionStrategy
metadata:
  name: argocon-demo
spec:
  checks:
    - key: argocd-health
    - key: yaml-lint
  environments:
    - branch: environments/development
    - branch: environments/staging
    - branch: environments/production
       gates:
        - key: deployment-freeze
      previouseEnvironmentGates:
        - key: performance-test
```

zachaller: I think in the past when we had active and proposed it was awkward to have them on the environment they where ran against vs being used as a gate. I think if we go back to leaking that info, moving it to "gates" makes that leak cleaner. 

I do think that we need to pick a side though to either share that detail in the spec and be good with it or just tell end users, hey we reconcile this against this environment and all the magic that happens don't worry about it we will continuously run these checks and report a status an said environment. 

The awkwardness here though is things like performance tests etc will generally only run post merge and that behavior is not exposed, but that is the point I guess. :P

zachaller: Yea we have been chatting about this internally and have mixed opinions.

zachaller: This PR changes behavior: https://github.com/zachaller/gitops-promoter/pull/3

PR #246, opened 2025-04-08
feat: add max payload size option for webhook reciever

robinlieb:
Add Option to to set a max payload size for the webhook reciever. The default is set to 25 MB, which is the current cap of Github (and GitLab). Could probably also reduce that to a smaller number, since the expected payload size should be much smaller.

Additionally returned an error when `beforeSha` is empty, which results in an HTTP 500 instead of a 200 when this property is missing.


internal/webhookreceiver/server_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func TestPostRootMaxPayload(t *testing.T) {
```
zachaller: Can we turn this into a Ginkgo test see [here](https://github.com/argoproj-labs/gitops-promoter/pull/161) for some context. We also let templating tests go through without ginkgo as well https://github.com/argoproj-labs/gitops-promoter/issues/248

zachaller: This is an example https://github.com/argoproj-labs/gitops-promoter/pull/249

internal/webhookreceiver/server.go
```diff
@@ -20,15 +20,35 @@ import (
 
 var logger = ctrl.Log.WithName("webhookReceiver")
 
+const (
+	defaultMaxWebhookPayloadSize = int64(25) * 1024 * 1024
```
crenshaw-dev: Should probably rename to `defaultMaxWebhookPayloadSizeBytes`.

internal/webhookreceiver/server.go
```diff
@@ -20,15 +20,35 @@ import (
...
+	}
+
+	for _, opt := range opts {
+		opt(&wr)
+	}
+
+	return wr
+}
+
+func WithMaxPayloadSize(size int64) Option {
```
crenshaw-dev: Should we make this configurable via the new ControllerConfiguration CRD? Maybe like this:

```
spec:
  webhook:
    maxPayloadSizeBytes: 0
```

internal/webhookreceiver/server.go
```diff
@@ -20,15 +20,35 @@ import (
...
+const (
+	defaultMaxWebhookPayloadSize = int64(25) * 1024 * 1024
+)
+
 type webhookReceiver struct {
-	mgr       controllerruntime.Manager
-	k8sClient client.Client
+	mgr            controllerruntime.Manager
+	k8sClient      client.Client
+	maxPayloadSize int64
```
crenshaw-dev: Maybe go with a `*uint32` where 0 means "no limit" and nil means "default"?

32 should be enough, I can't imagine anyone needing to set a limit over 4GB.

crenshaw-dev (changes requested):

internal/webhookreceiver/server_test.go
robinlieb: Changed it, thanks for providing the references!

internal/webhookreceiver/server.go
robinlieb: Implemented it in the proposed way.
Also moved the default directly in the API spec.

internal/webhookreceiver/server.go
robinlieb: Removed the default here since moved in the API spec.

internal/webhookreceiver/server.go
robinlieb: Implemented it with the no limit when 0 is set. Default is provided as a kubebuilder default.

crenshaw-dev:
@robinlieb the way it's set up now, is the max payload size hot-loaded from config? I think we'd want that to be the default for all fields in the controller config CR.

robinlieb:
@crenshaw-dev currently there is not hot reload implemented. Good objection, I will have a look and update the PR accordingly.

config/config/controllerconfiguration.yaml
```diff
@@ -11,3 +11,5 @@ spec:
     template:
       title: "Promote {{ trunc 5 .ChangeTransferPolicy.Status.Proposed.Dry.Sha }} to `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}`"
       description: "This PR is promoting the environment branch `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}` which is currently on dry sha {{ .ChangeTransferPolicy.Status.Active.Dry.Sha }} to dry sha {{ .ChangeTransferPolicy.Status.Proposed.Dry.Sha }}."
+  webhook:
+    maxPayloadSize: 26214400
```
crenshaw-dev:
```suggestion
    maxPayloadSize: "25Mi"
```

config/samples/promoter_v1alpha1_controllerconfiguration.yaml
```diff
@@ -10,4 +10,5 @@ spec:
     template:
       title: "Promote {{ trunc 5 .ChangeTransferPolicy.Status.Proposed.Dry.Sha }} to `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}`"
       description: "This PR is promoting the environment branch `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}` which is currently on dry sha {{ .ChangeTransferPolicy.Status.Active.Dry.Sha }} to dry sha {{ .ChangeTransferPolicy.Status.Proposed.Dry.Sha }}."
-
+  webhook:
+    maxPayloadSize: 26214400
```
crenshaw-dev:
```suggestion
    maxPayloadSize: "25Mi"
```

docs/example-resources/ControllerConfiguration.yaml
```diff
@@ -7,6 +7,8 @@ spec:
     template:
       title: "Promote {{ trunc 5 .ChangeTransferPolicy.Status.Proposed.Dry.Sha }} to `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}`"
       description: "This PR is promoting the environment branch `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}` which is currently on dry sha {{ .ChangeTransferPolicy.Status.Active.Dry.Sha }} to dry sha {{ .ChangeTransferPolicy.Status.Proposed.Dry.Sha }}."
+  webhook:
+    maxPayloadSize: 26214400
```
crenshaw-dev:
```suggestion
    maxPayloadSize: "25Mi"
```

api/v1alpha1/controllerconfiguration_types.go
```diff
@@ -26,6 +27,7 @@ import (
 // ControllerConfigurationSpec defines the desired state of ControllerConfiguration.
 type ControllerConfigurationSpec struct {
 	PullRequest PullRequestConfiguration `json:"pullRequest,omitempty"`
+	Webhook     WebhookConfiguration     `json:"webhook,omitempty"`
```
crenshaw-dev:
```suggestion
	// +required
	Webhook     WebhookConfiguration     `json:"webhook,omitempty"`
```

api/v1alpha1/controllerconfiguration_types.go
```diff
@@ -56,6 +58,13 @@ type PullRequestTemplate struct {
 	Description string `json:"description,omitempty"`
 }
 
+type WebhookConfiguration struct {
+	// Maximum allowed payload size in bytes.
+	// The default value is 25Mi.
+	// Set to 0 for no limit on the payload size.
+	MaxPayloadSize *resource.Quantity `json:"maxPayloadSize,omitempty"`
```
crenshaw-dev:
```suggestion
	// +required
	MaxPayloadSize resource.Quantity `json:"maxPayloadSize,omitempty"`
```

internal/settings/manager.go
```diff
@@ -17,6 +18,8 @@ const (
 	defaultArgoCDCommitStatusRequeueDuration   = 15 * time.Second
 )
 
+var defaultWebhookMaxPayloadSize = resource.MustParse("25Mi")
```
crenshaw-dev: Instead of defaulting here, let's make the field required and set the default via the install manifests.

crenshaw-dev (changes requested):
Sorry to ask for more changes, we're coming up with the config strategy as we go.

internal/settings/manager.go
robinlieb: Changed it.

PR #250, opened 2025-04-09
chore: simplify SHA fetching, better errors

crenshaw-dev:
Instead of looping over multiple SHAs, `GetBranchShas` now takes just one.

This simplifies the function and forces us to handle errors for different calls separately. Instead of getting an error saying one of two possible branches didn't have a DRY sha, you'll get an error specific to the failing branch.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	_ "embed"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


internal/git/git.go
```diff
@@ -101,67 +100,55 @@ type BranchShas struct {
...
 
-		if _, ok := shaMap[branch]; !ok {
-			shaMap[branch] = &BranchShas{}
-		}
-		shaMap[branch].Hydrated = strings.TrimSpace(stdout)
-		logger.V(4).Info("Got hydrated branch sha", "branch", branch, "sha", shaMap[branch].Hydrated)
-
-		// TODO: safe path join
-		metadataFile := g.pathLookup.Get(g.gap.GetGitHttpsRepoUrl(*g.gitRepo)+g.pathContext) + "/hydrator.metadata"
-		if _, err := os.Stat(metadataFile); err != nil {
```
crenshaw-dev, line 143: I removed this initial stat check, os.Open should throw the same error.

PR #271, opened 2025-04-21
fix: webhook status no-content

FourFifthsCode:
- **remove error response and replace with no content, change log for error as its more of a warning**


agaudreault (approved):
LGTM

PR #283, opened 2025-04-28
feat(config)!: require all config fields, default in manifests insted of code

crenshaw-dev:
This is more explicit and easy for the user to understand.

Downside? is that all fields must be set for tests.


zachaller:
I personally like requring them, it also serves as a nice documentation spot

zachaller (approved):

Issue #281, opened 2025-04-26
Consider replacing `autoMerge` boolean with enum


crenshaw-dev:
Was reading through [k8s API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) and came across this:

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

It's a good point. Should we consider something like `mergePolicy`, `mergeStrategy`, etc.? Maybe `mergePolicy: OnCommitStatusPassed` and `mergePolicy: Manual`.

zachaller:
https://github.com/argoproj-labs/gitops-promoter/pull/199

Not exactly related but share the same spirit just in opposite direction. Maybe, the API naming should be consistent between them.

crenshaw-dev:
Discussed it, but eventually decided to just stick with the boolean.

PR #285, opened 2025-04-29
chore(promotionstrategy)!: add mergePolicy with autoMerge enum (#281)

crenshaw-dev:
Made it a block to accommodate policies about skipping PRs if certain files aren't present (https://github.com/argoproj-labs/gitops-promoter/pull/199/files).


api/v1alpha1/promotionstrategy_types.go
```diff
@@ -78,12 +81,19 @@ type Environment struct {
...
-// GetAutoMerge returns the value of the AutoMerge field, defaulting to true if the field is nil.
-func (e *Environment) GetAutoMerge() bool {
-	if e.AutoMerge == nil {
-		return true
-	}
-	return *e.AutoMerge
+type AutoMergePolicy string
+
+const (
+	AutoMergePolicyDefault               AutoMergePolicy = ""
```
zachaller, line 87: Whats the point of this default why not add a default to the spec. Is this different than OnCommitStatusSuccess?

aka
```
// +kubebuilder:default:=OnCommitStatusSuccess
```

api/v1alpha1/promotionstrategy_types.go
```diff
@@ -78,12 +81,19 @@ type Environment struct {
...
+	AutoMergePolicyOnCommitStatusSuccess AutoMergePolicy = "OnCommitStatusSuccess"
+	AutoMergePolicyNever                 AutoMergePolicy = "Never"
+)
+
+type MergePolicy struct {
+	// AutoMerge determines whether the dry commit should be automatically merged into the next branch in the sequence.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum:="";OnCommitStatusSuccess;Never
+	AutoMerge AutoMergePolicy `json:"autoMerge,omitempty"`
```
zachaller, line 96: Maybe just, auto, we are in a merge block already?

api/v1alpha1/promotionstrategy_types.go
crenshaw-dev: That's probably the way to go... I think I was trying to avoid some kind of Argo CD diff weirdness with default

crenshaw-dev:
For now: nah

PR #286, opened 2025-04-29
chore(refactor): simplify by enforcing consistent order

crenshaw-dev:
We currently rely on branch names to match environments to CTPs and to env statuses.

This PR switches to only relying on order, which significantly simplifies the code, but also means we have to be careful to enforce order and length consistency among slices.

The Reconcile loop starts by constructing a list of CTPs that exactly matches the length and order of ps.Spec.Environments.

Then it calculates the status, which starts by reconstructing the ps.Status.Environments list to match ps.Spec.Environments.

After that point, any code can use any of the three slices with confidence that they're all in sync.

### DIFF
--- a/internal/controller/promotionstrategy_controller.go
+++ b/internal/controller/promotionstrategy_controller.go
@@ -284,7 +284,7 @@ func (r *PromotionStrategyReconciler) setEnvironmentCommitStatus(targetStatus *p
 	}
 }
 
-func (r *PromotionStrategyReconciler) createOrUpdatePreviousEnvironmentCommitStatus(ctx context.Context, ctp *promoterv1alpha1.ChangeTransferPolicy, phase promoterv1alpha1.CommitStatusPhase, previousEnvironmentStatus *promoterv1alpha1.EnvironmentStatus, previousCRPCSPhases []promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase) error {
+func (r *PromotionStrategyReconciler) createOrUpdatePreviousEnvironmentCommitStatus(ctx context.Context, ctp *promoterv1alpha1.ChangeTransferPolicy, phase promoterv1alpha1.CommitStatusPhase, previousEnvironmentBranch string, previousCRPCSPhases []promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase) error {
 	// TODO: do we like this name proposed-<name>?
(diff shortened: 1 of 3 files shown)


internal/controller/promotionstrategy_controller_test.go
```diff
@@ -312,30 +312,30 @@ var _ = Describe("PromotionStrategy Controller", func() {
 				g.Expect(promotionStrategy.Status.Environments).To(HaveLen(3))
 
 				// By("Checking that the ChangeTransferPolicy for development has shas that are not empty, meaning we have reconciled the resource")
-				// Active dry sha should be empty because hydrator.metadata is not present
+				// Active dry sha should be the same as proposed dry sha because we should have merged the branch with hydrator.metadata.
```
crenshaw-dev, line 315: idk why this test was passing in the first place... iiuc if hydrator.metadata is missing on the active branch, we should treat it the same as an outdated SHA and merge the proposed branch in.

I've changed the tests to assume we go ahead and merge the proposed branch (with hydrator.metadata) into the active branch.

internal/controller/promotionstrategy_controller.go
```diff
@@ -214,54 +211,54 @@ func (r *PromotionStrategyReconciler) upsertChangeTransferPolicy(ctx context.Con
...
-		}())
-
-		i, environmentStatus := utils.GetEnvironmentStatusByBranch(*ps, environment.Branch)
+		}
+	}
+	ps.Status.Environments = environmentStatuses
+
+	for i, ctp := range ctps {
+		// Update fields individually to avoid overwriting existing fields.
+		ps.Status.Environments[i].Branch = ctp.Spec.ActiveBranch
```
crenshaw-dev, line 234: As far as I could tell, there was no need to construct a new `CommitShaState`, since the struct only contains two fields anyway. So I just copy the one that's there.

internal/controller/promotionstrategy_controller.go
```diff
@@ -296,11 +293,6 @@ func (r *PromotionStrategyReconciler) createOrUpdatePreviousEnvironmentCommitSta
 	gvk := promoterv1alpha1.GroupVersion.WithKind(kind)
 	controllerRef := metav1.NewControllerRef(ctp, gvk)
 
-	branch := "no previous environment"
-	if previousEnvironmentStatus != nil {
```
crenshaw-dev, line 300: The nil case could never be hit. We don't call this function unless there's a previous env.

zachaller (approved):

PR #290, opened 2025-04-30
chore: refactor pull request controller

zachaller:
Simplify and refactor pull request controller logic switch to switch statement instead of if blocks

### DIFF
--- a/internal/controller/pullrequest_controller.go
+++ b/internal/controller/pullrequest_controller.go
@@ -115,11 +115,6 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 				if err := r.createPullRequest(ctx, &pr, provider); err != nil {
 					return ctrl.Result{}, fmt.Errorf("failed to create pull request: %w", err)
 				}
-			} else {
-				logger.Info("Updating PullRequest")
-				if err := r.updatePullRequest(ctx, pr, provider); err != nil {
-					return ctrl.Result{}, fmt.Errorf("failed to update pull request: %w", err)
-				}
 			}
 		} else if pr.Spec.State == promoterv1alpha1.PullRequestMerged {
 			logger.Info("Merging PullRequest")
@@ -141,6 +136,11 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			}
 			// No need to update the status here, as the pull request is deleted
 			return ctrl.Result{}, nil
+		}
+	} else {
+		logger.Info("Updating PullRequest")
+		if err := r.updatePullRequest(ctx, pr, provider); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update pull request: %w", err)
 		}
 	}
 
--- a/internal/controller/pullrequest_controller.go
+++ b/internal/controller/pullrequest_controller.go
@@ -79,7 +79,7 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	if pr.Status.State == promoterv1alpha1.PullRequestMerged || pr.Status.State == promoterv1alpha1.PullRequestClosed {
-		logger.Info("Cleaning up close|merged pull request", "pullRequestID", pr.Status.ID)
+		logger.Info("Cleaning up close and merged pull request", "pullRequestID", pr.Status.ID)
 		if err := r.Delete(ctx, &pr); err != nil && !errors.IsNotFound(err) {
 			logger.Error(err, "Failed to delete PullRequest")
 			return ctrl.Result{}, fmt.Errorf("failed to delete PullRequest: %w", err)
--- a/internal/controller/pullrequest_controller.go
+++ b/internal/controller/pullrequest_controller.go
@@ -122,8 +122,6 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			// We do not need to update status here because we close the PR on the SCM but we requeue the reconcile to
 			// clean up the PullRequest object
 			return ctrl.Result{Requeue: true}, nil
-		default:
-			return ctrl.Result{}, fmt.Errorf("unknown pull request state: %s", pr.Spec.State)
 		}
 	}
 
--- a/internal/controller/pullrequest_controller.go
+++ b/internal/controller/pullrequest_controller.go
@@ -58,6 +58,7 @@ type PullRequestReconciler struct {
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.2/pkg/reconcile
 func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
+	logger.Info("Reconciling PullRequest")
 
 	var pr promoterv1alpha1.PullRequest
 	if err := r.Get(ctx, req.NamespacedName, &pr); err != nil {
@@ -77,6 +78,7 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
+	logger.Info("Checking for open PR on provider")
 	found, err := provider.FindOpen(ctx, &pr)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to check for open PR: %w", err)
@@ -85,6 +87,7 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	if !found && pr.Status.State != "" {
 		logger.Info("Deleting Pull Request, no open PR found on provider")
 		if err := r.Delete(ctx, &pr); err != nil && !errors.IsNotFound(err) {
+			logger.Error(err, "Failed to delete PullRequest")
 			return ctrl.Result{}, fmt.Errorf("failed to delete PullRequest: %w", err)
 		}
 		return ctrl.Result{}, nil
@@ -95,15 +98,19 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
+	logger.Info("Reconciling PullRequest state", "desired", pr.Spec.State, "current", pr.Status.State)
 	switch pr.Spec.State {
 	case promoterv1alpha1.PullRequestOpen:
 		if pr.Status.State != promoterv1alpha1.PullRequestOpen {
+			logger.Info("Creating PullRequest")
 			if err := r.createPullRequest(ctx, &pr, provider); err != nil {
+				logger.Error(err, "Failed to create pull request")
 				return ctrl.Result{}, fmt.Errorf("failed to create pull request: %w", err)
 			}
 		}
 	case promoterv1alpha1.PullRequestMerged:
 		if pr.Status.State != promoterv1alpha1.PullRequestMerged {
+			logger.Info("Merging PullRequest")
 			if err := r.mergePullRequest(ctx, &pr, provider); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to merge pull request: %w", err)
 			}
@@ -111,6 +118,7 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	case promoterv1alpha1.PullRequestClosed:
 		if pr.Status.State != promoterv1alpha1.PullRequestClosed {
+			logger.Info("Closing PullRequest")
 			if err := r.closePullRequest(ctx, &pr, provider); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to close pull request: %w", err)
 			}
--- a/internal/controller/pullrequest_controller.go
+++ b/internal/controller/pullrequest_controller.go
@@ -104,7 +104,6 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		if pr.Status.State != promoterv1alpha1.PullRequestOpen {
 			logger.Info("Creating PullRequest")
 			if err := r.createPullRequest(ctx, &pr, provider); err != nil {
-				logger.Error(err, "Failed to create pull request")
 				return ctrl.Result{}, fmt.Errorf("failed to create pull request: %w", err)
 			}
 		}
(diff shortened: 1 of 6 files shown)


internal/controller/pullrequest_controller.go
```diff
@@ -60,119 +60,79 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 		}
-
-		logger.Error(err, "failed to get PullRequest", "namespace", req.Namespace, "name", req.Name)
 		return ctrl.Result{}, fmt.Errorf("failed to get PullRequest: %w", err)
 	}
 
-	pullRequestProvider, err := r.getPullRequestProvider(ctx, pr)
-	if err != nil {
+	provider, err := r.getPullRequestProvider(ctx, pr)
+	if err != nil || provider == nil {
```
crenshaw-dev: Instead of the double check, just return an error from `getPullRequestProvider`?

internal/controller/pullrequest_controller.go
```diff
@@ -60,119 +60,79 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	logger := log.FromContext(ctx)
 
 	var pr promoterv1alpha1.PullRequest
-	err := r.Get(ctx, req.NamespacedName, &pr, &client.GetOptions{})
-	if err != nil {
+	if err := r.Get(ctx, req.NamespacedName, &pr); err != nil {
```
crenshaw-dev, line 65: Inline checks break my brain a little, especially if it's more than just the nil check. Feel strongly about inlining it?

internal/controller/pullrequest_controller.go
```diff
@@ -60,119 +60,79 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	if pr.Spec.State == promoterv1alpha1.PullRequestClosed && pr.Status.State != promoterv1alpha1.PullRequestClosed {
-		err := r.closePullRequest(ctx, &pr, pullRequestProvider)
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to close pull request: %w", err)
+	case promoterv1alpha1.PullRequestMerged:
+		if pr.Status.State != promoterv1alpha1.PullRequestMerged {
+			if err := r.mergePullRequest(ctx, &pr, provider); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to merge pull request: %w", err)
+			}
+			return ctrl.Result{Requeue: true}, nil
```
crenshaw-dev: Shouldn't we update the status here? Seems like we'd fall into an endless reconcile loop.

internal/controller/pullrequest_controller.go
zachaller: I like them but I am also fairly impartial

internal/controller/pullrequest_controller.go
zachaller: Because we close the PR on the SCM provider, we will clean up resource on the next reconcile I added comments.

zachaller: I go back and forth on it, fyi, it's not needed logically it save a k8s update call removing it. But, we lose the very very small window where status is merged.

internal/controller/pullrequest_controller.go
```diff
@@ -58,121 +58,92 @@ type PullRequestReconciler struct {
...
+		if err := r.Delete(ctx, &pr); err != nil && !errors.IsNotFound(err) {
+			logger.Error(err, "Failed to delete PullRequest")
 			return ctrl.Result{}, fmt.Errorf("failed to delete PullRequest: %w", err)
 		}
 		return ctrl.Result{}, nil
 	}
 
-	// If we have a status state, and the status state is the same as the spec state, and the observed generation is the same as the generation, we don't need to reconcile.
-	if pr.Status.State != "" && pr.Spec.State == pr.Status.State && pr.Status.ObservedGeneration == pr.Generation {
+	if pr.Status.State == pr.Spec.State && pr.Status.ObservedGeneration == pr.Generation {
```
crenshaw-dev: gitRepositoryRef, title, and description are all mutable... do we need to confirm that status matches spec for those as well?

internal/controller/pullrequest_controller.go
```diff
@@ -58,121 +58,92 @@ type PullRequestReconciler struct {
...
-	found, err := pullRequestProvider.FindOpen(ctx, &pr)
+	logger.Info("Checking for open PR on provider")
+	found, err := provider.FindOpen(ctx, &pr)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to check for open PR: %w", err)
 	}
 
-	// We can't find an open PR on the provider, and we have had a status state before, we should now delete it because
-	// it no longer exists on provider.
 	if !found && pr.Status.State != "" {
```
crenshaw-dev: We should probably add docs on the `State` field to clarify that it should only be populated based on a response from the SCM confirming the actual state. Or we could switch to something that less likely to be accidentally populated, like `ID`. Both are vanishingly unlikely to be misused. But probably worth clarifying to avoid subtle bugs.

We should probably also document that the PullRequest lifecycle is strictly tied to the SCM PullRequest. If the user closes or merges the PR, we delete the PullRequest resource. That documentation could probably just go on this package instead of in docs/.

internal/controller/pullrequest_controller.go
```diff
@@ -58,121 +58,92 @@ type PullRequestReconciler struct {
...
-		logger.Info("Deleting Pull Request, because no open PR found on provider")
-		err = r.Delete(ctx, &pr)
-		if err != nil {
-			if errors.IsNotFound(err) {
-				return ctrl.Result{}, nil
-			}
+		logger.Info("Deleting Pull Request, no open PR found on provider")
```
crenshaw-dev:
```suggestion
		logger.Info("Deleting PullRequest - the corresponding SCM PR was previously identified but is no longer open", "pullRequestID", pr.Status.ID)
```

internal/controller/pullrequest_controller.go
crenshaw-dev: I could definitely imagine the Promoter updating title/description based on various state (though I can't think of an immediate use case).

To save space we could persist a title hash and description hash in the status.

internal/controller/pullrequest_controller.go
```diff
@@ -58,121 +58,92 @@ type PullRequestReconciler struct {
...
-		err = r.mergePullRequest(ctx, &pr, pullRequestProvider)
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to merge pull request: %w", err)
-		}
-
-		err = r.Status().Update(ctx, &pr)
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("merged pull request in the SCM, but failed to update PullRequest status: %w", err)
+	logger.Info("Reconciling PullRequest state", "desired", pr.Spec.State, "current", pr.Status.State)
+	switch pr.Spec.State {
```
crenshaw-dev: I think it would be more readable to invert the switch/ifs... i.e. remove the inner if conditions and wrap the whole switch statement in a single `if pr.Spec.State != pr.Status.State`.

internal/controller/pullrequest_controller.go
crenshaw-dev: Zach pointed out we don't delete until a reconcile, and we don't really reconcile on a regular interval. Just good to document whatever our current choice is.

internal/controller/pullrequest_controller.go
crenshaw-dev: Actually so the generation check resolves title/description.

crenshaw-dev:
```suggestion
	// If there hasn't been an SCM change to PR state, and there also hasn't been a spec change,
	// quit early. It's only possible to hit this condition if the PR is open on the SCM. Otherwise the
	// CR would have been deleted and reconciliation short-circuited.
	if pr.Status.State == pr.Spec.State && pr.Status.ObservedGeneration == pr.Generation {
```

internal/controller/pullrequest_controller.go
crenshaw-dev: There's something nice about all code paths updating the status... logic at the top of the reconcile could be super simple. If `status.state == "closed" || status.state == "merged"`, immediately delete. Saves the SCM call and might be more readable.

But meh, wouldn't spend an hour on it, either way is fine.

internal/controller/pullrequest_controller.go
```diff
@@ -58,121 +58,92 @@ type PullRequestReconciler struct {
...
-		// but is cleaner than deleting right away. Will revisit this.
-		return ctrl.Result{Requeue: true}, nil
 	}
 
 	if pr.Status.ObservedGeneration != pr.Generation {
-		// TODO: is there more work avoidance we can do here? Just because the generation has changed, do we know that
-		//       we need to hit the SCM? For example, an annotation update probably shouldn't cause an API call.
-		err = r.updatePullRequest(ctx, pr, pullRequestProvider)
-		if err != nil {
+		if err := r.updatePullRequest(ctx, pr, provider); err != nil {
```
crenshaw-dev, line 141: I think we hit this line on an initial PR opening, causing an unnecessary SCM call. Maybe also check `pr.Status.ObservedGeneration != ""` to detect a newly-created CR? I don't like the obscurity of that check, but would solve the problem.

internal/controller/pullrequest_controller.go
```diff
@@ -186,160 +157,84 @@ func (r *PullRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-			controllerutil.RemoveFinalizer(pr, finalizerName)
-			if err = r.Update(ctx, pr); err != nil {
-				return true, fmt.Errorf("failed to remove finalizer: %w", err)
-			}
-			return true, nil
+	} else if controllerutil.ContainsFinalizer(pr, finalizer) {
+		if err := r.closePullRequest(ctx, pr, provider); err != nil {
+			return false, fmt.Errorf("failed to close pull request: %w", err)
+		}
+		controllerutil.RemoveFinalizer(pr, finalizer)
```
crenshaw-dev: If the following line (update the CR to remove the finalizer) fails, will we fall into an endless loop of trying to close an already-closed PR? Not sure how `closePullRequest` behaves if the PR is already closed.

Could either ensure `closePullRequest` returns a nil error if the PR is already closed or could nest `r.Update` in a retry loop to mitigate failures.

internal/controller/pullrequest_controller.go
zachaller: I went down this route originally way way way way back in very early commits because I agree, the idea of the controller to match the state of the world at the top of the reconcile loop means we should not remove that SCM call it's doing the equivalent of getting the resource from k8s also at the top of the reconcile. The source of truth is the SCM.

internal/controller/pullrequest_controller.go
zachaller: Yea I don't like it either going to spend some time looking at removing the returns again.

internal/controller/pullrequest_controller.go
crenshaw-dev: True, but I think in all cases the status field is a direct reflection of the SCM state. I guess it's possible for a PR to be reopened on the SCM before the next reconcile, but its unlikely.

crenshaw-dev: Wrote out a truth table

| generation changed | spec/status drifted | call SCM | update status |
|-----|-----|-----|----|
| no | no | no | no |
| no | yes | yes | yes |
| yes | yes | yes | yes |
| yes | no | no | yes|

Based on that, I think this would be a decent pseudo-code:

```
// If status is merged or closed, delete PullRequest and short-circuit.

// Fetch new status from SCM.

if observedGeneration == generation && status == spec {
    // No drift, short-circuit.
}

if status != spec {
    if spec.state == "open" {
        if status.id == "" {
            // create the PR
        } else {
            // update the PR
        }
    } else if spec.state == "merged" {
         // merge the PR
    } else if spec.state == "closed" {
        // close the PR
    }
}

// In all cases, set the observedGeneration (which may not have actually changed) and update the status.

```

The differences from current implementation are:

1) Add initial step to clean up closed/merged PRs
2) Always update the status if there's drift (either generation changed or status != spec)
3) Only update the PR if spec.state is open, and if we don't already have a PR ID.

The nesting gets a little deep, but I think the simplicity makes it easy enough to follow and cuts down the number of API calls.

zachaller: What if the PR is merged or closed on the SCM?

zachaller: I think one way of saying it, is we should do what we do in other controllers and calculate a status at the top of the reconcile.

internal/controller/pullrequest_controller.go
```diff
@@ -58,121 +58,101 @@ type PullRequestReconciler struct {
...
+		} else if pr.Spec.State == promoterv1alpha1.PullRequestMerged {
+			logger.Info("Merging PullRequest")
+			if err := r.mergePullRequest(ctx, &pr, provider); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to merge pull request: %w", err)
+			}
+			if err := r.Client.Delete(ctx, &pr); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to delete PullRequest: %w", err)
+			}
+			// No need to update the status here, as the pull request is deleted
+			return ctrl.Result{}, nil
```
crenshaw-dev: I think we can drop this now. The k8s call is cheap, and it'll make sure that on the next reconciliation we short-circuit before FindOpen.

internal/controller/pullrequest_controller.go
```diff
@@ -58,121 +58,101 @@ type PullRequestReconciler struct {
...
-	if deleted {
+
+	if pr.Status.State == promoterv1alpha1.PullRequestMerged || pr.Status.State == promoterv1alpha1.PullRequestClosed {
+		logger.Info("Cleaning up close and merged pull request", "pullRequestID", pr.Status.ID)
+		if err := r.Delete(ctx, &pr); err != nil && !errors.IsNotFound(err) {
+			logger.Error(err, "Failed to delete PullRequest")
+			return ctrl.Result{}, fmt.Errorf("failed to delete PullRequest: %w", err)
+		}
 		return ctrl.Result{}, nil
 	}
```
crenshaw-dev: Since we do Client.Delete after merge/close, we don't need this logic. We'll hit the finalizer logic instead.

internal/controller/pullrequest_controller.go
```diff
@@ -58,121 +59,99 @@ type PullRequestReconciler struct {
...
 	pr.Status.ObservedGeneration = pr.Generation
-	err = r.Status().Update(ctx, &pr)
-	if err != nil {
+	if err := r.Status().Update(ctx, &pr); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to update PullRequest status: %w", err)
 	}
 
 	logger.Info("no known state transitions needed", "specState", pr.Spec.State, "statusState", pr.Status.State)
-	return ctrl.Result{}, nil
+	return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
```
zachaller, line 152: Will follow up with second PR to add this to settings

crenshaw-dev (approved):

PR #291, opened 2025-04-30
chore: more linting

crenshaw-dev:
Added everything, removed the ones I thought were silly, commented out a few that I like but that need more work.

Not really necessary, I just find refactoring relaxing. :-P

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,52 +3,112 @@ run:
   allow-parallel-runners: true
 linters:
   enable:
+    - asasalint
+    - asciicheck
+    - bidichk
     - bodyclose
     - canonicalheader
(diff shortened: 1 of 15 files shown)


.golangci.yml
```diff
@@ -3,52 +3,112 @@ run:
...
+    - usetesting
     - wastedassign
     - whitespace
     - wrapcheck
+    - zerologlint
   settings:
     copyloopvar:
       check-alias: true
-    goconst:
-      ignore-strings: "true"
```
crenshaw-dev, line 47: This was invalid config and is no longer used anyway.

crenshaw-dev:
Gonna do this piecemeal instead

PR #294, opened 2025-05-01
feat: add enum validation for PR status state

crenshaw-dev:
Probably lower value than the spec field, but still nice to be explicit.


crenshaw-dev:
I wonder if the "" is required if it's an optional field... probably not?

zachaller (approved):

PR #295, opened 2025-05-02
feat: add merge keys for lists

crenshaw-dev:
Worth discussing, not sure we need it yet. It would allow strategic merge patches of individual list items.

It would also enforce no duplicate status keys.


zachaller:
Yea we should have this, any reason it's not ready

crenshaw-dev:
@zachaller good to go!

zachaller (approved):

PR #305, opened 2025-05-07
fix: Update binary name in Dockerfile

### LINKED ISSUE #304: binary name is not consistent between Dockerfiles
When building locally with `make docker-build` the file `Dockerfile` is used and is configured with `/manager` as the entrypoint.
But in the k8s deployment manifest, the command is set to another binary name : `/gitops-promoter` (cf. https://github.com/argoproj-labs/gitops-promoter/pull/151)
However, it works for releases because the `Dockerfile.release` is used and configured with `/gitops-promoter` as the entrypoint

TL;DR binary name should be the same in both Dockerfiles

yogeek:
fixes #304

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-# Build the manager binary
+# Build the gitops-promoter binary
 FROM golang:1.24 AS builder
 ARG TARGETOS
 ARG TARGETARCH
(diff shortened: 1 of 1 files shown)


crenshaw-dev:
@zachaller any concerns about deviating from the kubebuilder defaults? Any reason we'd want to standardize on `manager` instead?

crenshaw-dev (approved):
Confirmed w/ Zach that this works

PR #309, opened 2025-05-09
feat: auto reconcile argocd commit status when applications update

zachaller:


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -261,47 +267,74 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
+		if err := c.Get(ctx, client.ObjectKeyFromObject(argoCDApplication), un, &client.GetOptions{}); err != nil {
+			return nil
+		}
+
+		var application argocd.ArgoCDApplication
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(un.Object, &application); err != nil {
+			return nil
+		}
+
+		appKey := application.GetNamespace() + "/" + application.GetName()
```
crenshaw-dev, line 292: struct instead of string? we'll wanna add multi-cluster support, good to avoid concerns about collisions and valid characters

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -261,47 +267,74 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
+			return nil
+		}
+
+		rwMutex.Lock()
+		revMap[appKey] = application.Status.Sync.Revision
+		rwMutex.Unlock()
+
+		var argoCDCommitStatusList promoterv1alpha1.ArgoCDCommitStatusList
+		if err := c.List(ctx, &argoCDCommitStatusList, &client.ListOptions{}); err != nil {
+			return nil
```
crenshaw-dev, line 310: Does silent failure make sense here? Maybe at least log?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -261,47 +267,74 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
+	revMap  = make(map[string]string)
+)
+
+func lookupArgoCDCommitStatusFromArgoCDApplication(c client.Client) func(ctx context.Context, argoCDApplication client.Object) []reconcile.Request {
+	return func(ctx context.Context, argoCDApplication client.Object) []reconcile.Request {
+		un := &unstructured.Unstructured{}
+		un.SetGroupVersionKind(gvk)
+
+		if err := c.Get(ctx, client.ObjectKeyFromObject(argoCDApplication), un, &client.GetOptions{}); err != nil {
+			return nil
```
crenshaw-dev, line 283: Does silent failure make sense here? If it's not a DoesNotExist, maybe at least log?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -261,47 +267,74 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
+					"application", appKey, "argocdcommitstatus", argoCDCommitStatus.Namespace+"/"+argoCDCommitStatus.Name)
+
+				return []reconcile.Request{{
+					NamespacedName: client.ObjectKeyFromObject(&argoCDCommitStatus),
+				}}
+			}
+		}
+
+		log.FromContext(ctx).Info("No ArgoCDCommitStatus found for ArgoCD application",
+			"namespace", argoCDApplication.GetNamespace(), "name", argoCDApplication.GetName())
```
```suggestion
		log.FromContext(ctx).Info("No ArgoCDCommitStatus found for ArgoCD application",
			"app-namespace", argoCDApplication.GetNamespace(), "application", argoCDApplication.GetName())
```

For consistency with Argo CD

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -261,47 +267,74 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
+		var argoCDCommitStatusList promoterv1alpha1.ArgoCDCommitStatusList
+		if err := c.List(ctx, &argoCDCommitStatusList, &client.ListOptions{}); err != nil {
+			return nil
+		}
+
+		// TODO: is there some way to do this without a loop? Can we use a field indexer? The one issue with field indexers is that
+		// they can not be used with lists (aka label selectors) so how else can we lookup.
+		for _, argoCDCommitStatus := range argoCDCommitStatusList.Items {
+			selector, err := metav1.LabelSelectorAsSelector(argoCDCommitStatus.Spec.ApplicationSelector)
+			if err == nil && selector.Matches(fields.Set(un.GetLabels())) {
```
crenshaw-dev, line 320: Should we handle the error? Log and skip? Something else?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -261,47 +267,74 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
+				log.FromContext(ctx).Info("ArgoCD application caused ArgoCDCommitStatus to reconcile",
+					"application", appKey, "argocdcommitstatus", argoCDCommitStatus.Namespace+"/"+argoCDCommitStatus.Name)
```
```suggestion
				log.FromContext(ctx).Info("ArgoCD application caused ArgoCDCommitStatus to reconcile",
					"app-namespace", argoCDApplication.GetNamespace(), "application", argoCDApplication.GetName(), "argocdcommitstatus", argoCDCommitStatus.Namespace+"/"+argoCDCommitStatus.Name)
```

For consistency with Argo CD

internal/controller/argocdcommitstatus_controller.go
zachaller: We are not erroring here, this means that we did not find an argocd application and so we should not request the commit status to be reconciled

crenshaw-dev: But this function is called in response to controller-runtime observing a change on a resource it thinks _does_ exist. So if we get an error that is DoesNotExist, it makes sense to silently move on, because clearly the app was just deleted. But if it's a different error, there might be a problem that we actually care about.

zachaller: added logs

internal/controller/argocdcommitstatus_controller.go
zachaller: we can log an error but we generally want to continue checking other applications in the list, the first one we find will cause the commit status to reconcile.

internal/controller/argocdcommitstatus_controller.go
zachaller: can we leave for now, I have a feeling a lot of this will change for multi cluster possibly

crenshaw-dev (approved):

PR #311, opened 2025-05-14
feat: add ClusterScmProvider CRD 

### LINKED ISSUE #308: Allow referencing SCMProvider from another namespace
Currently, all the resources need to be in the same namespace. I would like to be able to configure the SCMProvider and SCM secrets in a different namespace from where the GitRepository and PromotionStrategy are defined. 

What I am trying to do is allow end-users to configure their promotion strategies in their own namespace, but I do not want them to have access to the SCM secrets

```
team-a-namespace                 | promoter-system
- GitRepository: my-team-repo	 | - SCMProvider
- PromotionStrategy              | - Secret: credentials for the SCM
```

As far as I can tell, we should be able to just add a `namespace` field to the ref:

```yaml
kind: GitRepository
metadata:
  name: gitops-pdx-canada
  namespace: team-a-namespace
spec:
  gitlab:
    owner: my-org
    name: gitops
  scmProviderRef:
    name: github
    namespace: promoter-system
```
and then use it to fetch the SCMProvider

alexymantha:
Added the ClusterScmProvider CRD as discussed in the linked issue. I stayed pretty close to how it's done in ESO by using the same spec for both the namespaced and cluster version of the ScmProvider which also makes it possible to have a common interface to interact with (Cluster)ScmProviders. All the references to ScmProvider now go through the `GenericScmProvider` interface instead.

~~I made the following assumption: since both types share the same spec, there is now a `namespace` field in the `secretRef`. It will default to the same namespace as the ScmProvider if empty, and will throw an error if it's refering to another namespace when using the namespaced ScmProvider type. Another alternative is that we could remove the namespace field and only allow the secrets to be in the controller namespace when using ClusterScmProvider, but that limits its use quite a bit.~~

Closes #308


crenshaw-dev:
Thanks for the PR!

I'm not loving the dead `namespace` field in ScmProvider. Is that how it's handled in ESO?

api/v1alpha1/clusterscmprovider_types.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"reflect"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+var ClusterScmProviderKind = reflect.TypeOf(ClusterScmProvider{}).Name()
```
zachaller, line 28: Maybe just

ClusterScmProvider{}.Kind

api/v1alpha1/scmprovider_types.go
```diff
@@ -17,20 +17,24 @@ limitations under the License.
...
+	"reflect"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+var ScmProviderKind = reflect.TypeOf(ScmProvider{}).Name()
```
zachaller, line 30: ScmProvider{}.Kind?

zachaller:
Yea I don't like the namespace field either I think I like removing it and pulling secrets from the same namespace as the ClusterScmPrrovider

crenshaw-dev:
Wait so there are two potential namespace fields involved. 

One is in the ScmProvider, to make it mutually interchangeable with ClusterScmProvider, but the user isn't allowed to set any namespace besides the namespace the ScmProvider is in. 

The other is in the ClusterScmProvider. I assume it's there because the ClusterScmProvider is cluster-scoped, so you have to set a namespace for the secretRef. @zachaller were you thinking of the ClusterScmProvider being a namespaced resource?

zachaller:
Sorry was on phone the first message, I see the issue after reading more. I still don't love namespace will think more about controller namespace vs other options.

zachaller:
Looks like the AWS type for ESO: https://github.com/external-secrets/external-secrets/blob/main/apis/meta/v1/types.go#L52

crenshaw-dev:
Interesting. Gives me the ick, to me the API is king, even if the backend ends up a bit less nice.

zachaller:
cert-manager has a flag https://stackoverflow.com/questions/73832864/how-to-provide-reference-to-the-secret-namespace-in-clusterissuer


https://cert-manager.io/v1.5-docs/faq/cluster-resource/

zachaller:
> Interesting. Gives me the ick, to me the API is king, even if the backend ends up a bit less nice.

Yea I thought about just copying the type too, so ClusterScmProvide would have namespace and the  ScmProvider would not.

zachaller:
I lean towards controller namespace for now, until a real usecase comes up, I think doing it in the controller namespace leaves us open to the other options later if needed.

crenshaw-dev:
So cluster-scoped ClusterScmProvider, no namespace field, uses controller namespace.

ScmProvider doesn't change.

zachaller:
> So cluster-scoped ClusterScmProvider, no namespace field, uses controller namespace.
> 
> ScmProvider doesn't change.

Yea, I like that for now, until we can learn more about real usecase and discuss why controller namespace doesn't work well and maybe the config flag is a better fit to protect the API etc etc.

crenshaw-dev:
Controller ns might be a little weird when we add an instance ID later. But meh, I think it's a fine Starr while in pre-GA.

api/v1alpha1/common_types.go
```diff
@@ -23,6 +23,14 @@ type ObjectReference struct {
 	Name string `json:"name"`
 }
 
+type TypedObjectReference struct {
+	// Kind is the type of resource being referenced
+	// +kubebuilder:validation:Required
+	Kind string `json:"kind"`
```
crenshaw-dev: @zachaller what are your thoughts about this vs. `clusterScmProviderRef`? This matches ESO, which is nice.

If we go this way, we should have an enum validation on this field.

internal/settings/manager.go
```diff
@@ -67,6 +67,10 @@ func (m *Manager) GetPullRequestRequeueDuration(ctx context.Context) (time.Durat
 	return controllerConfiguration.Spec.PullRequestRequeueDuration.Duration, nil
 }
 
+func (m *Manager) GetConfig() ManagerConfig {
```
zachaller: Can we turn this into a GetGlobalNamespace that calls `m.GetControllerConfiguration(ctx)` like the other functions to support hot reloading.

internal/utils/utils.go
```diff
@@ -71,20 +100,28 @@ func GetScmProviderAndSecretFromRepositoryReference(ctx context.Context, k8sClie
...
+		Namespace: secretNamespace,
+		Name:      scmProvider.GetSpec().SecretRef.Name,
 	}
 	err = k8sClient.Get(ctx, objectKey, &secret)
 	if err != nil {
+		kind := scmProvider.GetObjectKind().GroupVersionKind().Kind
 		if k8serrors.IsNotFound(err) {
-			logger.Info("Secret from ScmProvider not found", "namespace", scmProvider.Namespace, "name", objectKey.Name)
-			return nil, nil, fmt.Errorf("secret from ScmProvider not found: %w", err)
+			logger.Info("Secret from %s not found"+kind, "namespace", secretNamespace, "name", objectKey.Name)
```
zachaller: This looks incorrect, the +kind with the %s

internal/utils/utils.go
```diff
@@ -71,20 +100,28 @@ func GetScmProviderAndSecretFromRepositoryReference(ctx context.Context, k8sClie
...
+			return nil, nil, fmt.Errorf("secret from %s not found: %w", kind, err)
 		}
 
-		logger.Error(err, "failed to get Secret from ScmProvider", "namespace", scmProvider.Namespace, "name", objectKey.Name)
-		return nil, nil, fmt.Errorf("failed to get Secret from ScmProvider: %w", err)
+		logger.Error(err, "failed to get Secret from %s"+kind, "namespace", secretNamespace, "name", objectKey.Name)
```
zachaller: Same looks incorrect?

internal/utils/utils_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+const (
+	namespace           = "test-namespace"
+	controllerNamespace = "controller-namespace"
+)
+
+func TestGetScmProviderFromGitRepository(t *testing.T) {
```
zachaller: Can we turn these tests into ginkgo tests see: template_test.go, that way we get [coverage stil](https://github.com/argoproj-labs/gitops-promoter/pull/161) and we standardize testing.

api/v1alpha1/common_types.go
zachaller: I like kind, it also matches what like RoleBindings due which I think we could also use that concept for here as well. I agree though with the string we should enable enum validation via cel.

zachaller: I think we can also default to ScmProvider vs ""

internal/utils/utils_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	clusterScmProvider := &promoterv1alpha1.ClusterScmProvider{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster-scm-provider",
+		},
+	}
+
+	scheme := runtime.NewScheme()
+	require.NoError(t, promoterv1alpha1.AddToScheme(scheme))
+	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(namespacedScmProvider, clusterScmProvider).Build()
```
zachaller: This is our first use of fake client for tests, I am mixed I see the utility just have nightmares from rollouts :( maybe for simple usage it's fine. Not sure about mixing test patterns

api/v1alpha1/clusterscmprovider_types.go
alexymantha: Tried it but the Kind will just be empty. Not sure if there's a better way

internal/settings/manager.go
alexymantha: I can rename it `GetGlobalNamespace` but we get the namespace at init time so there is nothing to hot-reload https://github.com/argoproj-labs/gitops-promoter/blob/06f08e8679d07113498927d66af5d68743bf7963/cmd/main.go#L91-L95

The namespace where the controller is running cannot change while it's running

api/v1alpha1/common_types.go
alexymantha: I wanted to add an enum validation but this is in a generic `TypedObjectReference` so we can't really add it directly to the field, unless I make this struct specific to the ScmProviders

crenshaw-dev: I'd just make it specific, I think it's worth it for the validation.

internal/utils/utils_test.go
alexymantha: I agree, especially when the fake clients are to do writes and try to mock complex behaviours. But I think it's pretty convenient for use-cases like this where we just want to get a static objects. I will let you and @crenshaw-dev decide if you want to set the precedent 😆, let me know if you want me to change it to something else.

api/v1alpha1/clusterscmprovider_types.go
zachaller: Yea, it's fine I forgot that Object meta is only poplulated when you Get from the server.

internal/settings/manager.go
zachaller: Ahh sure, yea lets rename and scope down so that other people don't start using it for other config fields.

internal/utils/utils.go
```diff
@@ -18,34 +18,63 @@ import (
...
 		}
 
-		logger.Error(err, "failed to get ScmProvider", "namespace", namespace, "name", objectKey.Name)
-		return nil, fmt.Errorf("failed to get ScmProvider: %w", err)
+		err := k8sClient.Get(ctx, objectKey, &scmProvider, &client.GetOptions{})
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				logger.Info("ClusterScmProvider not found", "name", objectKey.Name)
+				return nil, fmt.Errorf("ClusterScmProvider not found: %w", err)
+			}
```
crenshaw-dev: Can we just drop the special handling of not found? The only advantage is constructing a slightly clearer error message. I'm not sure it's worth the lines of code.

internal/utils/utils_test.go
crenshaw-dev: Chatted with Zach. I think we should just make sure that we expand existing tests to cover the ClusterScmProvider use case. If that doesn't give us the coverage we need, we can revisit adding more specific tests.

internal/settings/manager.go
```diff
@@ -67,6 +67,10 @@ func (m *Manager) GetPullRequestRequeueDuration(ctx context.Context) (time.Durat
 	return controllerConfiguration.Spec.PullRequestRequeueDuration.Duration, nil
 }
 
+func (m *Manager) GetGlobalNamespace() string {
+	return m.config.GlobalNamespace
+}
```
crenshaw-dev: Can we go ahead and rename this `ControllerNamespace` across the board, just to be more explicit?

And maybe make the field private so that no one can accidentally overwrite the value after the Manager has been initialized.

internal/settings/manager.go
zachaller: Can we call it `GetControllerNamespace`

internal/utils/utils_test.go
alexymantha: I tried adding it to one of the controller suites originally, but that didn't feel right because that behavior is shared between multiple controllers and doesn't belong to one in particular. Also, since the controllers will interact with the `GenericScmProvider` they shouldn't really have to know what kind of provider it is.

I'll try to add it again and see what it looks like

internal/settings/manager.go
alexymantha: Yep, renamed the field as well

internal/utils/utils_test.go
alexymantha: Done!

internal/settings/manager.go
alexymantha: Renamed the field and functions. For the private field, the config is already private and the only way to access its fields is these getters so I think we could leave the `ControllerNamespace` field like that since it makes it more convenient to initialize like this: https://github.com/argoproj-labs/gitops-promoter/blob/06f08e8679d07113498927d66af5d68743bf7963/internal/controller/suite_test.go#L153-L155 vs having a constructor

internal/utils/utils_test.go
zachaller: I would add it to the promotion strategy controller, that kind of tests the whole process, aka you can pretty much copy one of the simple tests but replace the creation with a ClusterScmProvicer

alexymantha: Sounds good, will do that. Thanks!

alexymantha: Done

alexymantha:
Thanks to both of you for the quick feedback!

internal/controller/gitrepository_controller_test.go
```diff
@@ -51,6 +51,11 @@ var _ = Describe("GitRepository Controller", func() {
 						Name:      resourceName,
 						Namespace: "default",
 					},
+					Spec: promoterv1alpha1.GitRepositorySpec{
+						ScmProviderRef: promoterv1alpha1.ScmProviderObjectReference{
+							Kind: promoterv1alpha1.ScmProviderKind,
```
crenshaw-dev, line 56: Hm, doesn't this also need a name?

internal/utils/utils.go
```diff
@@ -18,34 +18,53 @@ import (
...
 
-	var scmProvider promoterv1alpha1.ScmProvider
-	namespace := obj.GetNamespace()
-
-	objectKey := client.ObjectKey{
-		Namespace: namespace,
-		Name:      repositoryRef.Spec.ScmProviderRef.Name,
+	var provider promoterv1alpha1.GenericScmProvider
+	kind := repositoryRef.Spec.ScmProviderRef.Kind
+	if kind == "" {
```
crenshaw-dev: Safe to drop this check? Should be enforced by k8s validation. Worst-case scenario we fall back to the error case.

internal/utils/utils.go
```diff
@@ -18,34 +18,53 @@ import (
...
-		return nil, fmt.Errorf("failed to get ScmProvider: %w", err)
+		err := k8sClient.Get(ctx, objectKey, &scmProvider, &client.GetOptions{})
+		if err != nil {
+			logger.Error(err, "failed to get ClusterScmProvider", "name", objectKey.Name)
+			return nil, fmt.Errorf("failed to get ClusterScmProvider: %w", err)
+		}
+		provider = &scmProvider
+	case promoterv1alpha1.ScmProviderKind:
+		var scmProvider promoterv1alpha1.ScmProvider
+		namespace := obj.GetNamespace()
```
crenshaw-dev: Could either remove this alias or use it also on line 47.

crenshaw-dev (changes requested):
Only had nitpicks. Could you add docs? Besides that, I think this is good to go!

internal/controller/gitrepository_controller_test.go
alexymantha: It's not actually used in the tests as far as I can tell, I had to add the kind because otherwise it was failing schema validation when applied because an empty string is not a valid value. I'm not 100% sure why, because when applied locally it works fine when the kind is omitted and it defaults to `ScmProvider`. I'll try to dig into it, I feel it should default to `ScmProvider` in the tests as well without having to provide it manually?

alexymantha: Ah because the Go struct doesn't differentiate between not set and empty string so it thinks it's being set to `""`. It would need to be a string pointer to make the distinction

crenshaw-dev: Ah sure. Kind behavior makes sense, reasonable workaround as long as the k8s validation works on-cluster.

Wouldn't lose sleep over the name field, just surprised me that it was empty.

crenshaw-dev (approved):

Issue #307, opened 2025-05-08
Auto-resolve merge conflicts


crenshaw-dev:
If the promoter detects merge conflicts between the proposed and active branch, it should merge the active branch into the proposed branch using a command something like this:

```sh
git checkout environment/development-next
git merge environment/development --strategy-option=ours
git push
```

We take "ours" to resolve conflict, effectively trusting the hydrator's version of all conflicting files to be correct.

PR #318, opened 2025-05-21
feat: merge conflict resolution

zachaller:
fixes: https://github.com/argoproj-labs/gitops-promoter/issues/307


internal/controller/changetransferpolicy_controller.go
```diff
@@ -515,6 +520,37 @@ func (r *ChangeTransferPolicyReconciler) mergePullRequests(ctx context.Context,
...
+
+	// If we have a conflict, perform a merge with "ours" strategy
+	logger.Info("Conflicts detected, performing merge with 'ours' strategy", "proposed", ctp.Spec.ProposedBranch, "active", ctp.Spec.ActiveBranch)
+
+	err = gitOperations.MergeWithOursStrategy(ctx, ctp.Spec.ProposedBranch, ctp.Spec.ActiveBranch)
+	if err != nil {
+		return fmt.Errorf("failed to merge branches %q and %q with 'ours' strategy: %w", ctp.Spec.ProposedBranch, ctp.Spec.ActiveBranch, err)
+	}
+
+	r.Recorder.Event(ctp, "Normal", "MergeWithOursStrategy", fmt.Sprintf("Merged %s into %s with 'ours' strategy to resolve conflicts", ctp.Spec.ProposedBranch, ctp.Spec.ActiveBranch))
```
crenshaw-dev: I wonder if we should go with a more human-friendly event name like `ResolveConflicts`?

We should probably also put all the event names in some constants module.

internal/git/git.go
```diff
@@ -339,8 +339,75 @@ func (g *GitOperations) runCmd(ctx context.Context, directory string, args ...st
...
 	if err := cmd.Wait(); err != nil {
 		// exitErr := err.(*exec.ExitError)
-		return "", stderrBuf.String(), err
+		return stdoutBuf.String(), stderrBuf.String(), err
 	}
 
 	return stdoutBuf.String(), stderrBuf.String(), nil
 }
+
+// replace lines 348 to 386
```
crenshaw-dev: Leftover comment?

internal/git/git.go
```diff
@@ -339,8 +339,75 @@ func (g *GitOperations) runCmd(ctx context.Context, directory string, args ...st
...
+func (g *GitOperations) HasConflict(ctx context.Context, proposedBranch, activeBranch string) (bool, error) {
+	logger := log.FromContext(ctx)
+	repoPath := gitpaths.Get(g.gap.GetGitHttpsRepoUrl(*g.gitRepo) + g.pathContext)
+
+	// Fetch both branches from origin
+	if _, stderr, err := g.runCmd(ctx, repoPath, "fetch", "origin", activeBranch, proposedBranch); err != nil {
+		logger.Error(err, "could not fetch branches", "stderr", stderr)
+		return false, err
```
crenshaw-dev: Wrap the error?

internal/git/git.go
```diff
@@ -339,8 +339,75 @@ func (g *GitOperations) runCmd(ctx context.Context, directory string, args ...st
...
 	return stdoutBuf.String(), stderrBuf.String(), nil
 }
+
+// replace lines 348 to 386
+func (g *GitOperations) HasConflict(ctx context.Context, proposedBranch, activeBranch string) (bool, error) {
+	logger := log.FromContext(ctx)
+	repoPath := gitpaths.Get(g.gap.GetGitHttpsRepoUrl(*g.gitRepo) + g.pathContext)
+
+	// Fetch both branches from origin
+	if _, stderr, err := g.runCmd(ctx, repoPath, "fetch", "origin", activeBranch, proposedBranch); err != nil {
```
crenshaw-dev: Is this going to double-up our number of fetches, i.e. network round trips? Can we instead assume that this has already been done?

internal/git/git.go
```diff
@@ -339,8 +339,75 @@ func (g *GitOperations) runCmd(ctx context.Context, directory string, args ...st
...
+
+	// Always attempt to abort the merge to clean up
+	if _, abortStderr, abortErr := g.runCmd(ctx, repoPath, "merge", "--abort"); abortErr != nil {
+		if !strings.Contains(abortStderr, "MERGE_HEAD missing") { // Ignore the error if there is no merge in progress
+			logger.Error(abortErr, "could not abort merge", "stderr", abortStderr)
+			return false, fmt.Errorf("failed to abort merge: %w", abortErr)
+		}
+	}
+
+	if err != nil && conflictDetected {
```
crenshaw-dev: Is the err check necessary? Could we just check conflictDetected and then return early?

crenshaw-dev (changes requested):

internal/git/git.go
zachaller: removed, it just shortened the window of possible changes during the reconcile so meh..

internal/git/git.go
```diff
@@ -339,8 +339,69 @@ func (g *GitOperations) runCmd(ctx context.Context, directory string, args ...st
...
 	if err := cmd.Wait(); err != nil {
 		// exitErr := err.(*exec.ExitError)
-		return "", stderrBuf.String(), err
+		return stdoutBuf.String(), stderrBuf.String(), err
 	}
 
 	return stdoutBuf.String(), stderrBuf.String(), nil
 }
+
+func (g *GitOperations) HasConflict(ctx context.Context, proposedBranch, activeBranch string) (bool, error) {
```
crenshaw-dev, line 351: Ye olde docstring? Probably worth mentioning that we expect the fetch to have already been done.

internal/controller/changetransferpolicy_controller.go
zachaller: done

zachaller: I moved all events over as well

internal/git/git.go
zachaller: that requirment is required in lots of places not just here we can actually remove another fetch if we are ok with the chance of change mid reconcile.

crenshaw-dev: Yeah I feel okay about that within a reconcile. Presumably a new commit would trigger another reconcile.

zachaller: a new commit on something like the sync from branch might not cause a reconcile would need to check webhook

crenshaw-dev: Webhook currently matches on `beforeSha`, which is probably already too restrictive.

```go
	err := wr.k8sClient.List(ctx, &ctpLists, &client.ListOptions{
		FieldSelector: fields.SelectorFromSet(map[string]string{
			".status.proposed.hydrated.sha": beforeSha,
		}),
	})
```

We should probably match on proposed branch name at the very least, and maybe on active branch name.

zachaller: That would cause a lot more API calls (rate limited) for extra reconciles

crenshaw-dev: Actually we don't need the webhook to trigger another reconcile. If there's a conflict, the push will fail, and we'll fail the reconcile and requeue.

crenshaw-dev (approved):

PR #324, opened 2025-05-29
fix: avoid invalid examples for github domain

crenshaw-dev:
Setting github.domain to github.com causes it to use an incorrect API endpoint. Avoid using the invalid example in docs.

### DIFF
--- a/internal/scms/github/git_operations.go
+++ b/internal/scms/github/git_operations.go
@@ -29,7 +29,7 @@ func NewGithubGitAuthenticationProvider(scmProvider v1alpha1.GenericScmProvider,
 		panic(err)
 	}
 
-	if scmProvider.GetSpec().GitHub != nil && scmProvider.GetSpec().GitHub.Domain != "" && scmProvider.GetSpec().GitHub.Domain != "github.com" {
+	if scmProvider.GetSpec().GitHub != nil && scmProvider.GetSpec().GitHub.Domain != "" {
 		itr.BaseURL = fmt.Sprintf("https://%s/api/v3", scmProvider.GetSpec().GitHub.Domain)
 	}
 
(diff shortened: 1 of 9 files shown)


internal/scms/github/git_operations.go
```diff
@@ -29,7 +29,7 @@ func NewGithubGitAuthenticationProvider(scmProvider v1alpha1.GenericScmProvider,
 		panic(err)
 	}
 
-	if scmProvider.GetSpec().GitHub != nil && scmProvider.GetSpec().GitHub.Domain != "" {
+	if scmProvider.GetSpec().GitHub != nil && scmProvider.GetSpec().GitHub.Domain != "" && scmProvider.GetSpec().GitHub.Domain != "github.com" {
```
crenshaw-dev: @zachaller is this too radical? :-P Based on @FourFifthsCode's experience, I think setting github.com is incorrect and just breaks things.

zachaller: I think we should default to the cloud installation it will be most use cases and follows other patterns of special config for enterprise such as the many skd's for languages etc.

zachaller: We can make it more clear via docs etc or comments in spec that it needs to be adjusted for enterprise

crenshaw-dev: Not sure I follow "we should default to the cloud installation." We do default to the cloud installation, this change just causes "github.com" to also default to the cloud installation rather than setting the base URL as if it were an enterprise installation.

It's basically trying to catch the user's mistake and fix it for them.

zachaller: actually sorry miss read this, I think I am fine with the idea how do we feel about cel validation vs runtime?

zachaller: just a regex validation to not allow `^github.com`

crenshaw-dev: I like CEL, lemme try that

crenshaw-dev: Local test looks good:


> The ScmProvider "example-scm-provider" is invalid: spec.github.domain: Invalid value: "string": Instead of setting the domain to github.com, leave the field blank

zachaller (approved):

PR #327, opened 2025-06-03
feat: add multicluster support for watching argocd apps in other clusters

FourFifthsCode:
Watch argocd applications in external clusters using multi-controller support
* uses kubeconfig provider for connecting to external clusters


cmd/main.go
```diff
@@ -77,6 +84,8 @@ func main() {
 		"If set the metrics endpoint is served securely")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
+	flag.StringVar(&kubeconfigSecretLabel, "kubeconfig-secret-label", "sigs.k8s.io/multicluster-runtime-kubeconfig", "The label of the kubeconfig secret.")
+	flag.StringVar(&kubeconfigSecretKey, "kubeconfig-secret-key", "kubeconfig", "The key of the kubeconfig secret.")
```
zachaller: These should go into the config file not a cli flag

crenshaw-dev: Or just hard code an opinionated value?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -273,90 +289,104 @@ var (
...
+			var application argocd.ArgoCDApplication
+			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(un.Object, &application); err != nil {
+				log.FromContext(ctx).Error(err, "failed to convert unstructured object to typed object")
+				return nil
+			}
 
-		rwMutex.RLock()
-		appRef := revMap[appKey]
-		rwMutex.RUnlock()
+			appKey := application.GetNamespace() + "/" + application.GetName()
```
zachaller: This is going to have to account for clusters

FourFifthsCode: I think this will do the trick:
```go
// if clusterName is empty, then cluster == local cluster
appKey := fmt.Sprintf("%s/%s/%s", clusterName, application.GetNamespace(), application.GetName())
```

cmd/main.go
FourFifthsCode: Sometimes a different team or person creates the secrets and another deploys the app. I like the idea of having it configurable to reduce friction during deployment, but we can certainly set opinionated defaults. Hopefully best of both worlds that way.

internal/controller/argocdcommitstatus_controller.go
crenshaw-dev: I'd switch to a struct instead of a string key tbh. Strings make me nervous about collisions.

cmd/main.go
zachaller: I actually lean towards hard coding as well for now less options is good until a real use case comes up

FourFifthsCode: fair enough, I'll hard code it 😄

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: sounds good

FourFifthsCode:
temporarily using fork in `go.mod` until kubeconfig provider is upstreamed 🙏 :
```
replace (
	sigs.k8s.io/multicluster-runtime => github.com/FourFifthsCode/multicluster-runtime v0.0.0-20250609195707-c976645c49b1
	sigs.k8s.io/multicluster-runtime/providers/kubeconfig => github.com/FourFifthsCode/multicluster-runtime/providers/kubeconfig v0.0.0-20250609195707-c976645c49b1
)
```

Upstream PR: https://github.com/kubernetes-sigs/multicluster-runtime/pull/45

internal/controller/suite_test.go
```diff
@@ -88,18 +101,26 @@ func TestControllers(t *testing.T) {
 	RegisterFailHandler(Fail)
 
 	c, _ := GinkgoConfiguration()
-	// c.FocusFiles = []string{
-	// 	"changetransferpolicy_controller_test.go",
-	// 	"pullrequest_controller_test.go",
-	// 	"promotionstrategy_controller_test.go",
-	// }
-	// GinkgoWriter.TeeTo(os.Stdout)
+	c.FocusFiles = []string{
```
zachaller: Can we comment these back, just want to keep the default running all new files, because you have skipped some tests here such as template_test.go in utils.

FourFifthsCode: yep, I totally forgot to put those back

Makefile
```diff
@@ -3,6 +3,14 @@ IMG ?= quay.io/argoprojlabs/gitops-promoter:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.31.0
 
+CURRENT_DIR=$(shell pwd)
+
+# Docker command to use
+DOCKER?=docker
+
+MKDOCS_DOCKER_IMAGE?=python:3.7-alpine
```
crenshaw-dev: 3.7 is pretty old right? Will it run with a newer version?

Makefile
```diff
@@ -3,6 +3,14 @@ IMG ?= quay.io/argoprojlabs/gitops-promoter:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.31.0
 
+CURRENT_DIR=$(shell pwd)
+
+# Docker command to use
+DOCKER?=docker
```
crenshaw-dev: Can use `CONTAINER_TOOL` instead to match other targets.

cmd/main.go
```diff
@@ -128,7 +133,18 @@ func main() {
...
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	// Create the kubeconfig provider with options
+	providerOpts := kubeconfigprovider.Options{
+		Namespace:             controllerNamespace,
+		KubeconfigSecretLabel: "sigs.k8s.io/multicluster-runtime-kubeconfig",
+		KubeconfigSecretKey:   "kubeconfig",
+	}
+
+	// Create the provider first, then the manager with the provider
+	setupLog.Info("Creating provider")
```
crenshaw-dev: I wouldn't bother logging unless `.New` is likely to panic.

cmd/main.go
```diff
@@ -152,10 +168,12 @@ func main() {
 		// after the manager stops then its usage might be unsafe.
 		// LeaderElectionReleaseOnCancel: true,
 	})
-	if err != nil || mgr == nil {
+	if err != nil || mcMgr == nil {
 		panic("unable to start manager")
 	}
 
+	mgr := mcMgr.GetLocalManager()
```
crenshaw-dev: Just for extra clarity below, maybe `localMgr`?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -43,25 +45,47 @@ import (
...
+}
+
 // ArgoCDCommitStatusReconciler reconciles a ArgoCDCommitStatus object
 type ArgoCDCommitStatusReconciler struct {
-	client.Client
-	Scheme      *runtime.Scheme
-	SettingsMgr *settings.Manager
+	Manager            mcmanager.Manager
+	SettingsMgr        *settings.Manager
+	KubeConfigProvider *kubeconfig.Provider
```
crenshaw-dev, line 86: Instead of having a separate field for this, can we instead just call `Manager.GetProvider()`?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -99,16 +127,31 @@ func (r *ArgoCDCommitStatusReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-	err = r.List(ctx, &apps, &client.ListOptions{
-		LabelSelector: ls,
-	})
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to list Applications: %w", err)
+
+	// list clusters so we can query argocd applications from all clusters
+	clusters := r.KubeConfigProvider.ListClusters()
+	clusters = append(clusters, "") // add the local cluster
+	for _, clusterName := range clusters {
```
crenshaw-dev, line 133: We should add a log line at the top of the loop. If lists are taking a long time, the log lines will let us know that we need to parallelize this.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -255,89 +298,100 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
 		Complete(r)
 	if err != nil {
 		return fmt.Errorf("failed to create controller: %w", err)
 	}
+
 	return nil
 }
 
-func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.Context, argoCDCommitStatus promoterv1alpha1.ArgoCDCommitStatus, targetBranch string, sha string, phase promoterv1alpha1.CommitStatusPhase, desc string) error {
+func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.Context, cl client.Client, argoCDCommitStatus promoterv1alpha1.ArgoCDCommitStatus, targetBranch string, sha string, phase promoterv1alpha1.CommitStatusPhase, desc string) error {
```
crenshaw-dev: Instead of accepting a new param, can we just call `r.Manager.GetLocalManager().GetClient()` to get the local client?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -390,27 +444,27 @@ func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.
 	return nil
 }
 
-func (r *ArgoCDCommitStatusReconciler) getPromotionStrategy(ctx context.Context, namespace string, promotionStrategyRef promoterv1alpha1.ObjectReference) (*promoterv1alpha1.PromotionStrategy, error) {
+func (r *ArgoCDCommitStatusReconciler) getPromotionStrategy(ctx context.Context, cl client.Client, namespace string, promotionStrategyRef promoterv1alpha1.ObjectReference) (*promoterv1alpha1.PromotionStrategy, error) {
```
crenshaw-dev: Instead of accepting a new param, can we just call `r.Manager.GetLocalManager().GetClient()` to get the local client?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -390,27 +444,27 @@ func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.
...
-	err := r.Get(ctx, client.ObjectKey{Namespace: namespace, Name: promotionStrategyRef.Name}, &promotionStrategy, &client.GetOptions{})
+	err := cl.Get(ctx, client.ObjectKey{Namespace: namespace, Name: promotionStrategyRef.Name}, &promotionStrategy)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get PromotionStrategy object: %w", err)
 	}
 	return &promotionStrategy, nil
 }
 
-func (r *ArgoCDCommitStatusReconciler) getGitAuthProvider(ctx context.Context, argoCDCommitStatus promoterv1alpha1.ArgoCDCommitStatus) (scms.GitOperationsProvider, promoterv1alpha1.ObjectReference, error) {
+func (r *ArgoCDCommitStatusReconciler) getGitAuthProvider(ctx context.Context, cl client.Client, argoCDCommitStatus promoterv1alpha1.ArgoCDCommitStatus) (scms.GitOperationsProvider, promoterv1alpha1.ObjectReference, error) {
```
crenshaw-dev: Instead of accepting a new param, can we just call `r.Manager.GetLocalManager().GetClient()` to get the local client?

docs/commit-status-controllers/argocd.md
```diff
@@ -51,4 +51,51 @@ spec:
...
+     verbs:
+     - get
+     - list
+     - watch
+   ```
+
+#### Helper Script
+  Use this [helper script](https://github.com/FourFifthsCode/gitops-promoter/blob/multi-cluster-support/hack/create-kubeconfig-secret-rbac.sh) to automate the secret creation and RBAC setup process
+  ```
+    Usage: ./create-kubeconfig-secret-rbac.sh [options]
```
crenshaw-dev, line 91: After https://github.com/argoproj-labs/gitops-promoter/pull/348, I wonder if this should be its own command. @zachaller whatcha think?

crenshaw-dev (changes requested):
Quick first pass

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: Yeah, good idea, in fact, we can just do this once during SetupManager and make it a private variable. The cluster client is the only one that should be changing per reconcile

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: Added local client to struct

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: Added local client to struct

Makefile
FourFifthsCode: totally

Makefile
FourFifthsCode: yep, looks like 3.13 is latest and appears to work ok

cmd/main.go
FourFifthsCode: yeah, good point

cmd/main.go
FourFifthsCode: sounds good

docs/commit-status-controllers/argocd.md
FourFifthsCode: I'm a fan of using the cli instead of a separate script

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: Normally yes, but we need to access a method on this provider (ListClusters) that doesn't exist on the interface. I could do a typecast, but it seems clearer just to pass the type from the beginning.

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: sounds good

docs/commit-status-controllers/argocd.md
zachaller: hm... yea I go a bit back and forth on it I worry a little bit about creep in features for a subcommand to generate k8s configs when I feel there are already a lot of tools and personal preferences around it. I think when it's a script it seems a bit less "supported" and more of just a nice helper. I don't I am fine either way if you guys both think it's worth the maintenance of it as a full sub command I would be ok with that.

I would do that in it's own PR though.

FourFifthsCode: that's a fair point, Im good to start with the script and then evaluate again later and see if folks would prefer a cli command over it

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -279,78 +312,93 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
 
-func lookupArgoCDCommitStatusFromArgoCDApplication(c client.Client) func(ctx context.Context, argoCDApplication client.Object) []reconcile.Request {
-	return func(ctx context.Context, argoCDApplication client.Object) []reconcile.Request {
-		var application argocd.Application
-		if err := c.Get(ctx, client.ObjectKey{Namespace: argoCDApplication.GetNamespace(), Name: argoCDApplication.GetName()}, &application, &client.GetOptions{}); err != nil {
-			if k8s_errors.IsNotFound(err) {
-				log.FromContext(ctx).V(4).Info("Application not found", "application", argoCDApplication.GetName(), "app-namespace", argoCDApplication.GetNamespace())
+			// fetch the ArgoCDApplication from the cluster
+			if err := cl.GetClient().Get(ctx, client.ObjectKeyFromObject(argoCDApplication), application, &client.GetOptions{}); err != nil {
+				log.FromContext(ctx).Error(err, "failed to get ArgoCDApplication")
```
zachaller: can we turn this into a variable

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -279,78 +312,93 @@ func (r *ArgoCDCommitStatusReconciler) getMostRecentLastTransitionTime(aggregate
...
+	r.localClient = mcMgr.GetLocalManager().GetClient()
+
+	app := &argocd.Application{}
+	err := mcbuilder.ControllerManagedBy(mcMgr).
+		For(&promoterv1alpha1.ArgoCDCommitStatus{},
+			mcbuilder.WithEngageWithLocalCluster(true),
+			mcbuilder.WithEngageWithProviderClusters(false),
+			mcbuilder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).
+		Watches(app, lookupArgoCDCommitStatusFromArgoCDApplication(mcMgr),
```
zachaller: Why the variable `app`, can we not just pass `&argocd.Application{}`

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: sounds good

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: I think it was previously being passed to another function, you're totally right, should just pass it directly

internal/controller/suite_test.go
```diff
@@ -67,19 +71,28 @@ import (
...
+	gitServerPort    string
 )
 
 const (
-	EventuallyTimeout   = 90 * time.Second
-	WebhookReceiverPort = 3333
+	EventuallyTimeout         = 90 * time.Second
+	WebhookReceiverPort       = 3333
+	kubeconfigSecretNamespace = "default"
+	kubeconfigSecretLabel     = "kubeconfig"
```
zachaller: lets move these to the constants packages, create a new file called configurations.go

zachaller:
LGTM, once upstream (multicluster-runtime) is merged this is good to go

zachaller (approved):

crenshaw-dev (approved):

Issue #202, opened 2025-02-22
feat: add gitea / forgejo support


eripa:
Please and thank you :)

robinlioret:
Forgejo added. It should have the same API than Gitea. Can someone test it ?

zachaller:
If the API is that same and it does work, we should probably also create a gittea spec that just uses the forgejo provider and document etc

robinlioret:
Even if it's the same API right now (to confirm), nothing guarantees that it will remain the same un the future. Forgejo us a fork of Gitea, I don't think it's meant to remain compatible nor up to date ,with Gitea features and choices.

Gitea and Forgejo should be considered different SCM.

emirot:
@eripa  this was added in https://github.com/argoproj-labs/gitops-promoter/pull/838

@zachaller or @crenshaw-dev could you please close that issue

PR #329, opened 2025-06-04
feat: add forgejo support

robinlioret:
Implement #202


api/v1alpha1/common_types.go
```diff
@@ -17,6 +17,13 @@ type GitLab struct {
 	Domain string `json:"domain,omitempty"`
 }
 
+type Forgejo struct {
+	// Domain is the Forgejo domain, suche as "codeberg.org" or "forgejo.mycompany.com".
```
crenshaw-dev:
```suggestion
	// Domain is the Forgejo domain, such as "codeberg.org" or "forgejo.mycompany.com".
```

crenshaw-dev (changes requested):

internal/scms/forgejo/commit_status.go
```diff
@@ -0,0 +1,68 @@
...
+	)
+	if resp != nil {
+		metrics.RecordSCMCall(repo, metrics.SCMAPICommitStatus, metrics.SCMOperationCreate, resp.StatusCode, time.Since(start), nil)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to get commit info: %w", err)
+	}
+	logger.V(4).Info("forgejo response status", "status", resp.Status)
+
+	commitStatus.Status.Id = commitStatus.Spec.Sha
```
crenshaw-dev: This should be Forgejo's ID for the created commit status, assuming it has one.

internal/scms/forgejo/utils.go
```diff
@@ -0,0 +1,29 @@
...
+func mapMergeRequestState(state string) v1alpha1.PullRequestState {
+	switch state {
+	case "opened":
+		return v1alpha1.PullRequestOpen
+	case "closed":
+		return v1alpha1.PullRequestClosed
+	default:
+		return v1alpha1.PullRequestMerged
+	}
+}
```
crenshaw-dev: Maybe better to throw an error on unexpected strings? Forgejo might add stuff in the future that we'd want to handle appropriately.

docs/getting-started.md
```diff
@@ -164,6 +163,52 @@ spec:
...
+This Token should be in a secret as follow:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: <your-secret-name>
+type: Opaque
+stringData:
+  token: <your-access-token>
```
crenshaw-dev, line 179: Looks like we also have a `user` and `password` field to document.

internal/scms/forgejo/commit_status.go
robinlioret: Everything seems based on the SHA. Even the Commit struct does make any reference to the commit's id. The only reference I can find are the arguments to the functions.

I'll do some testing to be sure. Until then, let's keep this PR a draft.

internal/scms/forgejo/utils.go
robinlioret: I agree, i'll make the change. That is something we may want to do on the GitLab implementation as well.

docs/getting-started.md
robinlioret: Yes, I forgot it in the doc. I'll add it.

internal/scms/forgejo/utils.go
crenshaw-dev: Yep, good call!

internal/scms/forgejo/commit_status.go
crenshaw-dev: If you find that the SHA is indeed the ID, probably just add a short comment to explain that.

robinlioret: Tested. There is no such thing as CommitID in Forgejo. The closest thing I found is the "short" hash wich is basically the 10th first chars of the full sha.

robinlioret:
@crenshaw-dev How can I E2E test this provider ? I can't see any complete test scenarios for github or gitlab.

crenshaw-dev:
@robinlioret we don't have provider-specific e2e tests set up yet, that's all manual for now.

internal/scms/forgejo/pullrequest.go
```diff
@@ -0,0 +1,213 @@
...
+
+	repo, err := utils.GetGitRepositoryFromObjectKey(ctx, pr.k8sClient, k8sClient.ObjectKey{
+		Namespace: prObj.Namespace,
+		Name:      prObj.Name,
+	})
+	if err != nil {
+		return false, "", fmt.Errorf("failed to merge pull request: %w", err)
+	}
+
+	options := forgejo.ListPullRequestsOptions{}
```
zachaller: Does options have a way to filter open pull requests?

internal/scms/forgejo/pullrequest.go
```diff
@@ -0,0 +1,213 @@
...
+		return false, "", fmt.Errorf("failed to list pull requests: %w", err)
+	}
+	if resp != nil {
+		metrics.RecordSCMCall(repo, metrics.SCMAPIPullRequest, metrics.SCMOperationCreate, resp.StatusCode, time.Since(start), nil)
+	}
+	logger.V(4).Info("forgejo response status", "status", resp.Status)
+
+	for _, pr := range prs {
+		// REVIEW: head =? source or is it the other way around
+		// REVIEW: we may need to improve the pr selection. This is a hazard and can create weird situations is multiple PRs are targeting the same branches.
```
zachaller: Github and gitlab I think only let you open 1 pull per branch pair, not sure if foregejo is the same?

internal/scms/forgejo/pullrequest.go
zachaller: Even better can you filter for branch pairs?

robinlioret: That would be great, but it doesn't look like it: https://pkg.go.dev/codeberg.org/mvdkleijn/forgejo-sdk/forgejo#ListPullRequestsOptions

internal/scms/forgejo/pullrequest.go
robinlioret: I tested it. You're right. Let me remove the comments.

internal/scms/forgejo/pullrequest.go
crenshaw-dev: Looks like you could filter by open state:

```go
State StateType `json:"state"`
```

internal/scms/forgejo/commit_status.go
```diff
@@ -0,0 +1,73 @@
...
+		repo.Spec.Forgejo.Owner,
+		repo.Spec.Forgejo.Name,
+		commitStatus.Spec.Sha,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get commit info: %w", err)
+	}
+	if resp != nil {
+		metrics.RecordSCMCall(repo, metrics.SCMAPICommitStatus, metrics.SCMOperationCreate, resp.StatusCode, time.Since(start), nil)
+	}
```
crenshaw-dev, line 71: Metrics call should be before the error check so that we record calls that return errors.

internal/scms/forgejo/pullrequest.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	start := time.Now()
+	pullRequest, resp, err := pr.foregejoClient.CreatePullRequest(repo.Spec.Forgejo.Owner, repo.Spec.Forgejo.Name, options)
+	if err != nil {
+		return "", fmt.Errorf("failed to create pull request: %w", err)
+	}
+	if resp != nil {
+		metrics.RecordSCMCall(repo, metrics.SCMAPIPullRequest, metrics.SCMOperationCreate, resp.StatusCode, time.Since(start), nil)
+	}
```
crenshaw-dev, line 61: Metrics call before error check

internal/scms/forgejo/pullrequest.go
```diff
@@ -0,0 +1,213 @@
...
+
+	start := time.Now()
+	_, resp, err := pr.foregejoClient.EditPullRequest(repo.Spec.Forgejo.Owner, repo.Spec.Forgejo.Name, prID, options)
+	if err != nil {
+		return fmt.Errorf("failed to update pull request: %w", err)
+	}
+	if resp != nil {
+		metrics.RecordSCMCall(repo, metrics.SCMAPIPullRequest, metrics.SCMOperationCreate, resp.StatusCode, time.Since(start), nil)
+	}
```
crenshaw-dev, line 95: Metrics before errors

internal/scms/forgejo/pullrequest.go
```diff
@@ -0,0 +1,213 @@
...
+
+	start := time.Now()
+	_, resp, err := pr.foregejoClient.EditPullRequest(repo.Spec.Forgejo.Owner, repo.Spec.Forgejo.Name, prID, options)
+	if err != nil {
+		return fmt.Errorf("failed to close pull request: %w", err)
+	}
+	if resp != nil {
+		metrics.RecordSCMCall(repo, metrics.SCMAPIPullRequest, metrics.SCMOperationCreate, resp.StatusCode, time.Since(start), nil)
+	}
```
crenshaw-dev, line 134: etc.  :-)

internal/scms/forgejo/pullrequest.go
robinlioret: You're right, how did I miss it?

crenshaw-dev: Documentation Eyes, sets in after too much docs reading 🙂

robinlioret: THe experience talking. I changed the code. Will be there in the next push.

internal/scms/forgejo/commit_status.go
robinlioret: I revert the commit. Coming with the next push :)

FourFifthsCode:
@crenshaw-dev, I spun up a local forgejo environment and this tested ok with http auth and two branches configured with automerge.

internal/scms/forgejo/commit_status.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to set commit status: %w", err)
+	}
+	logger.V(4).Info("forgejo response status", "status", resp.Status)
+
+	csObj.Status.Id = strconv.FormatInt(commitStatus.ID, 10)
+	csObj.Status.Phase = csObj.Spec.Phase
+	csObj.Status.Sha = strconv.FormatInt(commitStatus.ID, 10)
```
crenshaw-dev: Should this instead by base-16 for consistency with the usual SHA presentation?

internal/scms/forgejo/commit_status.go
```diff
@@ -0,0 +1,82 @@
...
+	)
+	if resp != nil {
+		metrics.RecordSCMCall(repo, metrics.SCMAPICommitStatus, metrics.SCMOperationCreate, resp.StatusCode, time.Since(start), nil)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to set commit status: %w", err)
+	}
+	logger.V(4).Info("forgejo response status", "status", resp.Status)
+
+	csObj.Status.Id = strconv.FormatInt(commitStatus.ID, 10)
```
crenshaw-dev: Maybe we should also make the ID base-16

crenshaw-dev (approved):
Thank you!

PR #330, opened 2025-06-05
fix: use template if no commit message

zachaller:
with github to use the configured default commit messages on the PR you send an empty string, with gitlab they want us to remove the commit message from the options struct.

### DIFF
--- a/internal/scms/gitlab/pullrequest.go
+++ b/internal/scms/gitlab/pullrequest.go
@@ -185,10 +185,12 @@ func (pr *PullRequest) Merge(ctx context.Context, commitMessage string, prObj *v
 	}
 
 	options := &gitlab.AcceptMergeRequestOptions{
-		MergeCommitMessage:        gitlab.Ptr(commitMessage),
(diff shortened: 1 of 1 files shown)


internal/scms/gitlab/pullrequest.go
```diff
@@ -185,11 +185,13 @@ func (pr *PullRequest) Merge(ctx context.Context, commitMessage string, prObj *v
 	}
 
 	options := &gitlab.AcceptMergeRequestOptions{
-		MergeCommitMessage:        gitlab.Ptr(commitMessage),
 		MergeWhenPipelineSucceeds: gitlab.Ptr(false),
 		ShouldRemoveSourceBranch:  gitlab.Ptr(false),
 		Squash:                    gitlab.Ptr(false),
 	}
+	if commitMessage != "" {
```
crenshaw-dev, line 193:
```suggestion
	// Gitlab throws a 422 if you send it an empty commit message. So leave it as nil unless we have a message.
	if commitMessage != "" {
```

PR #354, opened 2025-06-25
fix: use typed Application to reduce memory use

crenshaw-dev:
unstructured objects don't get cached in an informer. Use a lightweight version of the Application CR.

We'll likely need to add more fields later to support templating for the ArgoCDCommitStatus URL.


internal/types/argocd/argocd_type.go
```diff
@@ -113,3 +63,16 @@ const (
 	SyncStatusCodeSynced    SyncStatusCode = "Synced"
 	SyncStatusCodeOutOfSync SyncStatusCode = "OutOfSync"
 )
+
+// +kubebuilder:object:root=true
+
+// ArgoCDApplicationList contains a list of ArgoCDApplications.
+type ArgoCDApplicationList struct {
```
zachaller: ApplicationList?

config/crd/bases/argoproj.io_applications.yaml
```diff
@@ -0,0 +1,90 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
crenshaw-dev: Haven't found the right combo of markers to keep this file from being generated while ensuring deepcopy still gets generated. https://book.kubebuilder.io/reference/markers/crd

zachaller (approved):

PR #361, opened 2025-06-26
feat: add conditions to controllers

zachaller:
This adds a single condition to each controller with a type of `Ready` this condition is used to bubble up errors during reconciliation.


internal/utils/utils.go
```diff
@@ -222,3 +227,73 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+type StatusConditionUpdater interface {
+	client.Object
+	GetConditions() *[]metav1.Condition
+}
+
+// HandleReconciliationResultGeneric handles reconciliation results for any object with status conditions.
+func HandleReconciliationResultGeneric[T StatusConditionUpdater](
+	ctx context.Context,
+	startTime time.Time,
+	obj T,
```
crenshaw-dev:
```suggestion
func HandleReconciliationResultGeneric(
	ctx context.Context,
	startTime time.Time,
	obj StatusConditionUpdater,
```

internal/types/conditions/conditions.go
```diff
@@ -0,0 +1,34 @@
...
+	ChangeTrasferPolicyConditionReason string
+)
+
+type (
+	CommitStatusConditionType   string
+	CommitStatusConditionReason string
+)
+
+type (
+	PullReqeustConditionType   string
```
crenshaw-dev:
```suggestion
	PullRequestConditionType   string
```

internal/controller/promotionstrategy_controller.go
```diff
@@ -62,13 +63,16 @@ type PromotionStrategyReconciler struct {
...
-func (r *PromotionStrategyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *PromotionStrategyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
 	logger := log.FromContext(ctx)
 	logger.Info("Reconciling PromotionStrategy")
 	startTime := time.Now()
 
 	var ps promoterv1alpha1.PromotionStrategy
-	err := r.Get(ctx, req.NamespacedName, &ps, &client.GetOptions{})
+
+	defer utils.HandleReconciliationResult(ctx, startTime, &ps, r.Client, r.Recorder, &err, string(conditions.Ready))
```
crenshaw-dev: Everyone passes `conditions.Ready`, I think you could just hard-code it.

internal/controller/promotionstrategy_controller_test.go
```diff
@@ -1303,7 +1303,7 @@ var _ = Describe("PromotionStrategy Controller", func() {
 	})
 
 	Context("When reconciling a resource with a proposed commit status", func() {
-		It("should successfully reconcile the resource", func() {
+		FIt("should successfully reconcile the resource", func() {
```
crenshaw-dev:
```suggestion
		It("should successfully reconcile the resource", func() {
```

zachaller: lol thanks

internal/utils/utils.go
```diff
@@ -222,3 +227,68 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+	logger.Info(fmt.Sprintf("Reconciling %s End", obj.GetObjectKind().GroupVersionKind().Kind), "duration", time.Since(startTime))
+
+	conditions := obj.GetConditions()
+	if conditions == nil {
+		logger.Error(errors.New("conditions is nil"), "Cannot update status conditions")
+		return
+	}
+
+	if *err != nil {
+		if !k8serrors.IsConflict(*err) {
```
crenshaw-dev:
```suggestion
		// Conflicts are frequent and expected, so don't bother emitting an event for those cases.
		if !k8serrors.IsConflict(*err) {
```

internal/utils/utils.go
```diff
@@ -222,3 +227,68 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+		return
+	}
+
+	logger.Info(fmt.Sprintf("Reconciling %s End", obj.GetObjectKind().GroupVersionKind().Kind), "duration", time.Since(startTime))
+
+	conditions := obj.GetConditions()
+	if conditions == nil {
+		logger.Error(errors.New("conditions is nil"), "Cannot update status conditions")
+		return
+	}
```
crenshaw-dev: Wouldn't we want to just initialize the conditions field?

internal/utils/utils.go
```diff
@@ -222,3 +227,68 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+	err *error,
+) {
+	logger := log.FromContext(ctx)
+
+	if obj.GetName() == "" && obj.GetNamespace() == "" {
+		logger.V(4).Info(obj.GetObjectKind().GroupVersionKind().Kind + " not found, skipping reconciliation")
+		return
+	}
+
+	logger.Info(fmt.Sprintf("Reconciling %s End", obj.GetObjectKind().GroupVersionKind().Kind), "duration", time.Since(startTime))
```
crenshaw-dev: Maybe move before the name/namespace empty check?

internal/utils/utils.go
```diff
@@ -222,3 +227,68 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+
+	if *err != nil {
+		if !k8serrors.IsConflict(*err) {
+			recorder.Eventf(obj, "Warning", "ReconcileError", "Reconciliation failed: %v", *err)
+		}
+		updateErr := updateReadyCondition(ctx, obj, client, conditions, metav1.ConditionFalse, string(promoterConditions.ReconciliationError), fmt.Sprintf("Reconciliation failed: %s", *err))
+		if updateErr != nil {
+			*err = fmt.Errorf("failed to update status with error condition: %w", updateErr)
+		}
+	} else {
```
crenshaw-dev: Maybe return early and avoid the nesting?

internal/utils/utils.go
```diff
@@ -222,3 +227,68 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+	startTime time.Time,
+	obj StatusConditionUpdater,
+	client client.Client,
+	recorder record.EventRecorder,
+	err *error,
+) {
+	logger := log.FromContext(ctx)
+
+	if obj.GetName() == "" && obj.GetNamespace() == "" {
+		logger.V(4).Info(obj.GetObjectKind().GroupVersionKind().Kind + " not found, skipping reconciliation")
```
crenshaw-dev, line 251:
```suggestion
		// This happens when the Get in the reconciliation log returns "not found." It's expected and safe to skip.
		logger.V(4).Info(obj.GetObjectKind().GroupVersionKind().Kind + " not found, skipping reconciliation")
```

internal/utils/utils.go
```diff
@@ -222,3 +227,68 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+		updateErr := updateReadyCondition(ctx, obj, client, conditions, metav1.ConditionTrue, string(promoterConditions.ReconciliationSuccess), "Reconciliation succeeded")
+		if updateErr != nil {
+			*err = fmt.Errorf("failed to update status with success condition: %w", updateErr)
+		}
+	}
+}
+
+func updateReadyCondition(ctx context.Context, obj StatusConditionUpdater, client client.Client, conditions *[]metav1.Condition, status metav1.ConditionStatus, reason, message string) error {
+	condition := metav1.Condition{
+		Type:               "Ready",
```
crenshaw-dev:
```suggestion
		Type:               string(conditions.Ready),
```

internal/utils/utils.go
```diff
@@ -222,3 +227,68 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+
+		updateErr := updateReadyCondition(ctx, obj, client, conditions, metav1.ConditionTrue, string(promoterConditions.ReconciliationSuccess), "Reconciliation succeeded")
+		if updateErr != nil {
+			*err = fmt.Errorf("failed to update status with success condition: %w", updateErr)
+		}
+	}
+}
+
+func updateReadyCondition(ctx context.Context, obj StatusConditionUpdater, client client.Client, conditions *[]metav1.Condition, status metav1.ConditionStatus, reason, message string) error {
+	condition := metav1.Condition{
```
crenshaw-dev, line 277: Need to add LastTransitionTime?

internal/types/conditions/conditions.go
```diff
@@ -0,0 +1,14 @@
+package conditions
```
crenshaw-dev, line 1: Wanna add a status-conditions.md docs page to describe what's here?

crenshaw-dev (changes requested):

internal/utils/utils.go
zachaller: No it auto get's added by SetStatusCondition

internal/utils/utils.go
```diff
@@ -222,3 +227,69 @@ func AreCommitStatusesPassing(commitStatuses []promoterv1alpha1.ChangeRequestPol
...
+	startTime time.Time,
+	obj StatusConditionUpdater,
+	client client.Client,
+	recorder record.EventRecorder,
+	err *error,
+) {
+	logger := log.FromContext(ctx)
+
+	if obj.GetName() == "" && obj.GetNamespace() == "" {
+		// This happens when the Get in the reconciliation log returns "not found." It's expected and safe to skip.
```
crenshaw-dev:
```suggestion
		// This happens when the Get in the Reconcile function returns "not found." It's expected and safe to skip.
```

typo, this is better anyway :-P

internal/types/conditions/conditions.go
zachaller: did it as spec docs?

crenshaw-dev: Ah sure that works. Looks like it needs to be duplicated across the different kinds and some detail added for possible Reason values. (And maybe Status values, just to be nice to people who don't wanna go over to the metav1 docs)

crenshaw-dev: Nvm you have those values, so just need coverage for different kinds.

zachaller: Yea, I think spec docs work for now. If we add more conditions it probably deserves a page right now though it's the same thing for every kind. We can add it to every kind or list the main ones that actually reconcile in the docs?

crenshaw-dev: I'd just add it to every kind. If that page gets too long, we can split statuses into their own collapsible sections.

zachaller: We don't have status field on any other kind besides promotion strategy, so it's just on that if we want statuses on all kinds should do it in another PR.

PR #370, opened 2025-06-30
chore(refactor): simplify git path logic by removing 'context' from key

crenshaw-dev:
Tests pass, but maybe that's not enough?

My motivation is that I want to replace my errgroup-based ls-remote parallelization with a single ls-remote call that fetches shas for all branches. To do that, I need to initialize a `git.Operations`, which requires the `pathContext` parameter. That param is currently set to a branch name. But if I'm acting on multiple branches, it's not clear what branch name to choose.

I assume the original intent was to have one clone for each branch so that concurrent operations wouldn't step on each other. But we have "checkout" calls in 5 different places in git.go, so we're definitely not sticking to the branch-per-Operation rule.

I think we should simplify. Have one clone per repo. As long as we're reconciling in sequence, that's fine. Once we start reconciling in parallel, we can add locking around receivers that modify file state.


crenshaw-dev:
I'm gonna refactor a different way

PR #386, opened 2025-07-03
fix: make sure zap flags get loaded into cobra

zachaller:


cmd/main.go
```diff
@@ -309,10 +302,28 @@ func newDashboardCommand() *cobra.Command {
...
+
 	cmd := &cobra.Command{
 		Use:   "promoter",
 		Short: "GitOps Promoter",
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+		},
 	}
+
+	// Create a temporary flag.FlagSet to bind zap options
```
crenshaw-dev:
```suggestion
	// Zap only operates on go-type flags. Cobra doesn't give us direct access to those flags.
	// So we apply the zap flags to a temp go flags set and then transfer them to the cobra flags.
```

crenshaw-dev (approved):

PR #387, opened 2025-07-03
chore: add generation change predicates to all controllers

crenshaw-dev:


internal/controller/changetransferpolicy_controller.go
```diff
@@ -140,9 +142,15 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, fmt.Errorf("failed to update status: %w", err)
 	}
 
-	requeueDuration, err := r.SettingsMgr.GetChangeTransferPolicyRequeueDuration(ctx)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to get global promotion configuration: %w", err)
+	// If there was a direct push to the active branch, we want to requeue the reconciliation after a short delay to
+	// allow the CTP to pick up the new state of the active branch. (Merges via PRs should trigger reconciliations since
+	// the CTP controller owns the PR object and will be notified when the PR is merged.)
+	requeueDuration := 1 * time.Second
```
crenshaw-dev: 1s isn't magic, it just needs to be >0 so that a requeue actually happens.

crenshaw-dev: Alternatively, we could fast-forward the status field. That would save API calls. But it would be less reliable, since we'd have special code to handle that case instead of relying on the normal reconcile loop.

PR #392, opened 2025-07-07
feat: Promoter Dashboard UI and Extension

Shirly8:
### 1. New UI
- Introduces a fully new, modular UI for GitOps Promoter.

**Structure includes:**
1. ui/components-lib — reusable components
2. ui/dashboard — main dashboard app
3. ui/extension — ArgoCD extension
4. ui/shared — shared logic, types, and utilities

### Webserver SSE Fix
- Added middleware since the /watch endpoint (for Server-Sent Events) was never gzipped.
- This prevents browsers from receiving invalid (compressed) SSE streams and ensures /watch delivers text/event-stream required for real-time updates


internal/webserver/server.go
```diff
@@ -0,0 +1,506 @@
...
+		return fmt.Errorf("failed to create controller: %w", err)
+	}
+	return nil
+}
+
+func (ws *WebServer) Start(ctx context.Context, addr string) error {
+	router := gin.New()
+	router.Use(ginlogr.Ginlogr(logger, time.RFC3339, true))
+	router.Use(ginlogr.RecoveryWithLogr(logger, time.RFC3339, true, true))
+	router.Use(NoGzipForSSE())
```
zachaller: We should be able to use compression

Shirly8: EventSource was not able to parse the SSE with the compression (on the client side). gzip supports exclusion just for /watch endpoint but I could look into it if bandwidth or load is a concern.

config/manager/dashboard-deployment.yaml
```diff
@@ -0,0 +1,92 @@
+
```
zachaller: I don't think I want to promote the dashboard as a hosted service, it leads to ideas that we will support various auth and other things not sure we want to go this route yet. Thinking...

web/embed.go
```diff
@@ -0,0 +1,6 @@
+package web
```
zachaller: can we move this to the internal/webserver package?

Shirly8: The embed directive would need to reference the static directory. so we'd have to move static files in there?

config/manager/dashboard-deployment.yaml
Shirly8: I agree. The dashboard is still embedded in the binary but removing the Kubernetes deployment.

cmd/main.go
```diff
@@ -323,12 +324,44 @@ func runController(
...
+			}
+
+			// Start manager in background
+			go func() {
+				if err := mgr.Start(ctx); err != nil {
+					panic(err)
+				}
+			}()
+
+			return ws.StartDashboard(ctx, ":8080")
```
zachaller: Can we make this port configurable via cli flag, as well as print a log line that takes us to `http://localhost:<port>` so that we can click it once we are started

web/embed.go
zachaller: Yup that make sesne, does it work under the UI directory?

ui/extension/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
zachaller: Can we document with a README.md how to build this extension for ArgoCD

web/embed.go
Shirly8: Wouldn't moving to UI directory mix up the Go code (web/embed.go) with JavaScript project structure. Keeping it in web/ also makes build process easier.

PR #399, opened 2025-07-08
feat: added PR status in CTP and PS

zachaller:


api/v1alpha1/changetransferpolicy_types.go
```diff
@@ -124,6 +127,12 @@ type ChangeTransferPolicyStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 }
 
+type PullRequestCommonStatus struct {
+	ID             string           `json:"id,omitempty"`
+	State          PullRequestState `json:"state,omitempty"`
```
crenshaw-dev, line 133: Probably oughta add enum validation to this.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -380,6 +385,35 @@ func (r *ChangeTransferPolicyReconciler) setCommitStatusState(ctx context.Contex
 	return nil
 }
 
+func (r *ChangeTransferPolicyReconciler) setPullRequestState(ctx context.Context, ctp *promoterv1alpha1.ChangeTransferPolicy) error {
+	pr := &promoterv1alpha1.PullRequestList{}
+	err := r.List(ctx, pr, &client.ListOptions{LabelSelector: labels.SelectorFromSet(map[string]string{
+		promoterv1alpha1.PromotionStrategyLabel:    utils.KubeSafeLabel(ctp.Labels[promoterv1alpha1.PromotionStrategyLabel]),
+		promoterv1alpha1.ChangeTransferPolicyLabel: utils.KubeSafeLabel(ctp.Name),
+		promoterv1alpha1.EnvironmentLabel:          utils.KubeSafeLabel(ctp.Spec.ActiveBranch),
+	})})
```
crenshaw-dev, line 394: The PR name should be inferable based on the logic in `creatOrUpdatePullRequest`. Should be able to just pull that logic into a util function and then do a `Get`.

zachaller: We already have a function for it: 

```
func GetPullRequestName(repoOwner, repoName, pcProposedBranch, pcActiveBranch string) string {
```

it requires repo owner which we do not have so we can make a call to get that, however I also kind of sort of want to remove the fixed name thing in general and use label lookups we do label look ups in other places as well

api/v1alpha1/changetransferpolicy_types.go
zachaller: done

crenshaw-dev (approved):

PR #401, opened 2025-07-08
feat: add url to pr status

zachaller:


internal/scms/fake/pullrequest.go
```diff
@@ -145,11 +147,18 @@ func (pr *PullRequest) Merge(ctx context.Context, commitMessage string, pullRequ
...
 	found, id := pr.findOpen(ctx, pullRequest)
 	mutexPR.RUnlock()
-	return found, id, nil
+
+	prState := v1alpha1.PullRequestCommonStatus{
+		ID:             id,
+		State:          v1alpha1.PullRequestOpen,
+		Url:            fmt.Sprintf("http://localhost:5000/%s/%s/pull/%s", pullRequest.Spec.RepositoryReference.Name, pullRequest.Spec.SourceBranch, id),
+		PRCreationTime: metav1.Now(),
+	}
```
crenshaw-dev, line 160: A lot of these functions mutate the PullRequest in-place instead of returning info. Should we standardize on one or the other?

internal/scms/forgejo/pullrequest.go
```diff
@@ -199,27 +203,36 @@ func (pr *PullRequest) FindOpen(ctx context.Context, prObj *promoterv1alpha1.Pul
...
+		prState, err := forgejoPullRequestStateToPullRequestState(*prItem)
 		if err != nil {
-			return false, "", err
+			return false, promoterv1alpha1.PullRequestCommonStatus{}, err
 		}
 
-		prObj.Status.ID = strconv.FormatInt(pr.Index, 10)
-		prObj.Status.State = prState
-		return true, prObj.Status.ID, nil
+		url, err := pr.GetUrl(ctx, prObj)
```
crenshaw-dev, line 221: I think all these calls are gonna already have a `gitRepo` on the context. Maybe just pass it as an arg so the receiver doesn't have to do another Get?

internal/scms/fake/pullrequest.go
zachaller: They should not be mutating any more, the idea of this switch was to remove the mutation of the PullRequest from the FindOpen because ideally I want the controller code to do that only. Let me try to change the arg from a pointer to prevent it.

crenshaw-dev: Gotcha. Maybe even narrow it down to `sourceBranch, targetBranch string` so we're not passing a large struct. But wouldn't bend over backwards to get there.

zachaller: we use namespace and other things this would be too big a change, if we want to can do it in a different PR

internal/scms/forgejo/pullrequest.go
zachaller: They don't and it's quite a path to pass it in as well, if we want to go this route we can but in another PR

internal/scms/github/pullrequest.go
```diff
@@ -201,10 +203,35 @@ func (pr *PullRequest) FindOpen(ctx context.Context, pullRequest *v1alpha1.PullR
...
-		pullRequest.Status.ID = strconv.Itoa(*pullRequests[0].Number)
-		pullRequest.Status.State = v1alpha1.PullRequestState(*pullRequests[0].State)
-		return true, pullRequest.Status.ID, nil
+		url, err := pr.GetUrl(ctx, pullRequest)
+		if err != nil {
+			return false, v1alpha1.PullRequestCommonStatus{}, fmt.Errorf("failed to get pull request URL: %w", err)
+		}
+
+		prStatus := v1alpha1.PullRequestCommonStatus{
+			ID:             strconv.Itoa(*pullRequests[0].Number),
```
crenshaw-dev, line 221: Wanna nil check?

internal/scms/gitlab/pullrequest.go
```diff
@@ -255,10 +257,33 @@ func (pr *PullRequest) FindOpen(ctx context.Context, prObj *v1alpha1.PullRequest
...
+		url, err := pr.GetUrl(ctx, prObj)
+		if err != nil {
+			return false, v1alpha1.PullRequestCommonStatus{}, fmt.Errorf("failed to get pull request URL: %w", err)
+		}
+
+		pullRequestStatus := v1alpha1.PullRequestCommonStatus{
+			ID:             strconv.Itoa(mrs[0].IID),
+			State:          mapMergeRequestState(mrs[0].State),
+			Url:            url,
+			PRCreationTime: metav1.Time{Time: *mrs[0].CreatedAt},
```
crenshaw-dev, line 269: Nil check?

internal/scms/github/pullrequest.go
```diff
@@ -201,10 +203,35 @@ func (pr *PullRequest) FindOpen(ctx context.Context, pullRequest *v1alpha1.PullR
...
+	return false, v1alpha1.PullRequestCommonStatus{}, nil
+}
+
+// GetUrl returns a hard coded URL for the pull request.
+func (pr *PullRequest) GetUrl(ctx context.Context, pullRequest v1alpha1.PullRequest) (string, error) {
+	gitRepo, _ := utils.GetGitRepositoryFromObjectKey(ctx, pr.k8sClient, client.ObjectKey{Namespace: pullRequest.Namespace, Name: pullRequest.Spec.RepositoryReference.Name})
+
+	prNumber, err := strconv.Atoi(pullRequest.Status.ID)
+	if err != nil {
+		return "", fmt.Errorf("failed to convert PR number to int when generating pull reqest url: %w", err)
```
crenshaw-dev:
```suggestion
		return "", fmt.Errorf("failed to convert PR number to int when generating pull request url: %w", err)
```

internal/scms/forgejo/pullrequest.go
```diff
@@ -199,30 +203,39 @@ func (pr *PullRequest) FindOpen(ctx context.Context, prObj *promoterv1alpha1.Pul
...
+		if err != nil {
+			return false, promoterv1alpha1.PullRequestCommonStatus{}, fmt.Errorf("failed to get pull request URL: %w", err)
+		}
+
+		pullRequestStatus := promoterv1alpha1.PullRequestCommonStatus{
+			ID:             strconv.FormatInt(prItem.Index, 10),
+			State:          prState,
+			Url:            url,
+			PRCreationTime: metav1.Time{Time: *prItem.Created},
```
crenshaw-dev, line 230: Nil check?

crenshaw-dev (changes requested):

crenshaw-dev (approved):

PR #427, opened 2025-07-17
fix(acdcs): requeue to observe health after transition threshold

crenshaw-dev:
If we observe an app as being healthy within 5s of it becoming healthy, we currently refuse to accept that health state, just in case it flaps.

But we don't currently requeue to check the app again after the 5s threshold has been reached.

This change requeues the ArgoCDCommitStatus after the threshold will be reached for all encountered healthy apps.

If the default requeue duration is shorter, we requeue after that instead.


internal/controller/promotionstrategy_controller_test.go
```diff
@@ -1963,7 +1958,7 @@ var _ = Describe("PromotionStrategy Controller", func() {
 				g.Expect(err).To(HaveOccurred())
 				g.Expect(errors.IsNotFound(err)).To(BeTrue())
 			}, constants.EventuallyTimeout).Should(Succeed())
-			Expect(waitedForDelay).To(BeTrue())
+			Expect(time.Since(lastTransitionTime.Time) >= 5*time.Second).To(BeTrue(), "We should have waited for the delay to ensure the application health wasn't flapping.")
```
zachaller: can we use the constant here `lastTransitionTimeThreshold`

zachaller (approved):

PR #428, opened 2025-07-17
feat: url template for commit statuses

FourFifthsCode:
Add `urlTemplate` field to ArgoCDCommitStatus CRD.

URLTemplate generates the URL to use in the CommitStatus, for example a link to the Argo CD UI. The template
is a go text template and receives .Environment and .ArgoCDCommitStatus variables. A function called urlQueryEscape
is available to escape url query parameters.

Example:
```go
{{- $baseURL := "https://dev.argocd.local" -}}
{{- if eq .Environment "environment/development" -}}
{{- $baseURL = "https://dev.argocd.local" -}}
{{- else if eq .Environment "environment/staging" -}}
{{- $baseURL = "https://staging.argocd.local" -}}
{{- else if eq .Environment "environment/production" -}}
{{- $baseURL = "https://prod.argocd.local" -}}
{{- end -}}
{{- $labels := "" -}}
{{- range $key, $value := .ArgoCDCommitStatus.Spec.ApplicationSelector.MatchLabels -}}
{{- $labels = printf "%s%s=%s," $labels $key $value -}}
{{- end -}}
{{ printf "%s/applications?labels=%s" $baseURL (urlQueryEscape $labels) }}
```

fix: multi-cluster kubeconfig provider key and docs

### DIFF
--- a/api/v1alpha1/argocdcommitstatus_types.go
+++ b/api/v1alpha1/argocdcommitstatus_types.go
@@ -32,6 +32,29 @@ type ArgoCDCommitStatusSpec struct {
 	// ApplicationSelector is a label selector that selects the Argo CD applications to which this commit status applies.
 	// +kubebuilder:validation:Required
 	ApplicationSelector *metav1.LabelSelector `json:"applicationSelector,omitempty"`
+
+	// URLTemplate generates the URL to use in the CommitStatus, for example a link to the Argo CD UI. The template
+	// is a go text template and receives .Environment and .ArgoCDCommitStatus variables. A function called urlQueryEscape
+	// is available to escape url query parameters.
+	//
+	// Example:
+	//
+	// {{- $baseURL := "https://dev.argocd.local" -}}
+	// {{- if eq .Environment "environment/development" -}}
+	// {{- $baseURL = "https://dev.argocd.local" -}}
+	// {{- else if eq .Environment "environment/staging" -}}
+	// {{- $baseURL = "https://staging.argocd.local" -}}
+	// {{- else if eq .Environment "environment/production" -}}
+	// {{- $baseURL = "https://prod.argocd.local" -}}
+	// {{- end -}}
+	// {{- $labels := "" -}}
+	// {{- range $key, $value := .ArgoCDCommitStatus.Spec.ApplicationSelector.MatchLabels -}}
+	// {{- $labels = printf "%s%s=%s," $labels $key $value -}}
+	// {{- end -}}
+	// {{ printf "%s/applications?labels=%s" $baseURL (urlQueryEscape $labels) }}
+	//
+	// +kubebuilder:validation:Optional
+	URLTemplate string `json:"urlTemplate,omitempty"`
 }
 
 // ArgoCDCommitStatusStatus defines the observed state of ArgoCDCommitStatus.
@@ -68,6 +91,8 @@ type ApplicationsSelected struct {
 	LastTransitionTime *metav1.Time `json:"lastTransitionTime"`
 	// Environment is the syncSource.targetBranch of the Argo CD application (in effect, its environment).
 	Environment string `json:"environment,omitempty"`
+	// ClusterName is the name of the cluster that the application manifest is deployed to.
+	ClusterName string `json:"clusterName"`
 }
 
 // +kubebuilder:object:root=true
--- a/docs/commit-status-controllers/argocd.md
+++ b/docs/commit-status-controllers/argocd.md
@@ -53,6 +53,21 @@ spec:
     name: webservice-tier-1
 ```
 
+### Commit Status URL Template
+To configure setting the url of a commit status, for example, a link to ArgoCD instance, set the 'urlTemplate' field. The template uses [Go templates](https://pkg.go.dev/text/template) syntax and most [sprig](https://masterminds.github.io/sprig/) functions are supported as well as an additional 'urlQueryEscape' function for escaping url query parameters. The template receives `.Environment` and `.ArgoCDCommitStatus` variables. 
+
+This example template generates an ArgoCD link in a multi-cluster setup that filters applications by label selector and environment. 
+
+```yaml
+apiVersion: promoter.argoproj.io/v1alpha1
+kind: ArgoCDCommitStatus
+metadata:
+  name: webservice-tier-1
+spec:
+  urlTemplate: |
+    {!docs/example-resources/ArgoCDCommitStatusURL.gotmpl!}
+```
+
 ## Multi-Cluster Support
 
 GitOps promoter can monitor Argo CD Applications across multiple Kubernetes clusters. This is particularly useful when managing promotions across environments that use different Argo CD instances.
@@ -62,9 +77,9 @@ GitOps promoter can monitor Argo CD Applications across multiple Kubernetes clus
 To enable multi-cluster support, you need to configure two components:
 
 #### Kubeconfig Secret
-   - Create a secret with key `kubeconfig` containing a standard `~/.kube/config` file
+   - Create a secret with key `kubeconfig` containing a standard `~/.kube/config` file as its value and label `sigs.k8s.io/multicluster-runtime-kubeconfig: "true"`
    - The secret must be created in the same namespace where gitops-promoter runs
-   - The controller uses the `current context` from the kubeconfig to determine which cluster to uses
+   - The controller uses the `current context` from the kubeconfig to determine which cluster to use
      
     !!! note
         Remove any additional clusters from the `kubeconfig` as they will be ignored
--- a/internal/controller/argocdcommitstatus_controller.go
+++ b/internal/controller/argocdcommitstatus_controller.go
@@ -87,6 +87,11 @@ type ArgoCDCommitStatusReconciler struct {
 	localClient        client.Client
 }
 
+type URLTemplateData struct {
+	Environment        string
+	ArgoCDCommitStatus promoterv1alpha1.ArgoCDCommitStatus
+}
+
 // +kubebuilder:rbac:groups=promoter.argoproj.io,resources=argocdcommitstatuses,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=promoter.argoproj.io,resources=argocdcommitstatuses/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=promoter.argoproj.io,resources=argocdcommitstatuses/finalizers,verbs=update
@@ -125,7 +130,7 @@ func (r *ArgoCDCommitStatusReconciler) Reconcile(ctx context.Context, req mcreco
 		return ctrl.Result{}, fmt.Errorf("failed to parse label selector: %w", err)
 	}
 	// TODO: we should setup a field index and only list apps related to the currently reconciled app
-	var apps argocd.ApplicationList
+	apps := make(map[string]argocd.ApplicationList)
 
 	// list clusters so we can query argocd applications from all clusters
 	clusters := r.KubeConfigProvider.ListClusters()
@@ -146,10 +151,14 @@ func (r *ArgoCDCommitStatusReconciler) Reconcile(ctx context.Context, req mcreco
 			return ctrl.Result{}, fmt.Errorf("failed to list ArgoCDApplications: %w", err)
 		}
 
-		apps.Items = append(apps.Items, clusterArgoCDApps.Items...)
+		apps[clusterName] = clusterArgoCDApps
 	}
 
-	logger.V(4).Info("Found Applications", "appCount", len(apps.Items))
+	appCount := 0
+	for _, clusterApps := range apps {
+		appCount += len(clusterApps.Items)
+	}
+	logger.V(4).Info("Found Applications", "appCount", appCount)
 
 	promotionStrategy := promoterv1alpha1.PromotionStrategy{}
 	err = r.localClient.Get(ctx, client.ObjectKey{Namespace: argoCDCommitStatus.Namespace, Name: argoCDCommitStatus.Spec.PromotionStrategyRef.Name}, &promotionStrategy)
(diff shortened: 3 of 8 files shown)


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -460,10 +473,24 @@ func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.
...
 	}
 
+	// Render URL from template if it exists, but don't block commit status update if it fails
+	if argoCDCommitStatus.Spec.URLTemplate != "" {
+		data := URLTemplateData{
+			Environment:        targetBranch,
+			ArgoCDCommitStatus: argoCDCommitStatus,
+		}
+		url, err := utils.RenderStringTemplate(argoCDCommitStatus.Spec.URLTemplate, data)
+		if err == nil {
```
zachaller: can we switch this logic around and return the error?

FourFifthsCode: I don't think we want to prevent a commit status from being created when a template is bad, figured it would be better to log the error and move on

zachaller: I think we actually might want that, we will still block the promotion without the commit status and by returing the error we will bubble that up to the user via conditions and an event which will let them know of the issue vs just it being in the logs?

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -460,10 +473,24 @@ func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.
...
+			logger.Info("Rendered URL template", "url", url, "environment", targetBranch, "commitStatus", desiredCommitStatus.Name, "namespace", desiredCommitStatus.Namespace)
+			desiredCommitStatus.Spec.Url = url
```
crenshaw-dev: We should parse the URL and confirm that its schema is either http or https (and document that limitation).

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: ah, that's a good point, would be good to update conditions and event

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: that's a great idea, wasn't sure exactly how to validate but I really like this approach

internal/controller/argocdcommitstatus_controller.go
zachaller: Those will happen automatically on error return

crenshaw-dev: URLTemplateSuccess might be an interesting standalone condition. Not for this PR but eventually.

FourFifthsCode: ah ok, I'll just return the error then 😄

internal/controller/promotionstrategy_controller.go
```diff
@@ -256,6 +256,12 @@ func (r *PromotionStrategyReconciler) createOrUpdatePreviousEnvironmentCommitSta
 	gvk := promoterv1alpha1.GroupVersion.WithKind(kind)
 	controllerRef := metav1.NewControllerRef(ctp, gvk)
 
+	// If there is only one commit status, use the URL from that commit status.
+	var url string
+	if len(previousCRPCSPhases) == 1 {
+		url = previousCRPCSPhases[0].Url
+	}
+
```
FourFifthsCode, line 264: This is where we are picking a commit status url if there is only one status, want to look at how to support multiple commit status URLs in the future.

api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -32,6 +32,29 @@ type ArgoCDCommitStatusSpec struct {
...
+	// {{- else if eq .Environment "environment/staging" -}}
+	// {{- $baseURL = "https://staging.argocd.local" -}}
+	// {{- else if eq .Environment "environment/production" -}}
+	// {{- $baseURL = "https://prod.argocd.local" -}}
+	// {{- end -}}
+	// {{- $labels := "" -}}
+	// {{- range $key, $value := .ArgoCDCommitStatus.Spec.ApplicationSelector.MatchLabels -}}
+	// {{- $labels = printf "%s%s=%s," $labels $key $value -}}
+	// {{- end -}}
+	// {{ printf "%s/applications?labels=%s" $baseURL (urlQueryEscape $labels) }}
```
crenshaw-dev: If you indent the code by 4 spaces, you get really cool godocs.  :-)

docs/commit-status-controllers/argocd.md
```diff
@@ -53,6 +53,24 @@ spec:
     name: webservice-tier-1
 ```
 
+### Commit Status URL Template
+To configure setting the url of a commit status, for example, a link to ArgoCD instance, set the 'urlTemplate' field. The template uses [Go templates](https://pkg.go.dev/text/template) syntax and most [sprig](https://masterminds.github.io/sprig/) functions are supported as well as an additional 'urlQueryEscape' function for escaping url query parameters. The template receives `.Environment` and `.ArgoCDCommitStatus` variables.
```
crenshaw-dev:
```suggestion
To configure setting the url of a commit status, for example, a link to an Argo CD instance, set the 'urlTemplate' field. The template uses [Go templates](https://pkg.go.dev/text/template) syntax and most [Sprig](https://masterminds.github.io/sprig/) functions (excluding `env`, `expandenv` and `getHostByName`) are supported as well as an additional 'urlQueryEscape' function for escaping url query parameters. The template receives `.Environment` and `.ArgoCDCommitStatus` variables. 
```

docs/commit-status-controllers/argocd.md
```diff
@@ -53,6 +53,24 @@ spec:
     name: webservice-tier-1
 ```
 
+### Commit Status URL Template
+To configure setting the url of a commit status, for example, a link to ArgoCD instance, set the 'urlTemplate' field. The template uses [Go templates](https://pkg.go.dev/text/template) syntax and most [sprig](https://masterminds.github.io/sprig/) functions are supported as well as an additional 'urlQueryEscape' function for escaping url query parameters. The template receives `.Environment` and `.ArgoCDCommitStatus` variables. 
+
+!!! note 
+    The rendered URL must use a scheme of either 'http' or 'https'
+
+This example template generates an ArgoCD link in a multi-cluster setup that filters applications by label selector and environment.
```
crenshaw-dev:
```suggestion
This example template generates an Argo CD link in a multi-cluster setup that filters applications by label selector and environment. 
```

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -460,10 +474,41 @@ func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.
...
+		if err != nil {
+			logger.Error(err, "failed to parse URL", "url", renderedURL, "environment", targetBranch, "commitStatus", desiredCommitStatus.Name, "namespace", desiredCommitStatus.Namespace)
+			return fmt.Errorf("failed to parse URL: %w", err)
+		}
+
+		// Check that the URL scheme is http or https
+		if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
+			err = fmt.Errorf("URL scheme is not http or https: %s", parsedURL.Scheme)
+			logger.Error(err, "URL scheme is not http or https", "url", parsedURL, "environment", targetBranch, "commitStatus", desiredCommitStatus.Name, "namespace", desiredCommitStatus.Namespace)
+			return err
```
```suggestion
			return fmt.Errorf("URL scheme is not http or https: %s", parsedURL.Scheme)
```

No need to log, since this will get returned up and logged as part of the reconciliation failure anyway.

internal/controller/promotionstrategy_controller_test.go
```diff
@@ -1987,6 +1986,10 @@ var _ = Describe("PromotionStrategy Controller", func() {
 				},
 			}
 
+			// Read URL template from example-resources
+			template, err := os.ReadFile("../../docs/example-resources/ArgoCDCommitStatusURL.gotmpl")
```
crenshaw-dev: Would it be possible to embed instead of reading the file at runtime?

docs/commit-status-controllers/argocd.md
crenshaw-dev: Maybe add a link to the `url.QueryEscape` docs too?

crenshaw-dev: Might also be good to explicitly say that `.Environment` is a string holding the environment name (i.e. environment branch name) for the group of Applications the URL is being generated for.

Probably should also link to the `ArgoCDCommitStatus` CRD docs and explain that `.ArgoCDCommitStatus` holds the whole CR in its current state.

docs/commit-status-controllers/argocd.md
```diff
@@ -53,6 +53,24 @@ spec:
...
+This example template generates an ArgoCD link in a multi-cluster setup that filters applications by label selector and environment. 
+
+```yaml
+apiVersion: promoter.argoproj.io/v1alpha1
+kind: ArgoCDCommitStatus
+metadata:
+  name: webservice-tier-1
+spec:
+  urlTemplate: |
+    {!docs/example-resources/ArgoCDCommitStatusURL.gotmpl!}
```
crenshaw-dev: I like having the multi-cluster example here. Can we also add a single-cluster example that we can plan to keep in sync with Intuit's running template? That'll let us make sure the docs always have a production-tested template available for folks to build on.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -125,7 +132,7 @@ func (r *ArgoCDCommitStatusReconciler) Reconcile(ctx context.Context, req mcreco
 		return ctrl.Result{}, fmt.Errorf("failed to parse label selector: %w", err)
 	}
 	// TODO: we should setup a field index and only list apps related to the currently reconciled app
-	var apps argocd.ApplicationList
+	apps := make(map[string]argocd.ApplicationList)
```
crenshaw-dev: Instead of structuring this as a map, I'd create a new struct called something like

```go
struct ApplicationsInEnvironment {
    argocd.ApplicationList
    ClusterName string
}
```

We're not really using the clusterName "key" to group any processing. We just need the data.

Using a struct communicates that intent better and also leaves us room to expand the data structure if we want to accommodate extra info in the future (such as a cluster URL).

crenshaw-dev (changes requested):
Looking really good! Relatively small comments. 

One other thing that should be done is updating the `sortApplicationsSelected` function to sort by cluster name (after environment but before namespace).

api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -54,7 +55,36 @@ type ArgoCDCommitStatusSpec struct {
 	// {{ printf "%s/applications?labels=%s" $baseURL (urlQueryEscape $labels) }}
 	//
 	// +kubebuilder:validation:Optional
-	URLTemplate string `json:"urlTemplate,omitempty"`
+	URLTemplate GoTemplate `json:"urlTemplate,omitempty"`
```
crenshaw-dev: Would rename this just `URL`.

api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -54,7 +55,36 @@ type ArgoCDCommitStatusSpec struct {
...
+}
+
+// GoTemplate is a template that can be rendered using the Go template engine.
+type GoTemplate struct {
```
crenshaw-dev: Maybe just call it `URL` or `URLConfig`?

api/v1alpha1/argocdcommitstatus_types.go
FourFifthsCode: That's cool!
\

internal/controller/promotionstrategy_controller_test.go
FourFifthsCode: ended up not using the gotmp file

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: make sense to me!

internal/controller/argocdcommitstatus_controller.go
FourFifthsCode: roger that

docs/commit-status-controllers/argocd.md
FourFifthsCode: Added more examples :)

docs/commit-status-controllers/argocd.md
FourFifthsCode: Good idea, added more details and examples

docs/commit-status-controllers/argocd.md
```diff
@@ -53,6 +53,62 @@ spec:
...
+- `missingkey=zero` - The operation returns the zero value for the map type's element.
+- `missingkey=error` - Execution stops immediately with an error.
+
+```yaml
+apiVersion: promoter.argoproj.io/v1alpha1
+kind: ArgoCDCommitStatus
+metadata:
+  name: webservice-tier-1
+spec:
+  urlTemplate:
```
crenshaw-dev:
```suggestion
  url: 
```

docs/example-resources/ArgoCDCommitStatus-URL-env.yaml
```diff
@@ -0,0 +1,26 @@
...
+      {{- $baseURL = "https://staging.argocd.local" -}}
+      {{- else if eq .Environment "environment/production" -}}
+      {{- $baseURL = "https://prod.argocd.local" -}}
+      {{- end -}}
+      {{- $labels := "" -}}
+      {{- range $key, $value := .ArgoCDCommitStatus.Spec.ApplicationSelector.MatchLabels -}}
+      {{- $labels = printf "%s%s=%s," $labels $key $value -}}
+      {{- end -}}
+      {{- printf "%s/applications?labels=%s" $baseURL (urlQueryEscape $labels) }}
+    options: ""
```
crenshaw-dev: Options is an array, right? Maybe just drop it from this example.

docs/example-resources/ArgoCDCommitStatus-URL.yaml
```diff
@@ -0,0 +1,19 @@
...
+    name: argocon-demo
+  url:
+    template: |
+      {{- $baseURL := "https://argocd.local" -}}
+      {{- $labels := "" -}}
+      {{- range $key, $value := .ArgoCDCommitStatus.Spec.ApplicationSelector.MatchLabels -}}
+      {{- $labels = printf "%s%s=%s," $labels $key $value -}}
+      {{- end -}}
+      {{- printf "%s/applications?labels=%s" $baseURL (urlQueryEscape $labels) }}
+    options: ""
```
crenshaw-dev: Drop from this example

docs/example-resources/ArgoCDCommitStatus.yaml
```diff
@@ -8,6 +8,22 @@ spec:
...
+      {{- $baseURL = "https://staging.argocd.local" -}}
+      {{- else if eq .Environment "environment/production" -}}
+      {{- $baseURL = "https://prod.argocd.local" -}}
+      {{- end -}}
+      {{- $labels := "" -}}
+      {{- range $key, $value := .ArgoCDCommitStatus.Spec.ApplicationSelector.MatchLabels -}}
+      {{- $labels = printf "%s%s=%s," $labels $key $value -}}
+      {{- end -}}
+      {{- printf "%s/applications?labels=%s" $baseURL (urlQueryEscape $labels) }}
+    options: missingkey=zero
```
crenshaw-dev:
```suggestion
    options: 
    - missingkey=zero
```

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -460,10 +483,39 @@ func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.
...
+	// Render URL from template if it exists, but don't block commit status update if it fails
+	if argoCDCommitStatus.Spec.URL.Template != "" {
+		data := URLTemplateData{
+			Environment:        targetBranch,
+			ArgoCDCommitStatus: argoCDCommitStatus,
+		}
+
+		renderedURL, err := utils.RenderStringTemplate(argoCDCommitStatus.Spec.URL.Template, data, argoCDCommitStatus.Spec.URL.Options...)
+		if err != nil {
+			logger.Error(err, "failed to render URL template", "argoCDCommitStatus", argoCDCommitStatus.Name, "namespace", argoCDCommitStatus.Namespace)
```
crenshaw-dev: No need to log here, the error will be logged when reconciliation fails.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -460,10 +483,39 @@ func (r *ArgoCDCommitStatusReconciler) updateAggregatedCommitStatus(ctx context.
...
+			return fmt.Errorf("failed to render URL template: %w", err)
+		}
+
+		// Parse the URL to check that it's valid
+		parsedURL, err := url.Parse(renderedURL)
+		if err != nil {
+			logger.Error(err, "failed to parse URL", "url", renderedURL, "environment", targetBranch, "commitStatus", desiredCommitStatus.Name, "namespace", desiredCommitStatus.Namespace)
```
crenshaw-dev: No need to log here, the error will be logged when reconciliation fails.

docs/commit-status-controllers/argocd.md
FourFifthsCode: Ah good catch

docs/example-resources/ArgoCDCommitStatus-URL-env.yaml
FourFifthsCode: sounds good

api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -32,6 +32,59 @@ type ArgoCDCommitStatusSpec struct {
...
+	//   {{- $baseURL = "https://staging.argocd.local" -}}
+	//   {{- else if eq .Environment "environment/production" -}}
+	//   {{- $baseURL = "https://prod.argocd.local" -}}
+	//   {{- end -}}
+	//   {{- $labels := "" -}}
+	//   {{- range $key, $value := .ArgoCDCommitStatus.Spec.ApplicationSelector.MatchLabels -}}
+	//   {{- $labels = printf "%s%s=%s," $labels $key $value -}}
+	//   {{- end -}}
+	//   {{ printf "%s/applications?labels=%s" $baseURL (urlQueryEscape $labels) }}
+	//
```
crenshaw-dev: Move to URLConfig.Template doc string

api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -32,6 +32,59 @@ type ArgoCDCommitStatusSpec struct {
...
+	URL URLConfig `json:"url,omitempty"`
+}
+
+// URLConfig is a template that can be rendered using the Go template engine.
+type URLConfig struct {
+	// Template is the template to use.
+	// +kubebuilder:validation:Optional
+	Template string `json:"template,omitempty"`
+
+	// Option sets options for the template. Options are described by
```
crenshaw-dev:
```suggestion
	// Options sets options for the template. Options are described by
```

internal/controller/promotionstrategy_controller_test.go
```diff
@@ -1987,6 +1986,21 @@ var _ = Describe("PromotionStrategy Controller", func() {
 				},
 			}
 
+			urlTemplate := `
```
crenshaw-dev: Can we go-embed the example YAML and extract the go template? Just trying to avoid maintaining two copies of the test/example template.

crenshaw-dev (changes requested):

internal/controller/promotionstrategy_controller.go
```diff
@@ -256,6 +256,12 @@ func (r *PromotionStrategyReconciler) createOrUpdatePreviousEnvironmentCommitSta
 	gvk := promoterv1alpha1.GroupVersion.WithKind(kind)
 	controllerRef := metav1.NewControllerRef(ctp, gvk)
 
+	// If there is only one commit status, use the URL from that commit status.
+	var url string
+	if len(previousCRPCSPhases) == 1 {
+		url = previousCRPCSPhases[0].Url
+	}
```
crenshaw-dev, line 263: Is there a good place to document this behavior?

api/v1alpha1/argocdcommitstatus_types.go
FourFifthsCode: sounds good

internal/controller/promotionstrategy_controller_test.go
FourFifthsCode: Since the mkdocs import ext doesn't have indentation methods, it became more of a hassle trying to make them work for both a test and yaml example than the gain of sharing the doc, in my opinion, I'd rather lose the benefit of testing the example and just hard code in the test.

internal/controller/promotionstrategy_controller.go
FourFifthsCode: You thinking in code comments or in the commit status mkdocs?

crenshaw-dev: gating-promotions.md has a section about the previous environment commitstatus, that seems like an okay place to call it out.

internal/controller/promotionstrategy_controller_test.go
crenshaw-dev: I might be missing a limitation, but the way I'm imagining it is:

1) go:embed the yaml that's being used in docs
2) unmarshal the yaml into an ArgoCDCommitStatus struct
3) pull the template text off the struct
4) use that template in the test

FourFifthsCode: Ah, yeah, if I just pull the whole yaml and parse it, then I don't have to worry about formatting. Before I was trying to separate the yaml and the gotempl. Alright, I'll do it that way 😄

docs/commit-status-controllers/argocd.md
```diff
@@ -53,6 +53,66 @@ spec:
...
+To configure setting the url of a commit status, for example, a link to an Argo CD instance, set the `url.template` field. The template uses [Go templates](https://pkg.go.dev/text/template) syntax and most [Sprig](https://masterminds.github.io/sprig/) functions (excluding `env`, `expandenv` and `getHostByName`) are supported as well as an additional [`urlQueryEscape`](https://pkg.go.dev/net/url#QueryEscape) function for escaping url query parameters. The template receives `.Environment` and `.ArgoCDCommitStatus` variables. 
+
+!!! important 
+    The rendered URL must use a scheme of either 'http' or 'https'
+
+#### Template Variables
+The following variables are available in the template:
+
+- `.Environment` - string holding the environment name (i.e. environment branch name) for the group of Applications the URL is being generated for.
+- `.ArgoCDCommitStatus` - holds the whole [CR](../crd-specs#argocdcommitstatus) in its current state
```
crenshaw-dev:
```suggestion
- `.ArgoCDCommitStatus` - holds the whole [CR](../../crd-specs#argocdcommitstatus) in its current state
```

crenshaw-dev (approved):

PR #434, opened 2025-07-23
docs: test unmarshaling of example resources

crenshaw-dev:
This will help avoid example resources drifting from actual CRD definitions.


FourFifthsCode (approved):
LGTM

PR #441, opened 2025-07-28
wip: add commit message when doing merge commit via PR

zachaller:


api/v1alpha1/pullrequest_types.go
```diff
@@ -44,6 +44,8 @@ type PullRequestSpec struct {
 	SourceBranch string `json:"sourceBranch"`
 	// Description is the description body of the pull/merge request
 	Description string `json:"description,omitempty"`
+	// MergeCommitMessage is the commit message to use when merging the pull request.
+	MergeCommitMessage string `json:"mergeCommitMessage,omitempty"`
```
crenshaw-dev: How about

```yaml
spec:
  commit:
    message:
```

Leaves room for fields like `type: Squash`.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -21,8 +21,11 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
+	"golang.org/x/exp/slices"
```
crenshaw-dev: You can just use the `slices` package, it's built-in now.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +163,206 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+	logger := log.FromContext(ctx)
+
+	history := make([]promoterv1alpha1.History, 0)
+	shaListActive, err := gitOperations.GetRevListFirstParent(ctx, "origin/"+ctp.Spec.ActiveBranch, "5")
+	if err != nil {
+		return fmt.Errorf("failed to get rev-list commit history for branch %q: %w", ctp.Spec.ActiveBranch, err)
+	}
+	logger.V(4).Info("Rev-list history for active branch", "shaList", shaListActive)
+
+	for _, sha := range shaListActive {
```
crenshaw-dev, line 178: There are a lot of failures here that fall back to logging instead of failing the reconcile. Is it actually necessary to be so accommodating? Can we start by aggressively failing and then carving out exceptions for specific cases?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +163,206 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
 	}, nil
 }
 
+func (r *ChangeTransferPolicyReconciler) calculateHistory(ctx context.Context, ctp *promoterv1alpha1.ChangeTransferPolicy, gitOperations *git.EnvironmentOperations) error {
+	logger := log.FromContext(ctx)
+
+	history := make([]promoterv1alpha1.History, 0)
+	shaListActive, err := gitOperations.GetRevListFirstParent(ctx, "origin/"+ctp.Spec.ActiveBranch, "5")
```
crenshaw-dev: Is 5 the max history length? Should we go ahead and make this configurable?

internal/git/git.go
```diff
@@ -586,3 +575,82 @@ func (g *EnvironmentOperations) MergeWithOursStrategy(ctx context.Context, propo
 	logger.Info("Successfully merged branches with 'ours' strategy", "proposedBranch", proposedBranch, "activeBranch", activeBranch)
 	return nil
 }
+
+// GetRevListFirstParent retrieves the first parent commit SHAs for the given branch using git rev-list.
+func (g *EnvironmentOperations) GetRevListFirstParent(ctx context.Context, branch string, maxCount string) ([]string, error) {
```
crenshaw-dev: Should maxcount be an int?

internal/git/git.go
```diff
@@ -586,3 +575,82 @@ func (g *EnvironmentOperations) MergeWithOursStrategy(ctx context.Context, propo
...
+		logger.Error(err, "failed to run git interpret-trailers", "stderr", stderr)
+		return nil, fmt.Errorf("failed to run git interpret-trailers: %w", err)
+	}
+	stdout := stdoutBuf.String()
+
+	lines := strings.Split(strings.TrimSpace(stdout), "\n")
+	trailers := make(map[string]string, len(lines))
+	for _, line := range lines {
+		if strings.Contains(line, ":") {
+			parts := strings.SplitN(line, ":", 2)
```
crenshaw-dev: `strings.Cut` might make this code nicer

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +163,206 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+		}
+
+		activeTrailers, err := gitOperations.GetTrailers(ctx, sha)
+		if err != nil {
+			// Best effort at getting the trailers if we cannot, we just don't generate history this sha
+			logger.Error(err, "failed to get trailers for active SHA", "sha", sha)
+			continue
+		}
+
+		if activeTrailers["No-Op"] == "true" {
```
crenshaw-dev: We should make the trailer keys constants.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +163,189 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+		input += "\n"
+	}
+
+	lines := strings.Split(input, "\n")
+	filtered := make([]string, 0, len(lines))
+
+	for _, line := range lines {
+		shouldKeep := true
+		for _, rm := range toRemove {
+			if strings.Contains(line, rm) {
```
```suggestion
			if strings.HasPrefix(line, rm+": ") {
```

internal/controller/changetransferpolicy_controller.go
```diff
@@ -506,23 +697,46 @@ func (r *ChangeTransferPolicyReconciler) creatOrUpdatePullRequest(ctx context.Co
...
 		return nil
 	}
 
+	commitTrailers := trailers{}
+	commitTrailers[constants.TrailerPullRequestID] = pr.Status.ID
+	commitTrailers[constants.TrailerPullRequestSourceBranch] = pr.Spec.SourceBranch
+	commitTrailers[constants.TrailerPullRequestTargetBranch] = pr.Spec.TargetBranch
+	commitTrailers[constants.TrailerPullRequestCreationTime] = pr.Status.PRCreationTime.Format(time.RFC3339)
+	commitTrailers[constants.TrailerPullRequestUrl] = pr.Status.Url
+	pr.Spec.PromoteCommitMessage = fmt.Sprintf("%s\n\n%s\n\n%s", pr.Spec.Title, pr.Spec.Description, commitTrailers)
```
crenshaw-dev: Shouldn't this be set further down, after constructing all the trailers?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +163,189 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+			logger.V(4).Info("No PullRequest-CreationTime found in trailers for active SHA", "sha", sha)
+		}
+
+		activeKeys, proposedKeys := getCommitStatusKeysFromTrailers(ctx, activeTrailers)
+
+		for _, key := range activeKeys {
+			h.Active.CommitStatuses = append(h.Active.CommitStatuses, promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase{
+				Key:   key,
+				Phase: activeTrailers[fmt.Sprintf(constants.TrailerCommitStatusActivePrefix+"%s-Phase", key)],
+				Url:   activeTrailers[fmt.Sprintf(constants.TrailerCommitStatusActivePrefix+"%s-Url", key)],
```
crenshaw-dev: I'd just use string concatenation instead of Sprintf for these.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -519,10 +711,32 @@ func (r *ChangeTransferPolicyReconciler) creatOrUpdatePullRequest(ctx context.Co
...
 		return nil
 	}
 
+	commitTrailers := trailers{}
+	commitTrailers[constants.TrailerPullRequestID] = pr.Status.ID
+	commitTrailers[constants.TrailerPullRequestSourceBranch] = pr.Spec.SourceBranch
+	commitTrailers[constants.TrailerPullRequestTargetBranch] = pr.Spec.TargetBranch
+	commitTrailers[constants.TrailerPullRequestCreationTime] = pr.Status.PRCreationTime.Format(time.RFC3339)
+	commitTrailers[constants.TrailerPullRequestUrl] = pr.Status.Url
+	pr.Spec.Commit.Message = fmt.Sprintf("%s\n\n%s\n\n%s", pr.Spec.Title, pr.Spec.Description, commitTrailers)
```
crenshaw-dev: Does this line do anything? Looks like setting the message happens in the retry function.

internal/types/constants/trailers.go
```diff
@@ -0,0 +1,28 @@
...
+	// TrailerCommitStatusProposedPrefix is the prefix for trailers indicating proposed commit statuses.
+	TrailerCommitStatusProposedPrefix = "CommitStatus-Proposed-"
+	// TrailerShaHydratedActive is the trailer key used to store the SHA of the active hydrated commit.
+	TrailerShaHydratedActive = "Sha-Hydrated-Active"
+	// TrailerShaHydratedProposed is the trailer key used to store the SHA of the proposed hydrated commit.
+	TrailerShaHydratedProposed = "Sha-Hydrated-Proposed"
+	// TrailerShaDryActive is the trailer key used to store the SHA of the active dry commit.
+	TrailerShaDryActive = "Sha-Dry-Active"
+	// TrailerShaDryProposed is the trailer key used to store the SHA of the proposed dry commit.
+	TrailerShaDryProposed = "Sha-Dry-Proposed"
```
crenshaw-dev: Maybe `Sha-Active-Hydrated` and `Sha-Proposed-Hydrated` so the alphabetic sorting groups the active and the proposed shas?

internal/types/constants/trailers.go
```diff
@@ -0,0 +1,28 @@
+package constants
+
+const (
```
crenshaw-dev, line 4: I'd suggest sorting this list of constants alphabetically, since that's how they'll show up in commit messages.

crenshaw-dev (changes requested):

internal/controller/changetransferpolicy_controller.go
zachaller: Yea I was waiting to make it configurable idk, the one risk is it's a lot of data, was going to leave it at 5 and see if people complained, but we can make it configurable in this PR if you think it's worth it

internal/controller/changetransferpolicy_controller.go
zachaller: I think any failure of history should not block reconcilation or even retry because of being able to smash merge on PR's etc etc

internal/types/constants/trailers.go
zachaller: I think I actually prefer keeping the groupings on Dry and Hydrated vs Active and Proposed

internal/controller/changetransferpolicy_controller.go
zachaller: This does not work because of these:

constants.TrailerCommitStatusActivePrefix,
constants.TrailerCommitStatusProposedPrefix,

crenshaw-dev: Just `HasPrefix` without the `: ` then?

internal/controller/changetransferpolicy_controller.go
crenshaw-dev: Gotcha... I think we should probably still fail on things that don't look like "they manually smashed merge." Will re-review with that in mind.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+
+	return historyEntry, true, nil
+}
+
+// populateActiveMetadata populates the active metadata for a history entry
+func (r *ChangeTransferPolicyReconciler) populateActiveMetadata(ctx context.Context, h *promoterv1alpha1.History, sha string, gitOperations *git.EnvironmentOperations) {
+	logger := log.FromContext(ctx)
+
+	if activeHydrated, err := gitOperations.GetShaMetadataFromGit(ctx, sha); err != nil {
+		logger.Error(err, "failed to get hydrated commit metadata for active SHA", "activeSha", sha)
```
crenshaw-dev: All this metadata should exist in git even if the PR was manually merged. So we should throw an error if we fail to get it.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+	} else {
+		h.Active.Hydrated = activeHydrated
+		h.Active.Hydrated.Body = removeKnownTrailers(h.Active.Hydrated.Body)
+	}
+
+	if activeDry, err := gitOperations.GetShaMetadataFromFile(ctx, sha); err != nil {
+		logger.Error(err, "failed to get dry commit metadata for active SHA", "activeSha", sha)
```
crenshaw-dev: This file should exist even after a manual merge, so we should throw the error if we fail to get it.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+	}
+}
+
+// populateProposedMetadata populates the proposed metadata for a history entry
+func (r *ChangeTransferPolicyReconciler) populateProposedMetadata(ctx context.Context, h *promoterv1alpha1.History, activeTrailers map[string]string, gitOperations *git.EnvironmentOperations) {
+	logger := log.FromContext(ctx)
+
+	proposedHydratedSha := activeTrailers[constants.TrailerShaHydratedProposed]
+	if proposedHydratedSha == "" {
+		logger.V(4).Info("No Sha-Hydrated-Proposed trailer found")
```
crenshaw-dev: Should use the constant in the log line.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+		return
+	}
+
+	if meta, err := gitOperations.GetShaMetadataFromGit(ctx, proposedHydratedSha); err != nil {
+		logger.Error(err, "failed to get hydrated commit metadata for proposed SHA", "proposedSha", proposedHydratedSha)
```
crenshaw-dev: This should exist even after a manual merge (especially since it's pulling from the proposed side). We should throw the error if this fails.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+	} else {
+		h.Proposed.Hydrated = meta
+	}
+}
+
+// populatePullRequestMetadata populates the pull request metadata for a history entry
+func (r *ChangeTransferPolicyReconciler) populatePullRequestMetadata(ctx context.Context, h *promoterv1alpha1.History, activeTrailers map[string]string) {
```
crenshaw-dev, line 254: Should use constants for all the trailer names when logging in this function.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+
+	if pullRequestUrl := activeTrailers[constants.TrailerPullRequestUrl]; pullRequestUrl != "" {
+		h.PullRequest.Url = pullRequestUrl
+	} else {
+		logger.V(4).Info("No PullRequest-Url found in trailers")
+	}
+
+	if timeStr := activeTrailers[constants.TrailerPullRequestCreationTime]; timeStr != "" {
+		if creationTime, err := time.Parse(time.RFC3339, timeStr); err != nil {
+			logger.Error(err, "failed to parse PullRequest-CreationTime", "time", timeStr)
```
crenshaw-dev: I kinda lean towards just throwing this to make it clear that someone's incorrectly formatting their dates.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+	logger := log.FromContext(ctx)
+
+	if pullRequestID := activeTrailers[constants.TrailerPullRequestID]; pullRequestID != "" {
+		h.PullRequest.ID = pullRequestID
+	} else {
+		logger.V(4).Info("No PullRequest-ID found in trailers")
+	}
+
+	if pullRequestUrl := activeTrailers[constants.TrailerPullRequestUrl]; pullRequestUrl != "" {
+		h.PullRequest.Url = pullRequestUrl
```
crenshaw-dev: We should probably consider this value trusted, but as an extra layer of protection maybe we should validate that the URL has either an http or https schema.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+// populateCommitStatuses populates the commit statuses for a history entry
+func (r *ChangeTransferPolicyReconciler) populateCommitStatuses(ctx context.Context, h *promoterv1alpha1.History, activeTrailers map[string]string) {
+	activeKeys, proposedKeys := getCommitStatusKeysFromTrailers(ctx, activeTrailers)
+
+	h.Active.CommitStatuses = make([]promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase, 0, len(activeKeys))
+	for _, key := range activeKeys {
+		h.Active.CommitStatuses = append(h.Active.CommitStatuses, promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase{
+			Key:   key,
+			Phase: activeTrailers[fmt.Sprintf(constants.TrailerCommitStatusActivePrefix+"%s-Phase", key)],
+			Url:   activeTrailers[fmt.Sprintf(constants.TrailerCommitStatusActivePrefix+"%s-Url", key)],
```
crenshaw-dev: Concat instead of sprintf?

internal/types/constants/trailers.go
```diff
@@ -0,0 +1,29 @@
+package constants
+
+// Keep sorted alphabetically
+const (
+	// TrailerCommitStatusActivePrefix is the prefix for trailers indicating active commit statuses.
+	TrailerCommitStatusActivePrefix = "CommitStatus-Active-"
```
crenshaw-dev: Looks like GitHub's convention is to only capitalize the first letter, e.g. `Co-authored-by`. Should we follow that? Might also mean abandoning camelcase, e.g. `Pullrequest-sourcebranch`.

internal/controller/changetransferpolicy_controller.go
crenshaw-dev: Hard-coding at 5 is probably fine. We should just document it.

internal/types/constants/trailers.go
```diff
@@ -0,0 +1,29 @@
...
+	// TrailerPullRequestSourceBranch is the trailer key used to store the source branch of the pull request.
+	TrailerPullRequestSourceBranch = "PullRequest-SourceBranch"
+	// TrailerPullRequestTargetBranch is the trailer key used to store the target branch of the pull request.
+	TrailerPullRequestTargetBranch = "PullRequest-TargetBranch"
+	// TrailerPullRequestCreationTime is the trailer key used to store the creation time of the pull request.
+	TrailerPullRequestCreationTime = "PullRequest-CreationTime"
+	// TrailerPullRequestUrl is the trailer key used to store the URL of the pull request.
+	TrailerPullRequestUrl = "PullRequest-Url"
+	// TrailerNoOp is the trailer key use to store the No-Op marker.
+	TrailerNoOp = "No-Op"
```
crenshaw-dev: Is this meant to be set when we do a fast-forward merge? I don't see it being set anywhere yet, only read.

zachaller: will fix it, its hard coded still in a string I missed when converting

internal/types/constants/trailers.go
zachaller: I went back and forth I kind of like the way we have it now but idk, I think Keys will probably be lower case in genreal which makes it easier to find them but we could also switch the seperator maybe?

zachaller: We should probably follow git

internal/controller/changetransferpolicy_controller.go
zachaller: if we validate would we just not populate if it's not http/https since we can't return an error we could log and not populate?

crenshaw-dev: Yeah, I think log it and leave it empty

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+			key = strings.TrimPrefix(key, prefix)
+			if key == "" {
+				logger.Info("Skipping empty trailer key", "trailer", trailer)
+				continue
+			}
+			parts := strings.Split(key, "-")
+			if len(parts) < 2 {
+				logger.Info("Skipping trailer with unexpected format", "trailer", trailer)
+				continue
+			}
```
crenshaw-dev: Maybe just strings.Cut?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+			csKey := strings.Join(parts[:len(parts)-1], "-")
+			// Append if it does not exist in keys
+			if slices.Index(keys, csKey) == -1 {
```
crenshaw-dev: Can probably switch to slices.Contains

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+		constants.TrailerShaHydratedActive,
+		constants.TrailerShaHydratedProposed,
+		constants.TrailerShaDryActive,
+		constants.TrailerShaDryProposed,
+		constants.TrailerNoOp,
+	}
+
+	if !strings.HasSuffix(input, "\n") {
+		input += "\n"
+	}
```
crenshaw-dev: Why ensure the trailing newline?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+
+	input = strings.Join(filtered, "\n")
+	input = strings.TrimSpace(input)
+	input += "\n"
+
+	if input == "\n" {
+		// If the input is empty after removing known trailers, we return an empty string.
+		return ""
+	}
```
crenshaw-dev: How about moving this immediately after the filter loop and just check the length of the filtered slice?

internal/controller/changetransferpolicy_controller.go
crenshaw-dev: Ah nvm, I see you're dropping the `-url` or `-phase` suffix. Maybe just add a comment to explain that.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,227 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+}
+
+// getCommitStatusKeysFromTrailers extracts the commit status keys from the trailers in the given context.
+func getCommitStatusKeysFromTrailers(ctx context.Context, trailers map[string]string) (activeKeys []string, proposedKeys []string) {
+	logger := log.FromContext(ctx)
+
+	extractKeys := func(prefix string) []string {
+		keys := []string{}
+		for key, trailer := range trailers {
```
crenshaw-dev: Small thing, but I'd just loop over keys, we don't actually do anything with the trailers expect log them. And we should be logging the keys instead.

internal/controller/changetransferpolicy_controller.go
zachaller: Don't think I can the logic is different, I am. just trying to get the key from the trailer key which means since I only have one -url or -phase I strip the last split on - off and then join back to a string could maybe do two specific cuts with know prefixes such as -url or -phase I think whats current is cleaner.

crenshaw-dev: Agreed, think it just needs a comment to explain what it's doing.

internal/controller/changetransferpolicy_controller.go
zachaller: To not mess up other trailers

zachaller: wait that is wrong just keeps a newline at the end of the commit message can probably remove it, although I don't hate newlines at the end of commit messages not sure if git has a recomendation

api/v1alpha1/changetransferpolicy_types.go
```diff
@@ -125,10 +125,12 @@ type ChangeTransferPolicyStatus struct {
 	Proposed CommitBranchState `json:"proposed,omitempty"`
 	// Active is the state of the active branch.
 	Active CommitBranchState `json:"active,omitempty"`
-
 	// PullRequest is the state of the pull request that was created for this ChangeTransferPolicy.
 	PullRequest *PullRequestCommonStatus `json:"pullRequest,omitempty"`
 
+	// History defines the history of promoted changes done by the ChangeTransferPolicy.
+	History []History `json:"history,omitempty"`
```
crenshaw-dev:
```suggestion
	// History defines the history of promoted changes done by the ChangeTransferPolicy. You can think of
	// it as a list of PRs merged by GitOps Promoter. It will not include changes that were manually merged.
	// The history length is hard-coded to be at most 5 entries. This may change in the future.
	// History is constructed on a best-effort basis and should be used for informational purposes only.
	// History is in reverse chronological order (newest is first).
	History []History `json:"history,omitempty"`
```

api/v1alpha1/changetransferpolicy_types.go
```diff
@@ -137,6 +139,25 @@ type ChangeTransferPolicyStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 }
 
+// History defines the history of promoted changes done by the ChangeTransferPolicy.
+type History struct {
```
crenshaw-dev:
```suggestion
// History describes a particular change that was promoted by the ChangeTransferPolicy.
type History struct {
```

api/v1alpha1/changetransferpolicy_types.go
```diff
@@ -137,6 +139,25 @@ type ChangeTransferPolicyStatus struct {
...
+	// Proposed is the state of the proposed branch.
+	Proposed CommitBranchStateHistoryProposed `json:"proposed,omitempty"`
```
crenshaw-dev:
```suggestion
	// Proposed is the state of the proposed branch at the time the PR was merged.
	Proposed CommitBranchStateHistoryProposed `json:"proposed,omitempty"`
```

api/v1alpha1/changetransferpolicy_types.go
```diff
@@ -137,6 +139,25 @@ type ChangeTransferPolicyStatus struct {
...
+	// Active is the state of the active branch.
+	Active CommitBranchState `json:"active,omitempty"`
```
crenshaw-dev:
```suggestion
	// Active is the state of the active branch at the time the PR was merged.
	Active CommitBranchState `json:"active,omitempty"`
```

Is this accurate?

api/v1alpha1/changetransferpolicy_types.go
```diff
@@ -137,6 +139,25 @@ type ChangeTransferPolicyStatus struct {
...
+	PullRequest *PullRequestCommonStatus `json:"pullRequest,omitempty"`
+}
+
+// CommitBranchStateHistoryProposed is identical to CommitBranchState minus the Dry state. In the context of History, the Dry state is not relevant as
+// the proposed dry side at merge becomes the Active.
+type CommitBranchStateHistoryProposed struct {
+	// Hydrated is the hydrated state of the branch, which is the commit that is currently being worked on.
+	Hydrated CommitShaState `json:"hydrated,omitempty"`
+	// CommitStatuses is a list of commit statuses that are being monitored for this branch.
+	CommitStatuses []ChangeRequestPolicyCommitStatusPhase `json:"commitStatuses,omitempty"`
```
crenshaw-dev:
```suggestion
	// CommitStatuses is a list of commit statuses that were being monitored for this branch.
	// This contains the state frozen at the moment the PR was merged.
	CommitStatuses []ChangeRequestPolicyCommitStatusPhase `json:"commitStatuses,omitempty"`
```

api/v1alpha1/promotionstrategy_types.go
```diff
@@ -143,6 +143,9 @@ type EnvironmentStatus struct {
 	// LastHealthyDryShas is a list of dry commits that were observed to be healthy in the environment.
 	// +kubebuilder:validation:Optional
 	LastHealthyDryShas []HealthyDryShas `json:"lastHealthyDryShas"`
+
+	// History defines the history of promoted changes done by the PromotionStrategy for each environment.
+	History []History `json:"history,omitempty"`
```
crenshaw-dev:
```suggestion
	// History defines the history of promoted changes done by the PromotionStrategy for each environment.
	// You can think of it as a list of PRs merged by GitOps Promoter. It will not include changes that were 
	// manually merged. The history length is hard-coded to be at most 5 entries. This may change in the future.
	// History is constructed on a best-effort basis and should be used for informational purposes only.
	// History is in reverse chronological order (newest is first).
	History []History `json:"history,omitempty"`
```

api/v1alpha1/pullrequest_types.go
```diff
@@ -44,13 +44,20 @@ type PullRequestSpec struct {
...
 	// State of the merge request closed/merged/open
 	// +kubebuilder:default:=open
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Enum=closed;merged;open
 	State PullRequestState `json:"state"`
 }
 
+// CommitConfiguration defines the commit configuration for how we will merge/squash/etc the pull request.
+type CommitConfiguration struct {
+	Message string `json:"message"`
```
crenshaw-dev, line 59:
```suggestion
	// Message is the commit message that will be written for the commit that's made when merging the PR.
	Message string `json:"message"`
```

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,226 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+// calculateHistory this function calculates the history by getting the first parents on the active branch and using the trailers to reconstruct the history.
+// calculateHistory calculates the history by getting the first parents on the active branch and using the trailers to reconstruct the history.
+// This function is best effort and will log errors but continue processing if it encounters issues with individual commits. This is because history is stored in git
+// in order to get out of a bad state requires re-writing git history or pushing a bunch of no-op commits greater than the max history limit.
+func (r *ChangeTransferPolicyReconciler) calculateHistory(ctx context.Context, ctp *promoterv1alpha1.ChangeTransferPolicy, gitOperations *git.EnvironmentOperations) error {
+	logger := log.FromContext(ctx)
+
+	shaListActive, err := gitOperations.GetRevListFirstParent(ctx, "origin/"+ctp.Spec.ActiveBranch, 5)
+	if err != nil {
+		return fmt.Errorf("failed to get rev-list commit history for branch %q: %w", ctp.Spec.ActiveBranch, err)
```
crenshaw-dev: Should we triple-down on "history is best-effort" and just log this error?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,226 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+	historyEntry := promoterv1alpha1.History{
+		Proposed:    promoterv1alpha1.CommitBranchStateHistoryProposed{},
+		Active:      promoterv1alpha1.CommitBranchState{},
+		PullRequest: &promoterv1alpha1.PullRequestCommonStatus{},
+	}
+
+	r.populateActiveMetadata(ctx, &historyEntry, sha, gitOperations)
+	r.populateProposedMetadata(ctx, &historyEntry, activeTrailers, gitOperations)
+	r.populatePullRequestMetadata(ctx, &historyEntry, activeTrailers)
+	r.populateCommitStatuses(ctx, &historyEntry, activeTrailers)
```
crenshaw-dev, line 214: All the info-level logs in these should probably be Debug-level. Otherwise, a single manually-merged PR could pepper the logs with irrelevant messages indefinitely.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,226 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+// populateCommitStatuses populates the commit statuses for a history entry
+func (r *ChangeTransferPolicyReconciler) populateCommitStatuses(ctx context.Context, h *promoterv1alpha1.History, activeTrailers map[string]string) {
+	activeKeys, proposedKeys := getCommitStatusKeysFromTrailers(ctx, activeTrailers)
+
+	h.Active.CommitStatuses = make([]promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase, 0, len(activeKeys))
+	for _, key := range activeKeys {
+		h.Active.CommitStatuses = append(h.Active.CommitStatuses, promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase{
+			Key:   key,
+			Phase: activeTrailers[constants.TrailerCommitStatusActivePrefix+key+"-phase"],
+			Url:   activeTrailers[constants.TrailerCommitStatusActivePrefix+key+"-url"],
```
crenshaw-dev: Let's apply the same URL validation here as above.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -155,6 +162,226 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+		})
+	}
+
+	h.Proposed.CommitStatuses = make([]promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase, 0, len(proposedKeys))
+	for _, key := range proposedKeys {
+		h.Proposed.CommitStatuses = append(h.Proposed.CommitStatuses, promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase{
+			Key:   key,
+			Phase: activeTrailers[constants.TrailerCommitStatusProposedPrefix+key+"-phase"],
+			Url:   activeTrailers[constants.TrailerCommitStatusProposedPrefix+key+"-url"],
```
crenshaw-dev: Let's apply the same URL validation here as above.

crenshaw-dev (changes requested):

internal/controller/changetransferpolicy_controller.go
zachaller: Fair I went back and forth I did them as Error to but then they log a stack trace so went back to Info, they where V(4).Info for a while too. It's like how much noise should we make I guess probably do like you said V(4) with the idea it will become a condition or something soon, will switch it back to V(4)

internal/controller/changetransferpolicy_controller.go
zachaller: mixed on this I kind of like keeping it an error but I do see how that is kind of counter intuitive but I think for this case I want to keep it.

PR #451, opened 2025-08-06
Add ArgoCD extension integration for promotion status display

Shirly8:
### Summary
Status Panel that displays promotion status in the ArgoCD top bar and navigation to the Promotion Strategy Tab

### Changes
* Fixed and improved extension styling of the UI
* Used pullRequest field from CRD to determine whether environments are promoted
* Status Calculation (Pending/Promoted/Failed ~ replaced 'success')
* Modulated extension (PromotionStrategy tab, status panel)

### Notes
* ArgoCD API returning stale data causing slight delay
* Polling Mechanism every 4 seconds


ui/extension/statusPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+}
+
+interface Resource {
+  kind: string;
+  name: string;
+  namespace: string;
+  liveState?: string;
+}
+
+const findPromotionStrategy = (application: Application) => {
```
zachaller, line 29: We should handle the case of multiple promotion stratigies in an application via just displaying some error or message.

zachaller (approved):

PR #455, opened 2025-08-08
fix: remove status  schema validation

zachaller:
I noticed a bunch of failures in tests because when we call the HandleReconciliationResult function, we get an empty status.phase also if we set the phase to pending there it breaks a lot of tests.

Doing something like
```
	var cs promoterv1alpha1.CommitStatus
	defer func() {
		cs.Status.Phase = promoterv1alpha1.CommitPhasePending // This has schema validation, so it should never be empty not sure status gets defaults at least during tests.
		utils.HandleReconciliationResult(ctx, startTime, &cs, r.Client, r.Recorder, &err)
	}()
```

### DIFF
--- a/api/v1alpha1/commitstatus_types.go
+++ b/api/v1alpha1/commitstatus_types.go
@@ -59,9 +59,12 @@ type CommitStatusStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 
 	// Id is the unique identifier of the commit status, set by the SCM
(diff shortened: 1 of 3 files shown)


crenshaw-dev:
Maybe just add an empty string to the enum? I think i did that locally at one point, and it seemed to work.

crenshaw-dev:
Setting omitempty might also do the trick

PR #479, opened 2025-08-27
chore: tighten cs key validation

crenshaw-dev:


api/v1alpha1/promotionstrategy_types.go
```diff
@@ -102,7 +102,9 @@ func (e *Environment) GetAutoMerge() bool {
 // CommitStatusSelector is used to select commit statuses by their key.
 type CommitStatusSelector struct {
 	// +required
-	// +kubebuilder:validation:Pattern:=(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?
```
crenshaw-dev, line 105: Removed the `?` wrapping the rest of the pattern, I don't think it's needed since this field is required.

PR #516, opened 2025-09-11
feat: inherit Ready conditions from child resources

crenshaw-dev:


docs/crd-specs.md
```diff
@@ -92,3 +92,35 @@ All fields are required, but defaults are provided in the installation manifests
...
+The `ChangeTransferPolicy` CRD may also have the following condition reasons:
+
+* `PullRequestNotReady`
+
+#### `PromotionStrategy`
+
+The `PromotionStrategy` CRD may also have the following condition reasons:
+
+* `PreviousEnvironmentCommitStatusNotReady`
+* `ChangeTransferPolicyNotReady`
```
zachaller, line 126: This should probably be on the CTP

crenshaw-dev: PS is the parent of the CTP, so the reason the PS is not Ready is because the CTP is not ready.

ChangeTransferPolicyNotReady on a CTP would be circular.

PR #557, opened 2025-10-08
feat: Add AzureDevops ScmProvider

### LINKED ISSUE #303: Support Azure DevOps Repos
Add support for Azure DevOps Repos as the SCM system

eegseth:
Adds Azure DevOps support as an SCM provider, with both PAT and Azure Workload Identity authentication methods.

Fixes #303


internal/scms/azuredevops/token_manager_test.go
```diff
@@ -0,0 +1,146 @@
+package azuredevops
```
zachaller: Can we get these tests to used Ginkgo an example can be seen there that uses it as a simple test runner without setting up controllers etc: https://github.com/argoproj-labs/gitops-promoter/blob/main/internal/utils/utils_test.go

api/v1alpha1/common_types.go
```diff
@@ -36,6 +36,44 @@ type Forgejo struct {
 	Domain string `json:"domain"`
 }
 
+// AzureDevOps is an Azure DevOps SCM provider configuration. It is used to configure the Azure DevOps settings.
+type AzureDevOps struct {
+	// Organization is the Azure DevOps organization name.
+	// +kubebuilder:validation:Required
+	Organization string `json:"organization"`
```
crenshaw-dev, line 46: Based on [Azure docs](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/naming-restrictions?view=azure-devops#organization-names), I think we can use this regex validation:

```regex
^[A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?$
```

We should have separate validation for a minimum length of 1 and a max length of 50.

api/v1alpha1/common_types.go
```diff
@@ -91,6 +129,16 @@ type ForgejoRepo struct {
...
 }
 
+// AzureDevOpsRepo is a repository in Azure DevOps, identified by its project and name.
+type AzureDevOpsRepo struct {
+	// Project is the project name in Azure DevOps.
+	// +kubebuilder:validation:Required
+	Project string `json:"project"`
+	// Name is the name of the repository.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
crenshaw-dev, line 148: Repo names have [a lot of rules](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/naming-restrictions?view=azure-devops#azure-repos-git) that are hard to regex. Maybe just add validation for the 64 character max? It should also have a 1-character minimum.

internal/scms/azuredevops/commit_status.go
```diff
@@ -0,0 +1,139 @@
...
+// CommitStatus implements the scms.CommitStatusProvider interface for Azure DevOps.
+type CommitStatus struct {
+	client    *azuredevops.Connection
+	k8sClient client.Client
+}
+
+var _ scms.CommitStatusProvider = &CommitStatus{}
+
+// NewAzdoCommitStatusProvider creates a new instance of CommitStatus for Azure DevOps.
+func NewAzdoCommitStatusProvider(ctx context.Context, k8sClient client.Client, scmProvider promoterv1alpha1.GenericScmProvider, secret v1.Secret, org string) (*CommitStatus, error) {
```
crenshaw-dev: I'd just go verbose and call it `NewAzureDevopsCommitStatusProvider`. It's painfully long, but I don't think the characters are worth breaking from convention.

internal/scms/azuredevops/git_operations.go
```diff
@@ -0,0 +1,279 @@
...
+	// AuthTypePAT indicates Personal Access Token authentication
+	AuthTypePAT AuthType = iota
+	// AuthTypeWorkloadIdentity indicates Azure workload identity authentication
+	AuthTypeWorkloadIdentity
+)
+
+// NewAzdoGitAuthenticationProvider creates a new instance of GitAuthenticationProvider for Azure DevOps.
+// It supports both Personal Access Token (PAT) and Azure Workload Identity authentication methods.
+// When workloadIdentity is configured, it will be used as the primary authentication method with PAT as fallback.
+func NewAzdoGitAuthenticationProvider(ctx context.Context, k8sClient client.Client, scmProvider v1alpha1.GenericScmProvider, secret *v1.Secret, repoRef client.ObjectKey) GitAuthenticationProvider {
```
crenshaw-dev, line 52: I think all the if conditions in this function can be inverted to return early and avoid deep nesting

api/v1alpha1/common_types.go
```diff
@@ -36,6 +36,47 @@ type Forgejo struct {
...
+	// ClientID is the Azure AD application (client) ID.
+	// If not specified, the AZURE_CLIENT_ID environment variable will be used.
+	// +kubebuilder:validation:Pattern=`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`
+	ClientID string `json:"clientId,omitempty"`
+	// ServiceAccountTokenPath is the path to the service account token file.
+	// This is typically mounted by the workload identity webhook.
+	// If not specified, the AZURE_FEDERATED_TOKEN_FILE environment variable will be used,
+	// or it will default to "/var/run/secrets/azure/tokens/azure-identity-token".
+	// +kubebuilder:default="/var/run/secrets/azure/tokens/azure-identity-token"
+	ServiceAccountTokenPath string `json:"serviceAccountTokenPath,omitempty"`
```
crenshaw-dev, line 79: What's the use case for using env vars instead of explicit field values? Is there some Azure automation that populates these env vars?

internal/scms/azuredevops/git_operations.go
```diff
@@ -0,0 +1,280 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/argoproj-labs/gitops-promoter/api/v1alpha1"
+)
+
+const (
+	// azureDevOpsTokenSecretKey is the key in the secret that contains the PAT token for Azure DevOps.
+	azureDevOpsTokenSecretKey = "token"
+	// azureDevOpsScope is the OAuth2 scope required for Azure DevOps API access
+	azureDevOpsScope = "499b84ac-1321-427f-aa17-267ca6975798/.default"
```
crenshaw-dev, line 25: Probably worth linking to some docs about this since it's a bit magic-stringy.

internal/scms/azuredevops/git_operations.go
```diff
@@ -0,0 +1,280 @@
...
+			authType:    AuthTypePAT,
+		}
+	}
+
+	logger.Info("Configuring Azure DevOps authentication with workload identity")
+
+	// Create workload identity credential
+	credential, err := createWorkloadIdentityCredential(workloadIdentity)
+	if err != nil {
+		logger.Error(err, "Failed to create workload identity credential, falling back to PAT authentication")
```
crenshaw-dev: Maybe just bail? If the user has opted into workload identity auth, I think it's pretty unlikely that falling back to a PAT makes sense.

internal/scms/azuredevops/git_operations.go
```diff
@@ -0,0 +1,280 @@
...
+// createWorkloadIdentityConnection creates an Azure DevOps connection using workload identity
+func createWorkloadIdentityConnection(ctx context.Context, scmProvider v1alpha1.GenericScmProvider, secret v1.Secret, organizationUrl string) (*azuredevops.Connection, GitAuthenticationProvider, error) {
+	// Create workload identity credential
+	credential, err := createWorkloadIdentityCredential(scmProvider.GetSpec().AzureDevOps.WorkloadIdentity)
+	if err != nil {
+		// Fall back to PAT authentication
+		token := string(secret.Data[azureDevOpsTokenSecretKey])
+		if token == "" {
+			return nil, GitAuthenticationProvider{}, fmt.Errorf("both workload identity and PAT authentication failed: workload identity error: %w, PAT token not found in secret", err)
+		}
```
crenshaw-dev: Ditto, if they've opted into workload identity, I wouldn't fall back to PAT.

api/v1alpha1/common_types.go
eegseth: Yeah, the Azure workload identity [Mutating Admission Webhook](https://azure.github.io/azure-workload-identity/docs/installation/mutating-admission-webhook.html#mutating-admission-webhook) injects these variables. I added a clarifying comment with link to the workload identity mutating admission webhook docs

internal/scms/azuredevops/git_operations.go
eegseth: Yep, added a link for it. The documentation around this scope guid is a bit sparse, but at least there is a link now

api/v1alpha1/common_types.go
crenshaw-dev: Is the use case for these fields to allow the user to have multiple tenant/client/token combos on a single GitOps Promoter instance? If so, we should probably document that.

And if the user is explicitly setting all this, are we really using "workload identity," or are we just auth'ing via a similar mechanism?

eegseth: No, the use case is just to have one tenant/client combo per GitOps Promoter instance.

The fields are optional because the Azure workload identity mutating admission webhook automatically injects these as environment variables when the deployment and service account are properly annotated/labeled.

The explicit field values are provided as an override mechanism for cases where:

* The cluster doesn't have the mutating admission webhook installed
* The user wants to override the injected values for some reason
* Testing or development scenarios where you want explicit control
*
And yes, we are still using "workload identity" even when setting explicit values - workload identity uses federated tokens bound to Kubernetes service accounts, regardless of whether the configuration comes from environment variables or explicit fields. The underlying authentication flow with Azure AD/Entra ID is the same.

crenshaw-dev:
Apologies for the delayed review. There was something bothering me that I couldn't quite figure out, and I think I finally made sense of it. 

All of the existing providers is accessible via one of two modes:

1) ClusterScmProvider: created by an admin along with creds, accessible to tenants in any namespace
2) ScmProvider: created by tenants in their own namespaces who must also provide their own credentials

This new provider breaks that binary. If I as a namespaced tenant can create an ScmProvider, I can take advantage of the creds (the workload identity) of the controller in the admin-controlled namespace. 

I think that, for this PR, we should make Azure DevOps workload identity available only in ClusterScmProviders. In a follow up PR we can add workload identity support to ScmProviders in a way that requires the tenant to provide their own credentials.

crenshaw-dev:
Closing in favor of https://github.com/argoproj-labs/gitops-promoter/pull/811. Thanks for getting the work started @eegseth!

PR #588, opened 2025-10-15
chore: improve missing branch error (#587)

crenshaw-dev:
Besides improving the error message from "fatal: 'origin/environments/qal-usw2-eks-next' is not a commit and a branch 'environments/qal-usw2-eks-next' cannot be created from it" to "fatal: couldn't find remote ref", this also simplifies the git operations to avoid maintaining a local version of the branch.

Fixes #587

PR #589, opened 2025-10-15
chore: use stateless git ops when possible

crenshaw-dev:
Changes that affect the local tree aren't threadsafe. Prefer using git in a way that doesn't affect the local tree.


zachaller:
another option: https://github.com/argoproj-labs/gitops-promoter/pull/590

crenshaw-dev:
@zachaller want me to just close this one? I think there's probably useful stuff to borrow here, but you know this code better

zachaller:
No this one has a bit better stuff in it those tests are bad though

zachaller:
We should merge this first though: https://github.com/argoproj-labs/gitops-promoter/pull/588 it has the same ish change in it

zachaller:
close for https://github.com/argoproj-labs/gitops-promoter/pull/593

PR #597, opened 2025-10-20
fix: don't update CS when there is no pending change

zachaller:
This prevents us from updating old commit status sha's on already closed PR's. This can happen when we just merge staging, and a new commit comes in updating dev which merges right aways (lowest env) and so we go update staging before staging get's a new PR.

# Add regression test for previous-environment commit status bug

## Summary

Adds a comprehensive test to validate the fix at `promotionstrategy_controller.go:341-350` that prevents updating commit statuses on already-merged PRs.

### The Bug (Fixed in previous commit)

Before the fix at line 341 was added, the controller would continuously update the previous-environment commit status to "pending" on already-merged commits when the previous environment's checks became non-passing. This created confusing state in the SCM where merged commits showed pending status checks.

**Production logs** (`cs-stats.txt`) show the bug occurring repeatedly:
```
11:45:18 - Setting phase="pending" for staging (proposedDrySha matches previousEnv active)
11:45:20 - Setting phase="pending" for staging (proposedDrySha matches previousEnv active)
11:45:21 - Setting phase="pending" for staging (proposedDrySha matches previousEnv active)
... (continued for minutes)
```

### The Fix

```go
// Line 341-350: Skip updating commit status when there's no open PR
if ctp.Status.Active.Dry.Sha == ctp.Status.Proposed.Dry.Sha {
    logger.V(4).Info("Skipping previous environment commit status update - no proposed change in current environment", ...)
    continue
}
```

This check prevents updates when `active == proposed`, meaning the PR has already been merged.

### New Test: "should not update commit status to pending after PR is already merged"

**Location**: `internal/controller/promotionstrategy_controller_test.go:2822-3050`

**Scenario**:
1. Create PromotionStrategy with dev → staging environments (both have active commit status checks)
2. Commit flows through and merges in both environments
3. Staging's previous-environment commit status is created and set to "success"
4. Staging PR merges → `active == proposed` (no open PR)
5. Dev's commit status is changed to "PENDING" (simulating failing checks)
6. Trigger PromotionStrategy reconciliation 5 times
7. **Verify**: Staging's commit status stays at "success" with original SHA (not updated to "pending")

### Validations

The test ensures:
- ✅ **Phase preservation**: Commit status phase stays at "success" (not reverted to "pending")
- ✅ **SHA preservation**: Commit status SHA stays pointing to the original merged commit
- ✅ **Skip logic activation**: Verified the skip logic at line 341 prevents updates
- ✅ **CTP state**: Verified staging maintains `active == proposed` throughout

### Test Quality

**Non-flaky design**:
- ✅ Zero `time.Sleep` calls (all timing handled via Eventually/Consistently)
- ✅ Only checks semantic values (phase, SHA) not K8s implementation details
- ✅ Uses Consistently for "should not change" assertions
- ✅ Deterministic behavior

**Performance**:
- Runtime: ~41 seconds
- Stability: 5/5 consecutive passes
- No arbitrary delays

### Without This Test
- Someone could accidentally remove/break the skip logic
- The production bug could be reintroduced
- No automated detection of regression

### With This Test
- ✅ Validates the fix works correctly
- ✅ Prevents regression (test fails if skip logic is broken)
- ✅ Documents expected behavior
- ✅ Runs fast and stable in CI

## Test Validation

We validated the test catches the bug by temporarily removing the fix:

```bash
# Commented out lines 341-350 (the skip logic)
$ go test -ginkgo.focus="should not update commit status to pending after PR is already merged"

Result: FAIL ✅ (correctly detected the bug)
Error: "Expected phase='success', Got phase='pending'"
```

## Related Issues

This test prevents the bug described in:
- Production logs: `cs-stats.txt` (showing repeated pending status on merged commits)
- Timeline: Bug occurred before fix at line 341 was added
- Symptom: Commit statuses on merged PRs being updated to "pending" repeatedly

## Testing Done

- ✅ Test passes with fix present (5/5 consecutive runs)
- ✅ Test fails with fix removed (validates bug detection)
- ✅ No flakiness observed
- ✅ Runs in ~41 seconds
- ✅ All assertions use Eventually/Consistently (no sleeps)

## Checklist

- [x] Test added
- [x] Test validates the fix works
- [x] Test catches regression when fix is removed
- [x] No `time.Sleep` calls (non-flaky)
- [x] Test runs in reasonable time (~41s)
- [x] Documented test scenario and validations


internal/controller/promotionstrategy_controller_test.go
```diff
@@ -2868,3 +2868,213 @@ func argocdApplications(namespace string, name string) (argocd.Application, argo
...
+					}
+					promotionStrategy.Annotations[promoterv1alpha1.ReconcileAtAnnotation] = metav1.Now().Format(time.RFC3339)
+					err = k8sClient.Patch(ctx, promotionStrategy, client.MergeFrom(orig))
+					g.Expect(err).To(Succeed())
+				}, constants.EventuallyTimeout).Should(Succeed())
+			}
+
+			By("Verifying commit status was NOT updated after staging PR merged")
+			// The bug would be: commit status gets updated to "pending" even though staging's active == proposed
+			// The fix at line 341 prevents this by skipping the update when active == proposed
```
crenshaw-dev: "The fix at line 341" maybe drop that part

PR #599, opened 2025-10-20
Remove panel extension

jwinters01:


ui/extension/index.tsx
```diff
@@ -64,8 +64,8 @@ window.extensionsAPI?.registerResourceExtension(
...
-window.extensionsAPI?.registerStatusPanelExtension(
-  StatusPanelComponent,
-  'Promotion Strategy',
-  'promotion_strategy_status'
-);
+// window.extensionsAPI?.registerStatusPanelExtension(
+//   StatusPanelComponent,
+//   'Promotion Strategy',
+//   'promotion_strategy_status'
+// );
```
jwinters01: This is the obvious quick fix. I am curious though if we would like to have this configurable via `EXTENSION_JS_VARS`? Do we intend to bring this back any time? My initial thought is no.

crenshaw-dev: I think "no," I'd just remove it completely.

crenshaw-dev (approved):

Issue #598, opened 2025-10-20
pull request: failed to convert PR number to int


daudcanugerah:
Version: v0.14.0

The issue occurs when I close the pull request and create a new change.
I believe it's related to issue #521 

```│ manager 2025-10-20T07:02:27.126380218Z    DEBUG    events    PullRequest "ivosights-r10-k8s-environment-twitter-worker-production-next-environment-twitter-worker-production-96d34c9e" is not Ready because "Reconciliati │
│ onError": Reconciliation failed: failed to close pull request: failed to close pull request: failed to convert PR number to int: strconv.Atoi: parsing "": invalid syntax    {"type": "Warning", "object": {"kind":"Chang │
│ eTransferPolicy","namespace":"argocd","name":"twitter-worker-environment-twitter-worker-production-4323530","uid":"c0586d3d-bbb8-46a9-b07a-779760cf9ceb","apiVersion":"promoter.argoproj.io/v1alpha1","resourceVersion":" │
│ 9359636235"}, "reason": "PullRequestNotReady"}                                                                                                                                                            ```

daudcanugerah:
I think this issue occurs because I enabled the “Automatically delete head branches” option after a PR is merged.

crenshaw-dev:
> when I close the pull request and create a new change

So you're manually hitting the "close PR" button in the gihub UI?

> I enabled the “Automatically delete head branches” option

This would cause other problems, because it would remove the branch that your hydrator is pushing to.

crenshaw-dev:
Give 0.15.0 a try. Even if the bug isn't fixed, the error message should be better.

PR #600, opened 2025-10-20
chore: better communicate missing ID errors

crenshaw-dev:
Today the PR controller will [produce opaque errors](https://github.com/argoproj-labs/gitops-promoter/issues/598) like this when the PR is set to be closed or merged but does not have an ID in the status:

> failed to convert PR number to int: strconv.Atoi: parsing "": invalid syntax

This change is meant to guarantee that we catch the problem earlier and that we communicate it more clearly.

1. If the CTP controller is about to merge a PR with a missing ID, it will preemptively fail reconciliation with a clear error message.
2. If any controller (or user) attempts to merge or close a PR that doesn't have an ID, the k8s API will throw a validation error.

Unrelated cleanup:

* Fix potential deadlocks in pullrequest fake provider.

### DIFF
--- a/internal/controller/changetransferpolicy_controller.go
+++ b/internal/controller/changetransferpolicy_controller.go
@@ -894,42 +894,45 @@ func (r *ChangeTransferPolicyReconciler) mergePullRequests(ctx context.Context, 
 			"activeDryCommitTime", ctp.Status.Active.Dry.CommitTime)
 	}
 
-	if pullRequest.Spec.State == promoterv1alpha1.PullRequestOpen && pullRequest.Status.State == promoterv1alpha1.PullRequestOpen {
-		if pullRequest.Status.ID == "" {
-			// We could rely on XValidation to catch the missing ID when setting the PR to merged, but this gives a
-			// better error message.
-
-			// If the PR has a ready condition with status false, get that reason/message for this error message.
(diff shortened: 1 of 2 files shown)


internal/controller/changetransferpolicy_controller.go
```diff
@@ -894,29 +894,45 @@ func (r *ChangeTransferPolicyReconciler) mergePullRequests(ctx context.Context,
...
-			}
-			pr.Spec.State = promoterv1alpha1.PullRequestMerged
-			return r.Update(ctx, &pr)
-		})
-		if err != nil {
-			return &pullRequest, fmt.Errorf("failed to update PR %q: %w", pullRequest.Name, err)
-		}
-		r.Recorder.Event(ctp, "Normal", constants.PullRequestMergedReason, fmt.Sprintf(constants.PullRequestMergedMessage, pullRequest.Name))
-		logger.Info("Merged pull request")
+	if pullRequest.Status.State != promoterv1alpha1.PullRequestOpen {
```
crenshaw-dev, line 897: Apologies for the refactor here. The linter was unhappy about a new nested if, and I basically agree with it. I think this is easier to read, but maybe a bit harder to review.

zachaller (approved):

Issue #635, opened 2025-10-22
ArgoCDCommitStatus controller tries to list Argo CD Applicaitons even when watchLocalApplications = false


crenshaw-dev:
```yaml
apiVersion: promoter.argoproj.io/v1alpha1
kind: ControllerConfiguration
spec:
  argocdCommitStatus:
    watchLocalApplications: false
```

```yaml
apiVersion: promoter.argoproj.io/v1alpha1
kind: ArgoCDCommitStatus
status:
  conditions:
    - lastTransitionTime: '2025-10-22T21:03:11Z'
      message: >-
        Reconciliation failed: failed to list ArgoCDApplications: no matches for
        kind "Application" in version "argoproj.io/v1alpha1"
      observedGeneration: 1
      reason: ReconciliationError
      status: 'False'
      type: Ready
```

PR #636, opened 2025-10-22
fix(acdcs): don't list Applications on local cluster when disabled

crenshaw-dev:
Fixes https://github.com/argoproj-labs/gitops-promoter/issues/635


crenshaw-dev:
Draft because I need to remember to revert the test changes

Issue #640, opened 2025-10-23
Refuse to merge a PR if the source branch is set to be auto-deleted


crenshaw-dev:
GitHub allows repo admins to configure branches to be auto-deleted after a PR is merged.

Since the Promoter basically assumes a persistent "proposed" branch, we should refuse to merge the PR if the SCM API informs us that the source branch would be deleted.

I know GitHub provides this information, but I'm not sure about other SCMs.

seankhliao:
We're working around this by having the repo set to auto delete, but using branch rulesets to prevent the promoter branches from getting deleted.
Not sure how that's surfaced in the api.

crenshaw-dev:
@seankhliao we'll probably eventually want programmatic solutions to the things you're using repo settings and rulesets for. But for now would you mind writing up a docs page for GitHub support describing your configs and reasoning for them?

PR #643, opened 2025-10-23
fix(pr): prevent accidental branch deletion (#640)

crenshaw-dev:
Fixes: https://github.com/argoproj-labs/gitops-promoter/issues/640

TBD:

* How to test
* Whether GitHub returns the correct value when the repo is set to delete branches but a branch protection rule prevents it
* How GitLab behaves
* Whether Forgejo has a similar setting
* How to make sure the PR controller reconciles again if the user changes the setting


crenshaw-dev:
GitHub API doesn't seem to provide the value as documented. Even if it did, it might not take branch protection into default.

Maybe we just build this into the GitRepository controller instead.

PR #658, opened 2025-10-27
chore: use watch predicate for watching argocd apps

zachaller:


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -535,10 +537,15 @@ func (r *ArgoCDCommitStatusReconciler) SetupWithManager(ctx context.Context, mcM
 			mcbuilder.WithEngageWithProviderClusters(false),
 			mcbuilder.WithPredicates(predicate.GenerationChangedPredicate{}),
 		).
-		WithOptions(controller.Options{MaxConcurrentReconciles: maxConcurrentReconciles, RateLimiter: rateLimiter}).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: maxConcurrentReconciles,
+			RateLimiter:             rateLimiter,
+			UsePriorityQueue:        ptr.To(true),
```
crenshaw-dev, line 543: Why enable priority queue?

zachaller: It's splits out re-syncs and non update changes to lower priorities and is actually the new default in the un-released version of controller-runtime

crenshaw-dev: Should we apply it to all the controllers while we're at it?

zachaller: I was just going to let those get defaulted since we didn't run into issue today with perf

PR #663, opened 2025-10-28
chore: add webhook delivery id logs

zachaller:


internal/webhookreceiver/server.go
```diff
@@ -116,8 +116,15 @@ func (wr *WebhookReceiver) postRoot(w http.ResponseWriter, r *http.Request) {
 
 	// Determine provider from headers
 	provider := wr.DetectProvider(r)
+
+	// Extract and log a single delivery ID from common webhook headers (GitHub, GitLab, Forgejo/Gitea).
+	deliveryID := wr.extractDeliveryID(r)
+	if deliveryID != "" {
+		logger.Info("webhook delivery id detected", "deliveryID", deliveryID)
+	}
+
```
crenshaw-dev, line 123: Maybe just set up a log context here to carry the field instead of explicitly adding it to calls below?

zachaller: done

PR #671, opened 2025-10-31
feat(ui): create App View Extension

jwinters01:
* Creates App View Extension
* Removes Resource View Extension
* Creates `build-extension-local` make target that places the created bundle in `/tmp/extensions/promoter` where a locally running argo-cd instance would expect it
app view extension

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,12 +133,13 @@ build-extension: ## Build ArgoCD extension.
 	cd ui/components-lib && npm install
 	cd ui/extension && npm install && npm run build
 
-.PHONY: build-extension-local
-build-extension-local: ## Build ArgoCD extension and output to /tmp/extensions/promoter directory.
-	cd ui/components-lib && npm install
-	cd ui/extension && npm install && npm run build
+.PHONY: install-extension-local
+install-extension-local: ## Install ArgoCD extension to /tmp/extensions/promoter directory.
 	mkdir -p /tmp/extensions/promoter
 	cp ui/extension/dist/extension-promoter.js /tmp/extensions/promoter/
+
+.PHONY: build-extension-local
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -133,6 +133,13 @@ build-extension: ## Build ArgoCD extension.
 	cd ui/components-lib && npm install
 	cd ui/extension && npm install && npm run build
 
+.PHONY: build-extension-local
+build-extension-local: ## Build ArgoCD extension and output to /tmp/extensions/promoter directory.
+	cd ui/components-lib && npm install
+	cd ui/extension && npm install && npm run build
+	mkdir -p /tmp/extensions/promoter
+	cp ui/extension/dist/extension-promoter.js /tmp/extensions/promoter/
```
zachaller, line 139: Maybe separate build-extension-local and install-extension-local?

jwinters01: are you suggesting that `build-extension-local` shouldn't also "install" the bundle in `/tmp/extensions`? I wanted a single script to bundle and place the extension in the right spot, would you be against `build-extension-local` also calling the separate `install-extension-local` after the build steps? (keeping in mind that `build-extension` already exists as a standard build target)

jwinters01: take a look at my recent commit

zachaller: Yea your change is good, I just wanted a way to build without installing.

crenshaw-dev (approved):

zachaller (approved):

zachaller:
tested this and it works and looks good, it did not default to the view but can follow up to add that

zachaller:
Screenshot 2025-11-26 at 4 02 40 PM

PR #690, opened 2025-11-13
fix!: switch to checks API

zachaller:
https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28

Check Run API

In a check suite, GitHub limits the number of check runs with the same name to 1000. Once these check runs exceed 1000, GitHub will start to automatically delete older check runs.

vs

Commit Status API

Note: there is a limit of 1000 statuses per sha and context within a repository. Attempts to create more than 1000 statuses will result in a validation error.


docs/getting-started.md
```diff
@@ -28,11 +28,16 @@ During the creation the GitHub App, you will need to configure the following set
...
+| Action         | Permission     | Notes                                            |
+| -------------- | -------------- | ------------------------------------------------ |
+| `Checks`       | Read and write | Required for creating check runs on commits      |
+| `Contents`     | Read and write | Required for reading and writing repository code |
+| `Pull requests`| Read and write | Required for creating and managing pull requests |
+
+> [!NOTE]
+> GitOps Promoter uses the GitHub [Checks API](https://docs.github.com/en/rest/checks/runs) to report commit status.
+> The older Commit Status API is no longer used. If you previously configured `Commit statuses` permissions, you can
+> remove them and replace with `Checks` permissions instead.
```
crenshaw-dev: I'd put this in release notes instead of docs. It's too early to have migration info in docs. :-P

internal/scms/github/commit_status.go
```diff
@@ -17,62 +17,211 @@ import (
...
 		return nil, fmt.Errorf("failed to get GitRepository: %w", err)
 	}
 
+	return gitRepo, nil
+}
+
+// shouldUpdateCheckRun determines if we should update an existing check run or create a new one
+func (cs *CommitStatus) shouldUpdateCheckRun(commitStatus *promoterv1alpha1.CommitStatus) bool {
+	return commitStatus.Status.Sha == commitStatus.Spec.Sha && commitStatus.Status.Id != ""
+}
```
crenshaw-dev: Maybe just inline this?

crenshaw-dev (approved):

PR #691, opened 2025-11-13
internal/scms/github: allow any merge method for Github PRs

### LINKED ISSUE #673: Auto merging Github PRs fail if "merge" method is disabled
In our repos, we typically only allow the "squash and merge" (`squash` in the api) method for merging PRs, to keep a clean linear history for the main branch.

In github, this setting applies to all PRs on the repo.

Currently the gitops-promoter hardcodes the use of the `merge` method, which will fail on repos where that's disallowed.
https://github.com/argoproj-labs/gitops-promoter/blob/2406e00848f0d8ab3fa3cb25acc5ecfc22b6265a/internal/scms/github/pullrequest.go#L171

As the field is optional with the default being `merge`, I think the gitops promoter can just leave it unset
https://pkg.go.dev/github.com/google/go-github/v76/github#PullRequestOptions

seankhliao:
The MergeMethod parameter is optional, and already defaults to `merge` if left unset.
Unsetting it should allow it to work with any other merge method if `merge` was disabled in settings.
https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#merge-a-pull-request

Fixes #673

### DIFF
--- a/internal/scms/github/pullrequest.go
+++ b/internal/scms/github/pullrequest.go
@@ -168,7 +168,6 @@ func (pr *PullRequest) Merge(ctx context.Context, pullRequest v1alpha1.PullReque
 		prNumber,
 		pullRequest.Spec.Commit.Message,
 		&github.PullRequestOptions{
-			MergeMethod:        "merge",
 			DontDefaultIfBlank: false,
 			SHA:                pullRequest.Spec.MergeSha,
 		})


zachaller:
I have a feeling this will break history calculations? Although once we move no-op support to the hydrator I think this makes sense. Have you tested this and validated history still behaves?

seankhliao:
Is that history in argocd or in the promoter?

seankhliao:
I think I should close this:

* it didn't work as I expected
* we can instead use branch rulesets to require linear history / squash merge for the main branch, instead of disabling all other merge types for the repo
* wrt to history, it didn't break in my testing, using our own ci to push to the -next branch, but it did break with argocd hydrator producing commits which don't merge cleanly.

zachaller:
Interesting good to know that might change quite a bit with https://github.com/argoproj-labs/gitops-promoter/pull/746 and https://github.com/argoproj/argo-cd/pull/25056

Issue #719, opened 2025-11-20
ChangeTransferPolicy controller fails to update status if hydrator.metadata repoURL is SSH


jjsiv:
If the ArgoCD Application is configured with SSH repoURL, the hydrator.metadata will subsequently look like this:
```
{
  "repoURL": "git@github.com:jjsiv-homelab/example-project-deploy.git",
  "drySha": "312aeae19a7530ffb7049cea249770e817751f08",
  "author": "Jedrzej Kotkowski <jedrzejk143@gmail.com>",
  "date": "2025-11-20T19:51:30Z",
  "subject": "staging & prod"
}
```

Then the CTP controller will fail to update the status on the ChangeTransferPolicy for the branch:
```
[...]

PR #723, opened 2025-11-21
fix: convert ref commit urls to https, https url from SCM

zachaller:
fixes: https://github.com/argoproj-labs/gitops-promoter/issues/719


zachaller:
We could also probably drop the conversion function (ConvertSSHToHTTPS) and just make the CI supply https urls. This change would then only be to pull the repo url from the SCM provide and not use the metafile, this is probably safe.

PR #726, opened 2025-11-22
fix: cleanup timed commit statuses

zachaller:
### Problem
When environment branches are renamed or removed from a TimedCommitStatus spec, the associated CommitStatus resources are not cleaned up, leaving orphaned resources.

### Solution
- Added `cleanupOrphanedCommitStatuses()` method to delete CommitStatus resources that are no longer referenced
- Follows the same pattern as the PromotionStrategy controller's orphaned CTP cleanup
- Added `TimedCommitStatusLabel` constant for consistency
- Added test coverage to verify cleanup on environment removal

### Testing
- New test: "should cleanup orphaned CommitStatus resources" ✅
- Verifies CommitStatus resources are deleted when environments are removed from spec

### DIFF
--- a/internal/controller/timedcommitstatus_controller_test.go
+++ b/internal/controller/timedcommitstatus_controller_test.go
@@ -668,21 +668,16 @@ var _ = Describe("TimedCommitStatus Controller", func() {
 	Context("When environment branch names are changed", func() {
 		ctx := context.Background()
 
-		It("should cleanup orphaned CommitStatus resources", func() {
-			By("Creating initial resources with standard branch naming")
+		var name string
+		var promotionStrategy *promoterv1alpha1.PromotionStrategy
+		var timedCommitStatus *promoterv1alpha1.TimedCommitStatus
+
+		BeforeEach(func() {
+			By("Creating the test resources")
 			var scmSecret *v1.Secret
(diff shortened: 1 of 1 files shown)


internal/controller/timedcommitstatus_controller.go
```diff
@@ -267,6 +274,64 @@ func (r *TimedCommitStatusReconciler) processEnvironments(ctx context.Context, t
 	return transitionedEnvironments, commitStatuses, nil
 }
 
+// cleanupOrphanedCommitStatuses deletes CommitStatus resources that are owned by this TimedCommitStatus
+// but are not in the current list of valid CommitStatus resources (i.e., they correspond to removed or renamed environments).
+//
+//nolint:dupl // Similar to PromotionStrategy cleanup but works with different types
+func (r *TimedCommitStatusReconciler) cleanupOrphanedCommitStatuses(ctx context.Context, tcs *promoterv1alpha1.TimedCommitStatus, validCommitStatuses []*promoterv1alpha1.CommitStatus) error {
```
zachaller, line 281: When we add a few more controllers it might make sense to move this to a generic shared func

zachaller: Will do this after when I add GitCommitStatus

PR #728, opened 2025-11-22
feat: Scheduled commit status controller

zachaller:


tenitski:
Can you please clarify the behaviour in the following scenarios:

1. There is number of commits outside of the maintenance window. Is it a single PR that gets created and updated with the latest hydrated state OR there is a PR per commit?

2. So there were changes outside of the window, but then NO changes inside the window. Will the promoter now automatically pick the pending PR(s) created outside the window and merge them? Or it only processes the PRs created inside the window?

PR #729, opened 2025-11-22
feat: Git commit status controller

zachaller:


api/v1alpha1/gitcommitstatus_types.go
```diff
@@ -0,0 +1,235 @@
...
+	//   - Environment.ProposedCommitStatuses (applies to specific environment)
+	// +required
+	PromotionStrategyRef ObjectReference `json:"promotionStrategyRef"`
+
+	// Key is the unique identifier for this validation rule.
+	// It is used as the commit status key and in status messages.
+	// This key is matched against PromotionStrategy's proposedCommitStatuses or activeCommitStatuses
+	// to determine which environments this validation applies to.
+	// +required
+	// +kubebuilder:validation:MaxLength=63
```
crenshaw-dev, line 42:
```suggestion
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=63
```

I think `+required` actually passes for an empty string.

api/v1alpha1/gitcommitstatus_types.go
```diff
@@ -0,0 +1,228 @@
...
+	// controls which commit's data is used in the expression evaluation.
+	//
+	// Examples:
+	//   validateCommit: "active" - "Don't promote if a revert commit is detected"
+	//   validateCommit: "proposed" - "Don't promote unless new commit follows naming convention"
+	//
+	// +optional
+	// +kubebuilder:default="active"
+	// +kubebuilder:validation:Enum=active;proposed
+	ValidateCommit string `json:"validateCommit,omitempty"`
```
crenshaw-dev: I notice the other was called `reportOn`, I'm not sure which I like better.

api/v1alpha1/gitcommitstatus_types.go
```diff
@@ -0,0 +1,228 @@
...
+	// Available variables in the expression context:
+	//   - Commit.SHA (string): the commit SHA being validated (active or proposed based on ValidateCommit)
+	//   - Commit.Subject (string): the first line of the commit message
+	//   - Commit.Body (string): the commit message body (everything after the subject line)
+	//   - Commit.Author (string): commit author email address
+	//   - Commit.Committer (string): committer email address
+	//   - Commit.Trailers (map[string][]string): git trailers parsed from commit message
+	//
+	// +required
+	Expression string `json:"expression"`
```
crenshaw-dev, line 98: Should we have it return a structure so that we can expand features in the future?

```yaml
expression: |
  {
    "passed": 1 == 2,
    "message": Commit.SHA + " is busted"
  }
```

Or maybe that would just end up being a new field:

```yaml
messageExpression: |-
  Commit.SHA + " is busted"
```

api/v1alpha1/gitcommitstatus_types.go
```diff
@@ -0,0 +1,228 @@
...
+	// +listMapKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// GitCommitStatusEnvironmentStatus defines the observed validation status for a specific environment.
+type GitCommitStatusEnvironmentStatus struct {
+	// Branch is the environment branch name being validated.
+	// +required
+	Branch string `json:"branch"`
```
crenshaw-dev, line 128: minlength 1

api/v1alpha1/gitcommitstatus_types.go
```diff
@@ -0,0 +1,228 @@
...
+	// May be empty if the PromotionStrategy hasn't reconciled yet.
+	// +optional
+	ActiveHydratedSha string `json:"activeHydratedSha,omitempty"`
+
+	// ValidatedSha is the commit SHA that was actually validated by the expression.
+	// This will match either ProposedHydratedSha or ActiveHydratedSha depending on
+	// the ValidateCommit setting ("proposed" or "active").
+	// This field clarifies which commit's data was used in the expression evaluation.
+	// +optional
+	ValidatedSha string `json:"validatedSha,omitempty"`
```
crenshaw-dev: Sha regex validation

api/v1alpha1/gitcommitstatus_types.go
zachaller: This is different, reportOn for the this controller is fixed on proposed

zachaller: ValidateCommit is the commit we are validating against or another way is what commit the expression runs against

api/v1alpha1/gitcommitstatus_types.go
zachaller: We don't do this on any other status sha's if we want to do this I would do it in a new pr where we do just that.

api/v1alpha1/gitcommitstatus_types.go
zachaller: So this one we are split a bit argocd commit status has it on it's Environment status TimedCommitStatus branch field does not have anything. We should probably update TimedCommitStatus as well. Being a status field in a follow up PR?

api/v1alpha1/gitcommitstatus_types.go
zachaller: Yea, I cant think of a reason to change the return to anything other than bool but I do think in a follow up PR it would be good to have description templateable.

api/v1alpha1/gitcommitstatus_types.go
crenshaw-dev: Yep!

api/v1alpha1/gitcommitstatus_types.go
zachaller: going back and forth on this again it might be nice to be able to use this in the description template aka pull data from the commit trailers etc and could maybe be how we do all built in controllers. My only complaint is that it validating that the expression has a passed or triggered or whatever the required key is.

I am also not sure how I feel about two fields one for the trigger/pass type logic and one for the template/status output

internal/controller/gitcommitstatus_controller_test.go
```diff
@@ -0,0 +1,794 @@
...
+		}
+
+		setupInitialTestGitRepoOnServer(ctx, name, name)
+
+		Expect(k8sClient.Create(ctx, scmSecret)).To(Succeed())
+		Expect(k8sClient.Create(ctx, scmProvider)).To(Succeed())
+		Expect(k8sClient.Create(ctx, gitRepo)).To(Succeed())
+		Expect(k8sClient.Create(ctx, promotionStrategy)).To(Succeed())
+
+		By("Waiting for PromotionStrategy to be reconciled with initial state")
```
crenshaw-dev: Do we need to do this? Shouldn't the gcs controller be able to gracefully handle waiting for the PS to be up and running?

internal/controller/gitcommitstatus_controller_test.go
```diff
@@ -0,0 +1,794 @@
+/*
```
crenshaw-dev, line 1: The tests are super thorough, but I think they all rely on directly inspecting the GCS status to see if the expected status fields are set. Should we have a PromotionStrategy test that tests one example of GCS use truly end-to-end?

crenshaw-dev: Looks like the tests only cover global commit statuses. The PS test could have a per-env config just so we get that code tested.

internal/controller/suite_test.go
```diff
@@ -402,6 +424,14 @@ var _ = BeforeSuite(func() {
 	}).SetupWithManager(ctx, multiClusterManager)
 	Expect(err).ToNot(HaveOccurred())
 
+	err = (&GitCommitStatusReconciler{
+		Client:      k8sManager.GetClient(),
+		Scheme:      k8sManager.GetScheme(),
+		Recorder:    k8sManager.GetEventRecorderFor("GitCommitStatus"),
+		SettingsMgr: settingsMgr,
```
crenshaw-dev, line 431: Should we be setting EnqueueCTP on these so tests behave more like real life?

internal/controller/gitcommitstatus_controller_test.go
zachaller: removed

crenshaw-dev (approved):

PR #740, opened 2025-11-27
feat: WebRequestCommitStatus controller

zachaller:


api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+	PromotionStrategyRef ObjectReference `json:"promotionStrategyRef"`
+
+	// Key is the unique identifier for this validation rule.
+	// It is used as the commit status key and in status messages.
+	// This key is matched against PromotionStrategy's proposedCommitStatuses or activeCommitStatuses
+	// to determine which environments this validation applies to.
+	// +required
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	Key string `json:"key"`
```
crenshaw-dev, line 46:
```suggestion
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=63
	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
	Key string `json:"key"`
```

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Description is a human-readable description of this validation that will be shown in the SCM provider
+	// (GitHub, GitLab, etc.) as the commit status description.
+	// If not specified, defaults to empty string.
+	// +optional
+	// +kubebuilder:validation:MaxLength=140
```
crenshaw-dev: Is there a specific reason for the 140 length?

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+	// Supports Go templates in URL, Headers, and Body fields.
+	// +required
+	HTTPRequest HTTPRequestSpec `json:"httpRequest"`
+
+	// Expression is evaluated using the expr library (github.com/expr-lang/expr) against the HTTP response.
+	// The expression must return a boolean value where true indicates the validation passed.
+	//
+	// Available variables in the expression context:
+	//   - Response.StatusCode (int): the HTTP response status code
+	//   - Response.Body (any): parsed JSON as map[string]any, or raw string if not JSON
```
crenshaw-dev, line 112: Maybe also support `[]any`?

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+	//
+	// +required
+	Expression string `json:"expression"`
+
+	// PollingInterval controls how often to retry the HTTP request while in pending state.
+	// When reportOn is "proposed": stops polling after success for a given SHA.
+	// When reportOn is "active": always polls at this interval.
+	// +optional
+	// +kubebuilder:default="1m"
+	PollingInterval metav1.Duration `json:"pollingInterval,omitempty"`
```
crenshaw-dev: Maybe

```yaml
polling:
  interval: 1m
```

In case we need future expansion?

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+type HTTPRequestSpec struct {
+	// URL is the HTTP endpoint to request.
+	// Supports Go templates with the following variables:
+	//   - {{ .ProposedHydratedSha }}: the proposed commit SHA
+	//   - {{ .ActiveHydratedSha }}: the active/deployed commit SHA
+	//   - {{ .Key }}: the WebRequestCommitStatus key
+	//   - {{ .Name }}: the WebRequestCommitStatus resource name
+	//   - {{ .Namespace }}: the WebRequestCommitStatus namespace
+	// +required
+	URL string `json:"url"`
```
crenshaw-dev: Maybe call it URLTemplate? We could also probably set minLength to 1.

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Method is the HTTP method to use.
+	// +required
+	// +kubebuilder:validation:Enum=GET;POST;PUT;PATCH
+	Method string `json:"method"`
```
crenshaw-dev, line 161: Also `DELETE`?

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Headers are additional HTTP headers to include in the request.
+	// Values support Go templates with the same variables as URL.
+	// +optional
+	Headers map[string]string `json:"headers,omitempty"`
```
crenshaw-dev: Should the keys support templating too?

Is there any reason we'd want to go more structured, e.g. 

```yaml
headers:
- name: dynamic
  keyTemplate: 'Something.Terrible'
  valueTemplate: 'Something.Even.Worse'
```

Maybe that's overengineering. 🤷

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+	Body string `json:"body,omitempty"`
+
+	// Timeout is the maximum time to wait for the HTTP request to complete.
+	// +optional
+	// +kubebuilder:default="30s"
+	Timeout metav1.Duration `json:"timeout,omitempty"`
+
+	// AuthSecretRef references a Kubernetes Secret containing authentication credentials.
+	// +optional
+	AuthSecretRef *AuthSecretRef `json:"authSecretRef,omitempty"`
```
crenshaw-dev: Maybe nest? auth.secretRef?

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// ReportedSha is the SHA where the CommitStatus was reported (based on reportOn field).
+	// +optional
+	ReportedSha string `json:"reportedSha,omitempty"`
+
+	// LastSuccessfulSha is the SHA that last achieved success.
+	// Used to detect SHA changes - when ReportedSha != LastSuccessfulSha, polling continues.
+	// When they match and phase is success, polling stops (for reportOn: proposed).
+	// +optional
+	LastSuccessfulSha string `json:"lastSuccessfulSha,omitempty"`
```
crenshaw-dev, line 469: Regex validation for these?

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,275 @@
...
+	// +required
+	Phase string `json:"phase"`
+
+	// LastRequestTime is when the HTTP request was last made.
+	// +optional
+	LastRequestTime *metav1.Time `json:"lastRequestTime,omitempty"`
+
+	// ResponseStatusCode is the HTTP status code from the last request.
+	// +optional
+	ResponseStatusCode *int `json:"responseStatusCode,omitempty"`
```
crenshaw-dev: response.statusCode?

internal/controller/webrequestcommitstatus_controller.go
```diff
@@ -0,0 +1,789 @@
...
+		return true
+	}
+	if checkKeyInSelectors(wrcs.Spec.Key, psEnv.ActiveCommitStatuses) {
+		return true
+	}
+	return false
+}
+
+// findPreviousEnvStatus finds the previous status for a given environment branch
+func findPreviousEnvStatus(previousStatus *promoterv1alpha1.WebRequestCommitStatusStatus, branch string) *promoterv1alpha1.WebRequestCommitStatusEnvironmentStatus {
```
crenshaw-dev: Simplify to getStatusByBranch? There's nothing "previous-specific" about the func.

internal/controller/webrequestcommitstatus_controller.go
```diff
@@ -0,0 +1,789 @@
...
+		if previousStatus.Environments[i].Environment == branch {
+			return &previousStatus.Environments[i]
+		}
+	}
+	return nil
+}
+
+// shouldSkipRequest checks if we should skip the HTTP request for this environment
+func shouldSkipRequest(reportOn string, previousEnvStatus *promoterv1alpha1.WebRequestCommitStatusEnvironmentStatus, reportedSha string) bool {
+	if reportOn != ReportOnProposed || previousEnvStatus == nil {
```
crenshaw-dev: My tired brain doesn't immediately see what this means, maybe add a comment?

crenshaw-dev (changes requested):

api/v1alpha1/webrequestcommitstatus_types.go
zachaller: It seems weird to call delete multiple times

api/v1alpha1/webrequestcommitstatus_types.go
zachaller: templating is available in more than just url, I updated docs to make it more clear url, body, headers

api/v1alpha1/webrequestcommitstatus_types.go
zachaller: I don't love that, I think the map[string]string is fine, only question is if we think we want to support multiple headers of the same key which would require map[string][]string but i don't love that for the request side. There are lots of ways to do multiple too such as Header: val1;val2; etc etc

internal/controller/webrequestcommitstatus_controller.go
zachaller: It is the previous reconcile state

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,571 @@
...
+	//   - Documentation or runbook
+	//
+	// Examples:
+	//   - "https://approvals.example.com/{{ .ReportedSha }}"
+	//   - "https://dashboard.example.com/{{ .Branch }}/status"
+	//   - "https://change-api.example.com/changes?asset={{ index .NamespaceMetadata.Labels \"asset-id\" }}"
+	//
+	// If not specified, defaults to empty string (no URL shown).
+	// +optional
+	// +kubebuilder:validation:Pattern="^(https?://.*)?$"
```
crenshaw-dev: Should we validate this at all? The whole field could be calculated, like

```
{{ if x -}}
https://blah
{{- else -}}
https://bleh
{{- end }}
```

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,571 @@
...
+	//   - Response.StatusCode == 200 && Response.Body.approved == true
+	//   - Response.StatusCode == 200 && Response.Body.status == "approved"
+	//   - len(Response.Headers["X-Approval"]) > 0
+	//
+	// +required
+	Expression string `json:"expression"`
+
+	// Polling controls polling behavior for the HTTP request.
+	// +optional
+	Polling PollingSpec `json:"polling,omitempty"`
```
crenshaw-dev, line 126: What do you think of following the client-go retry model? We don't have to support all the fields initially, but a config supporting equivalent features would be:

```go
wait.Backoff{
	// Steps:    4,
	Duration: 10 * time.Millisecond,
	// Factor:   5.0,
	// Jitter:   0.1,
}
```

api/v1alpha1/webrequestcommitstatus_types.go
crenshaw-dev: Gotcha. Even without templating, what about structuring? 

```yaml
headers:
- name:
  value:
```

Allows using a merge key and leaves room for future expansion.

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,571 @@
...
+	//         name: my-tls-cert
+	//
+	// +optional
+	Authentication *HttpAuthentication `json:"authentication,omitempty"`
+}
+
+// HttpAuthentication defines authentication options for HTTP requests.
+//
+// Only one authentication method should be specified. If multiple methods are provided,
+// the first one found in this order will be used: Basic, Bearer, OAuth2, TLS.
```
crenshaw-dev:
```suggestion
// Only one authentication method should be specified.
```

Not needed, since we have validation.

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,571 @@
...
+// HTTP Basic Auth encodes the username and password as base64 and sends them
+// in the Authorization header: "Authorization: Basic <base64(username:password)>"
+//
+// Credentials must be stored in a Kubernetes secret and referenced via secretRef:
+//
+//	basic:
+//	  secretRef:
+//	    name: my-basic-auth-secret
+//	    usernameKey: username  # optional, defaults to "username"
+//	    passwordKey: password  # optional, defaults to "password"
```
crenshaw-dev: Can we just hard-code this? Makes it more difficult to maliciously lift arbitrary credentials.

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -0,0 +1,571 @@
...
+// - JWT tokens
+// - Personal access tokens
+// - Static authentication tokens
+//
+// The token must be stored in a Kubernetes secret and referenced via secretRef:
+//
+//	bearer:
+//	  secretRef:
+//	    name: my-bearer-token-secret
+//	    key: token  # optional, defaults to "token"
```
crenshaw-dev: Ditto re: hardcoding. We've gone with hard-coded keys for SCM credentials, too.

docs/gating-promotions.md
```diff
@@ -166,6 +166,17 @@ Key features:
...
+
+The [WebRequestCommitStatus](commit-status-controllers/webrequest.md) controller enables integration with external systems for promotion decisions by making HTTP requests to external endpoints and evaluating responses.
+
+Key features:
+- Makes configurable HTTP requests to any HTTP-accessible endpoint
+- Evaluates responses using expressions (expr language)
+- Supports authentication (Bearer, Basic Auth, custom headers)
+- Configurable polling intervals
+- Can gate on proposed or active commits
+
```
crenshaw-dev, line 179: Should we have a note in the multi-tenancy doc about the security risks of granting a tenant access to create these resources? There's nothing stopping them from doing in-cluster network probing or hammering the webhook endpoint. We could suggest that the admin set up network policies on their controller pod to mitigate the issue.

crenshaw-dev (changes requested):

zachaller:
closing in favor of https://github.com/argoproj-labs/gitops-promoter/pull/936

PR #746, opened 2025-12-01
fix: no op removal

zachaller:
# Remove No-Op Commit Handling from Promoter

## Summary

This PR removes no-op commit handling from the promoter, delegating duplicate detection to the hydrator.

## Why

The hydrator is the right place to detect duplicates because **it's the system that knows the contents of what's being promoted**.

The promoter only sees commit SHAs and metadata - it doesn't know what's inside the manifests. The hydrator, on the other hand:
- Renders the manifests (helm template, kustomize build, etc.)
- Knows what's currently on the staging branch
- Can compare the two and decide whether a new commit is needed

By handling this at the hydrator level, we avoid creating unnecessary commits in the first place, rather than having the promoter detect and work around them after the fact.

## Changes

- Removed `IsPullRequestRequired()` and `PromoteEnvironmentWithMerge()` from git operations
- Removed `TrailerNoOp` constant and related handling
- Simplified PR creation logic
- Added `docs/custom-hydrator.md` documenting the hydrator contract and git notes for duplicate detection

## Depends on:

https://github.com/argoproj/argo-cd/pull/25056


internal/controller/commitstatus_controller.go
```diff
@@ -226,25 +226,10 @@ func (r *CommitStatusReconciler) triggerReconcileChangeTransferPolicy(ctx contex
...
-			ctpUpdated.Annotations[promoterv1alpha1.ReconcileAtAnnotation] = time.Now().Format(time.RFC3339Nano)
-			if err != nil {
-				return fmt.Errorf("failed to get ChangeTransferPolicy %q: %w", ctp.Name, err)
-			}
-			return r.Update(ctx, &ctpUpdated)
-		})
-		if err != nil {
-			return fmt.Errorf("failed to update ChangeTransferPolicy %q: %w", ctp.Name, err)
+		// Use the enqueue function to trigger reconciliation
+		if r.EnqueueCTP != nil {
```
crenshaw-dev, line 231: Is there any valid scenario where this is nil? Maybe just allow the panic? If it panics, we screwed up somewhere.

internal/controller/promotionstrategy_controller.go
```diff
@@ -445,18 +537,66 @@ func (r *PromotionStrategyReconciler) updatePreviousEnvironmentCommitStatus(ctx
...
+	// The NoteDrySha (from git note) is the source for which dry SHA the hydrator
+	// has processed. This is important because when manifests don't change between dry commits,
+	// the hydrator may only update the git note without creating a new commit.
+	// In that case, the hydrator.metadata file (which sets Proposed.Dry.Sha) still contains
+	// the old dry SHA, but the git note confirms hydration is complete for the new dry SHA.
+	//
+	// Priority:
+	// 1. If NoteDrySha is set, use it accounts for no-change hydrations
+	// 2. Fall back to Proposed.Dry.Sha (for hydrators that don't use git notes)
+	previousEnvNoteDrySha := previousEnvironmentStatus.Proposed.Hydrated.NoteDrySha
```
crenshaw-dev: Maybe `previousEnvironmentStatus.Proposed.Dry.NoteSha`?

zachaller: done

internal/git/git.go
```diff
@@ -592,6 +512,60 @@ func AddTrailerToCommitMessage(ctx context.Context, commitMessage, trailerKey, t
...
+// FetchNotes fetches the git notes from the remote repository.
+func (g *EnvironmentOperations) FetchNotes(ctx context.Context) error {
+	logger := log.FromContext(ctx)
+	gitPath := gitpaths.Get(g.gap.GetGitHttpsRepoUrl(*g.gitRepo) + g.activeBranch)
+	if gitPath == "" {
+		return fmt.Errorf("no repo path found for repo %q", g.gitRepo.Name)
+	}
+
+	// Fetch the notes ref from origin. We use + to force update in case of divergence.
+	_, stderr, err := g.runCmd(ctx, gitPath, "fetch", "origin", "+"+HydratorNotesRef+":"+HydratorNotesRef)
```
crenshaw-dev, line 525: I think this should have a `metrics.RecordGitOperation` since it's a network-bound action.

zachaller: done

api/v1alpha1/changetransferpolicy_types.go
```diff
@@ -83,13 +83,34 @@ type CommitBranchState struct {
...
+	// Author is the author of the dry commit that was used to hydrate the branch.
+	Author string `json:"author,omitempty"`
+	// Date is the date of the dry commit that was used to hydrate the branch.
+	Date metav1.Time `json:"date,omitempty"`
+	// Subject is the subject line of the dry commit that was used to hydrate the branch.
+	Subject string `json:"subject,omitempty"`
+	// Body is the body of the dry commit that was used to hydrate the branch without the subject.
+	Body string `json:"body,omitempty"`
+	// References are the references to other commits, that went into the hydration of the branch.
+	References []RevisionReference `json:"references,omitempty"`
```
crenshaw-dev, line 111: Maybe just limit this struct to the `DrySha` field until we know for sure we're adding more?

zachaller: I would need to add a new struct for it I personally like the idea of being consistent with the 3 types where we store that information.

PR #758, opened 2025-12-05
test: add unit tests for metrics

emirot:
## What

Adding unit tests for `RecordSCMCall` function

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -48,6 +48,7 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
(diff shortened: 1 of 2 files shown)


crenshaw-dev:
Can you convert the testing framework to Ginkgo? We don't have things set up to use the testing lib.

emirot:
> Can you convert the testing framework to Ginkgo? We don't have things set up to use the testing lib.

@crenshaw-dev I did, but I don't understand why I don't see the line covered in codecov or the change in percentage.

emirot:
@crenshaw-dev I see the improvement here https://app.codecov.io/github/argoproj-labs/gitops-promoter/pull/758 but it does not show lines or report back in the PR

crenshaw-dev:
I think Codecov is only going to show coverage info for lines that actually changed. If you look at the file explorer though, you can see that the relevant lines are covered: https://app.codecov.io/gh/argoproj-labs/gitops-promoter/pull/758/blob/internal/metrics/metrics.go?dropdown=coverage

My guess is that they were already indirectly covered by other tests.

emirot:
> I think Codecov is only going to show coverage info for lines that actually changed. If you look at the file explorer though, you can see that the relevant lines are covered: https://app.codecov.io/gh/argoproj-labs/gitops-promoter/pull/758/blob/internal/metrics/metrics.go?dropdown=coverage
> 
> My guess is that they were already indirectly covered by other tests.

They don't look to be if I look at that function on main branch https://app.codecov.io/gh/argoproj-labs/gitops-promoter/blob/main/internal%2Fmetrics%2Fmetrics.go#L217

emirot:
As seen we @zachaller and @crenshaw-dev we don't want to add this kind of tests

PR #761, opened 2025-12-06
chore update k8s e2e version

emirot:
## What

Rotate k8s version e2e to a newer one


crenshaw-dev (approved):

emirot:
@crenshaw-dev I am wondering why 1.28 is still pending? Would that block to auto merge?

crenshaw-dev:
@emirot branch protection rules still specified 1.28, so i just deleted that rule and added one for 1.31.

PR #765, opened 2025-12-07
chore remove a nil dereference

emirot:
## What 

Return error if resp is nil

## Why

Nilway detected 

```
/Users/nolan/Documents/oss/gitops-promoter/internal/scms/gitlab/utils.go:14:11: error: Potential nil panic detected. Observed nil flow from source to dereference point: 
        - client-go@v1.5.0/merge_requests.go:327:23: unassigned variable `resp` returned from `ListMergeRequests()` in position 1
        - client-go@v1.5.0/merge_requests.go:28:3: returned as result 1 from interface method `MergeRequestsServiceInterface.ListMergeRequests()` (implemented by `MergeRequestsService.ListMergeRequests()`)
        - gitlab/pullrequest.go:259:3: result 1 of `ListMergeRequests()` passed as arg `resp` to `logGitLabRateLimitsIfAvailable()` via the assignment(s):
                - `pr.client.MergeRequests.ListMergeRequests(...)` to `resp` at gitlab/pullrequest.go:249:7
        - gitlab/utils.go:14:11: function parameter `resp` accessed field `Header`
```

### DIFF
--- a/internal/scms/gitlab/pullrequest.go
+++ b/internal/scms/gitlab/pullrequest.go
@@ -248,16 +248,16 @@ func (pr *PullRequest) FindOpen(ctx context.Context, pullRequest v1alpha1.PullRe
 
 	start := time.Now()
 	mrs, resp, err := pr.client.MergeRequests.ListMergeRequests(options)
-	if err != nil {
-		return false, "", time.Time{}, fmt.Errorf("failed to list pull requests: %w", err)
-	}
 	if resp == nil {
+		statusCode := -1
+		metrics.RecordSCMCall(repo, metrics.SCMAPIPullRequest, metrics.SCMOperationList, statusCode, time.Since(start), nil)
 		logger.V(4).Info("gitlab response status", "status", "nil response")
 		return false, "", time.Time{}, errors.New("received nil response from GitLab API")
 	}
-
 	metrics.RecordSCMCall(repo, metrics.SCMAPIPullRequest, metrics.SCMOperationList, resp.StatusCode, time.Since(start), nil)
-
+	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/scms/gitlab/pullrequest.go
```diff
@@ -247,12 +248,15 @@ func (pr *PullRequest) FindOpen(ctx context.Context, pullRequest v1alpha1.PullRe
 
 	start := time.Now()
 	mrs, resp, err := pr.client.MergeRequests.ListMergeRequests(options)
-	if resp != nil {
-		metrics.RecordSCMCall(repo, metrics.SCMAPIPullRequest, metrics.SCMOperationList, resp.StatusCode, time.Since(start), nil)
-	}
 	if err != nil {
 		return false, "", time.Time{}, fmt.Errorf("failed to list pull requests: %w", err)
```
crenshaw-dev: Does the gitlab client return an error for non-200 responses? If so, we still want to make sure we push the metric for the failed request.

emirot: From what I see it does return an error for non-200 response code https://gitlab.com/gitlab-org/api/client-go/-/blob/main/merge_requests.go#L316

crenshaw-dev: Then I don't think we can move the metrics call down...

But we can still add the `if resp == nil` check, I think that's valid and fine.

emirot: Thanks @crenshaw-dev for the explanation, I have updated

emirot:
Closing in favor of https://github.com/argoproj-labs/gitops-promoter/commit/6b6f76dc3d47b140b881a8864febcc75413e2c67

PR #802, opened 2025-12-16
fix: avoid reconciliation for same errors in different order

pbhatnagar-oss:
avoid reconciliation for same errors in different order. The ArgoCDCommitStatus resource is constantly updated by the controller when there is an error such as the one below. The order in the error message array changes causing a reconcile.

### DIFF
--- a/internal/controller/argocdcommitstatus_controller_test.go
+++ b/internal/controller/argocdcommitstatus_controller_test.go
@@ -515,11 +515,13 @@ var _ = Describe("ArgoCDCommitStatus Controller", func() {
 				updated := &promoterv1alpha1.ArgoCDCommitStatus{}
 				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: "sorting-test", Namespace: "default"}, updated)).To(Succeed())
 
-				// Trigger reconciliation by updating an annotation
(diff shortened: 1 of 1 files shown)


internal/controller/argocdcommitstatus_controller_test.go
```diff
@@ -499,14 +500,44 @@ var _ = Describe("ArgoCDCommitStatus Controller", func() {
...
+			// This catches nondeterministic behavior that the controller's map iteration would cause
+			for i := 0; i < 5; i++ {
+				updated := &promoterv1alpha1.ArgoCDCommitStatus{}
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: "sorting-test", Namespace: "default"}, updated)).To(Succeed())
+
+				// Trigger reconciliation by updating an annotation
+				if updated.Annotations == nil {
+					updated.Annotations = make(map[string]string)
+				}
+				updated.Annotations["test-iteration"] = fmt.Sprintf("%d", i)
```
crenshaw-dev: I don't think annotations trigger reconciles in ACDCS. Can we directly enqueue?

zachaller: good catch

PR #806, opened 2025-12-17
fix: rate limit ps enqueuing of ctp's

zachaller:


internal/controller/promotionstrategy_controller.go
```diff
@@ -341,14 +342,64 @@ func (r *PromotionStrategyReconciler) calculateStatus(ps *promoterv1alpha1.Promo
...
+		//       * bool: 1 byte + 7 bytes padding = 8 bytes
+		//   - Map overhead: ~8 bytes per entry
+		//   Total: ~144 bytes per CTP
+		//
+		// Memory bounds (assuming 32-char namespace and name):
+		//   - 100 stale entries = ~14 KB
+		//   - 1,000 stale entries = ~144 KB
+		//   - 10,000 stale entries = ~1.4 MB
+		//
+		// With 1 hour cleanup interval, worst case is 1 hour of deleted CTPs in memory.
```
crenshaw-dev: Probably okay to move the memory analysis to the PR description.

internal/controller/promotionstrategy_controller.go
```diff
@@ -376,32 +437,83 @@ func (r *PromotionStrategyReconciler) enqueueOutOfSyncCTPs(ctx context.Context,
...
+
+			continue
+		}
+
+		// Not rate limited - enqueue immediately
+		state.lastEnqueueTime = now
+		state.hasScheduledRetry = false
+		r.enqueueStateMutex.Unlock()
+
+		logger.V(4).Info("Enqueueing out-of-sync CTP",
```
crenshaw-dev: This is a lot to embed in this function, can it be its own function call?

internal/controller/promotionstrategy_controller.go
```diff
@@ -341,14 +342,64 @@ func (r *PromotionStrategyReconciler) calculateStatus(ps *promoterv1alpha1.Promo
...
+		r.enqueueStateMutex.Lock()
+		if r.enqueueStates == nil {
+			r.enqueueStates = make(map[client.ObjectKey]*ctpEnqueueState)
+		}
+		r.enqueueStateMutex.Unlock()
+	}
+
+	// Lazy initialize map and start background cleanup timer
+	if r.enqueueStates == nil {
+		r.enqueueStates = make(map[client.ObjectKey]*ctpEnqueueState)
```
crenshaw-dev: Doesn't the above guarantee that this can't be hit?

internal/controller/promotionstrategy_controller.go
zachaller: ehh.. sure I kind of liked it there but can just remove the whole thing if you don't like it, it nice to know what is in each cache that we have. I imagine something like that would be really nice in where argocd had an in memory cache

internal/controller/promotionstrategy_controller.go
zachaller: lol yea I missed that I thought it was int he first nil check, good spot

internal/controller/promotionstrategy_controller.go
zachaller: I kind of like it in one with all the mutex work being contained, but looking into it more

internal/controller/promotionstrategy_controller.go
zachaller: moved it around a bit

internal/controller/promotionstrategy_controller.go
```diff
@@ -376,32 +388,148 @@ func (r *PromotionStrategyReconciler) enqueueOutOfSyncCTPs(ctx context.Context,
...
+// - Schedules a delayed retry if within the rate limit threshold
+// - Enqueues immediately if not rate limited
+func (r *PromotionStrategyReconciler) handleRateLimitedEnqueue(
+	ctx context.Context,
+	key client.ObjectKey,
+	ctpName string,
+	effectiveSha string,
+	targetSha string,
+	now time.Time,
+	enqueueThreshold time.Duration,
```
crenshaw-dev: Could be a global const?

internal/controller/promotionstrategy_controller.go
```diff
@@ -376,32 +388,148 @@ func (r *PromotionStrategyReconciler) enqueueOutOfSyncCTPs(ctx context.Context,
...
+
+// handleRateLimitedEnqueue applies rate limiting to a CTP enqueue request.
+// It checks if the CTP was enqueued recently and either:
+// - Skips if a delayed retry is already scheduled
+// - Schedules a delayed retry if within the rate limit threshold
+// - Enqueues immediately if not rate limited
+func (r *PromotionStrategyReconciler) handleRateLimitedEnqueue(
+	ctx context.Context,
+	key client.ObjectKey,
+	ctpName string,
```
crenshaw-dev: Could we just get this from the object key?

internal/controller/promotionstrategy_controller.go
```diff
@@ -376,32 +388,148 @@ func (r *PromotionStrategyReconciler) enqueueOutOfSyncCTPs(ctx context.Context,
...
+	effectiveSha string,
+	targetSha string,
+	now time.Time,
```
crenshaw-dev: Would this be different from just using time.Now()?

internal/controller/promotionstrategy_controller.go
```diff
@@ -376,32 +388,148 @@ func (r *PromotionStrategyReconciler) enqueueOutOfSyncCTPs(ctx context.Context,
...
+// It checks if the CTP was enqueued recently and either:
+// - Skips if a delayed retry is already scheduled
+// - Schedules a delayed retry if within the rate limit threshold
+// - Enqueues immediately if not rate limited
+func (r *PromotionStrategyReconciler) handleRateLimitedEnqueue(
+	ctx context.Context,
+	key client.ObjectKey,
+	ctpName string,
+	effectiveSha string,
+	targetSha string,
```
crenshaw-dev: Instead of passing these explicitly, just add them to the logging context?

internal/controller/promotionstrategy_controller.go
zachaller: going to just move this into the function instead it's not used anywhere but one spot not sure that is worth the global const

crenshaw-dev (approved):

PR #808, opened 2025-12-17
chore(acdcs): simplify app phase calculation

crenshaw-dev:


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -75,11 +75,6 @@ var (
 	revMap  = make(map[appRevisionKey]string)
 )
 
-type aggregate struct {
-	application  *argocd.Application
-	commitStatus *promoterv1alpha1.CommitStatus
-}
```
crenshaw-dev, line 74: We don't need this struct at all, because everything in `commitStatus` is trivially derived from `application`.

PR #810, opened 2025-12-17
chore(acdcs): remove unnecessary reconcile filter

crenshaw-dev:
I think now that we have the predicate, this filter isn't needed.

I've tested internally and haven't seen an explosion of ArgoCDCommitStatus reconciles.

### DIFF
--- a/internal/controller/argocdcommitstatus_controller.go
+++ b/internal/controller/argocdcommitstatus_controller.go
@@ -73,12 +73,6 @@ type aggregate struct {
 	commitStatus *promoterv1alpha1.CommitStatus
 }
 
-type appRevisionKey struct {
-	clusterName string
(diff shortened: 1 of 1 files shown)


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -450,40 +437,15 @@ func lookupArgoCDCommitStatusFromArgoCDApplication(mgr mcmanager.Manager) mchand
 
 			application := &argocd.Application{}
 
-			// if clusterName is empty, then cluster == local cluster
-			appKey := appRevisionKey{
-				clusterName: clusterName,
```
zachaller, line 456: It's kind of nice that the mc-runtime takes care of this cluster part of filtering.

PR #815, opened 2025-12-18
chore: don't update resource if already deleted

crenshaw-dev:
This reduced the number of errors like `Precondition failed: UID in precondition: 7a65beca-85d0-48be-b58c-fa53de735e9a, UID in object meta` from 37 in one run to 8.

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -275,7 +275,7 @@ func HandleReconciliationResult(
 	}
 
 	// If the deletion timestamp is set on the object, bail out early.
-	if obj.GetDeletionTimestamp() != nil {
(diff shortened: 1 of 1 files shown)


emirot (approved):

internal/utils/utils.go
```diff
@@ -274,6 +274,12 @@ func HandleReconciliationResult(
 		return
 	}
 
+	// If the deletion timestamp is set on the object, bail out early.
+	if obj.GetDeletionTimestamp() != nil {
```
zachaller: !obj.GetDeletionTimestamp().IsZero()?

crenshaw-dev: fixed!

zachaller (approved):

PR #820, opened 2025-12-19
chore: add nilaways

emirot:
## What

Adding nialawys in CI and fixing issues.


.github/workflows/ci.yaml
```diff
@@ -67,9 +67,9 @@ jobs:
           version: v2.5.0
           args: --timeout=5m
           skip-cache: false
-#      # We want to enable this but github seems to cancel the job, might be related to memory limits on the runner.
-#      - name: Nilaway
-#        run: make nilaway-no-test
+      # We want to enable this but github seems to cancel the job, might be related to memory limits on the runner.
+      - name: Nilaway
```
zachaller: Due to the length of time this takes to run I am thinking we pull it into its own job so it runs in parrellel

emirot: @zachaller 1minute looks really reasonable no ?
Screenshot 2025-12-19 at 9 05 41 PM

zachaller: Yea, even at a minute there is so much time people wait for the CI to pass I tink every minute we can reduce is worth it. I think you can do a new Job inside the same CI file even and it should just work I think?

Plus as the code base continues to grow I think this time will also grow quite a bit as well compared to just adding tests for new code so yea...

emirot: @zachaller sounds good I have updated

PR #823, opened 2025-12-22
chore: more validation

crenshaw-dev:


api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -105,25 +105,40 @@ func (cs *ArgoCDCommitStatus) GetConditions() *[]metav1.Condition {
...
 }
 
-// ApplicationsSelected represents the Argo CD applications that are selected by the commit status.
+// ApplicationsSelected represents the Argo CD applications that are selected by the commit status. The fields in this
+// struct are all required, since the controller should always fully construct this information.
 type ApplicationsSelected struct {
 	// Namespace is the namespace of the Argo CD application.
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
```
zachaller: namespaces can only be 63 chars: 

https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#namespaces-and-dns

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

api/v1alpha1/commitstatus_types.go
```diff
@@ -31,18 +31,22 @@ type CommitStatusSpec struct {
 	// +kubebuilder:validation:Required
 	RepositoryReference ObjectReference `json:"gitRepositoryRef"`
 
+	// SHA is the commit SHA to set the status on.
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=64
-	// +kubebuilder:validation:Pattern=`^[a-fA-F0-9]+$`
+	// +kubebuilder:validation:MinLength=40
```
zachaller, line 37: Are these length not redundant on the regex, I also notice in some we have min and others we just have regex

api/v1alpha1/gitcommitstatus_types.go
```diff
@@ -133,20 +134,27 @@ type GitCommitStatusEnvironmentStatus struct {
 	// promotions based on the validation of the ACTIVE commit.
 	// May be empty if the PromotionStrategy hasn't reconciled yet.
 	// +required
+	// +kubebuilder:validation:MaxLength=40
```
zachaller: same comment missing min but have regex set to have to be 40, we should make all the same

api/v1alpha1/commitstatus_types.go
crenshaw-dev: The length isn't redundant on regexes for fields that are optional, because those fields can't have a MinLength.

I added some docs explaining why the sha fields are validated the way they are, take a glance at that and see if it makes sense.

zachaller:
Might want to consider: https://git-scm.com/docs/hash-function-transition

PR #824, opened 2025-12-22
chore: add git refspec validation

crenshaw-dev:
I only managed to get one of three CEL validations in under the performance estimator check. Still, better than nothing probably?


crenshaw-dev:
Probably more trouble than it's worth.

PR #845, opened 2026-01-08
chore: update gorelease version

emirot:
## What

update gorelease version

## Test 

```
➜  gitops-promoter git:(chore/update/gorelease) ✗ goreleaser check
  • checking                                  path=.goreleaser.yaml
  • dockers and docker_manifests are being phased out and will eventually be replaced by dockers_v2, check https://goreleaser.com/deprecations#dockers for more info
  • dockers and docker_manifests are being phased out and will eventually be replaced by dockers_v2, check https://goreleaser.com/deprecations#dockers for more info
  • 1 configuration file(s) validated
  • thanks for using GoReleaser!
```

### DIFF
--- a/ui/dashboard/package-lock.json
+++ b/ui/dashboard/package-lock.json
@@ -75,6 +75,7 @@
       "integrity": "sha512-UlLAnTPrFdNGoFtbSXwcGFQBtQZJCNjaN6hQNP3UPvuNXT1i82N26KL3dZeIpNalWywr9IuQuncaAfUaS1g6sQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -76,7 +76,7 @@ docker_manifests:
 archives:
   - id: binary
     formats: [tar.gz]
-    builds:
+    ids:
```
emirot, line 79: Why this change?

Otherwise I would get
 • DEPRECATED: archives.builds should not be used anymore, check https://goreleaser.com/deprecations#archivesbuilds for more info

zachaller:
Can we update the Makefile as well

emirot:
> Can we update the Makefile as well

Thanks @zachaller good call

PR #851, opened 2026-01-09
feat: add docker image latest on every push to main branch

emirot:
## What

add docker image latest on every push from main

## Why

Currently, there is no latest image pushed, so by pulling main branch and installing gitops promoter in k8s it won't find the latest image and would fail to deploy.

## Extra

I wanted to avoid duplication of ci and gorelease file but from what I saw the duplication cannot be avoided it will be in the same file and not necessarily better.
Also see https://github.com/orgs/goreleaser/discussions/2132

### DIFF
--- a/ui/components-lib/src/components/LiveManifestView.scss
+++ b/ui/components-lib/src/components/LiveManifestView.scss
@@ -1,4 +1,4 @@
-@import '../styles/_colors.scss';
+@use '../styles/_colors.scss';
 
 .strategy-json-editor {
   background: white;
@@ -7,7 +7,7 @@
   border-radius: 8px;
   box-sizing: border-box;
   box-shadow: 0 1px 3px rgba(0, 0, 0, 0.1), 0 1px 2px rgba(0, 0, 0, 0.06);
-  border: 1px solid $argo-color-gray-4;
+  border: 1px solid colors.$argo-color-gray-4;
   margin: 2rem 2.5rem;
   height: calc(100vh - 140px);
   display: flex;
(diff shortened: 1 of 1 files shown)


.github/workflows/release-latest.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Setup Go
+        uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
+        with:
+          go-version: "1.25"
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
crenshaw-dev: Wanna go ahead and pin this?

emirot: I do not need that as the other workflow does not include it, I removed it

zachaller:
Yea I don't love the copy of files mostly but it was also the only way I could think of doing it, not sure if we want to add a comment or something to the two files mentioning keeping them in sync if changes are made that are in the shared parts?

Other than, that I do think the feature is worth the duplication is LGTM

emirot:
> Yea I don't love the copy of files mostly but it was also the only way I could think of doing it, not sure if we want to add a comment or something to the two files mentioning keeping them in sync if changes are made that are in the shared parts?
> 
> Other than, that I do think the feature is worth the duplication is LGTM

@zachaller 
Yes hard to avoid duplication in that case and when tried it made things more convoluted 
I have added your suggestions https://github.com/argoproj-labs/gitops-promoter/pull/851/commits/8d331430f00d9ca66277f75753e05e94c5148943

zachaller (approved):

PR #859, opened 2026-01-12
feat: switch to in-tree lib for webhook processing

zachaller:
a step towards: https://cloud-native.slack.com/archives/C06NXPYQ06B/p1765536752779509

Also going to start trigginer on PullReqest events so make sense to start using this

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/gin-contrib/gzip v1.2.5
 	github.com/gin-gonic/gin v1.11.0
 	github.com/go-logr/logr v1.4.3
+	github.com/go-playground/webhooks/v6 v6.4.0
 	github.com/go-task/slim-sprig/v3 v3.0.0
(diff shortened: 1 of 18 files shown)


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/gin-contrib/gzip v1.2.5
 	github.com/gin-gonic/gin v1.11.0
 	github.com/go-logr/logr v1.4.3
+	github.com/go-playground/webhooks/v6 v6.4.0
```
crenshaw-dev: This library isn't really maintained... not sure how we should feel about that. https://github.com/go-playground/webhooks/issues/167

zachaller: argcd is using this as well, so if it's not maintained maybe we maintain it there really isn't much out there for this.

zachaller: https://github.com/argoproj/argo-cd/blob/master/go.mod#L43

PR #883, opened 2026-01-15
feat: add tiltfile 

emirot:
## What

Add a titlfile 

## Why

To help with local dev

## Test

1. change a go internal log
It would restart the pod and the log would be present
E.g
```
[manager] 2026-01-13T03:08:49.642246943Z	INFO	Reconciling PromotionStrategy nolan2	{"controller": "promotionstrategy", "controllerGroup": "promoter.argoproj.io", "controllerKind": "PromotionStrategy", "PromotionStrategy": {"name":"demo-gitea","namespace":"default"}, "namespace": "default", "name": "demo-gitea", "reconcileID": "00a85ac5-2e63-4348-bb7e-bc64b4f78271"}
```

2. Make a UI change

Making a simple UI change in TopBar.tsx in the title
It takes less than 10 seconds to display the change and be displayed on localhost:8080

## Extra

Another docker image could be created just for that dev flow which could make the rebuild faster but it is already good I believe could be introduce later on if we feel the need to

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -38,10 +38,15 @@ local_resource(
     allow_parallel=True,
 )
 
+# detect cluster architecture for build
+cluster_version = decode_yaml(local('kubectl version -o yaml'))
+platform = cluster_version['serverVersion']['platform']
+arch = platform.split('/')[1]
+
 # Build binary locally
 local_resource(
     'build-binary',
-    cmd='CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/gitops-promoter ./cmd',
+    cmd='CGO_ENABLED=0 GOOS=linux GOARCH=' + arch + ' go build -o bin/gitops-promoter ./cmd',
     deps=['cmd', 'api', 'internal'],
     resource_deps=['build-dashboard'],
     labels=['build'],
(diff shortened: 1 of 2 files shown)


Tiltfile
```diff
@@ -0,0 +1,98 @@
...
+    context='.',
+    dockerfile='Dockerfile',
+    only=[
+        'cmd', 'api', 'internal', 'ui', 'hack/git',
+        'go.mod', 'go.sum', 'Dockerfile',
+    ],
+    ignore=[
+        '**/node_modules',
+    ],
+)
```
FourFifthsCode: one thing I really like about Tilt is its ability to sync files directly into the container and restarting its process without needing to rebuild the image on code changes. I recommend using docker build with restart for this to improve the feedback loop speed. See example of this in https://github.com/argoproj/argo-cd/blob/master/Tiltfile#L80

Tiltfile
```diff
@@ -0,0 +1,98 @@
+# -*- mode: Python -*-
+
+# Tiltfile for gitops-promoter
+
+# Allow using a custom Kubernetes context
+allow_k8s_contexts(k8s_context())
```
FourFifthsCode: We might want to make this optional by either a config file or passing in arguments. It's too easy sometimes to be connected to a production cluster and forget and tilt's default mode will prevent accidental deployments

Tiltfile
```diff
@@ -0,0 +1,98 @@
...
+        'promoter-manager-rolebinding:clusterrolebinding',
+        'promoter-leader-election-role:role',
+        'promoter-leader-election-rolebinding:rolebinding',
+    ],
+    resource_deps=['generate-manifests'],
+    labels=['promoter'],
+    port_forwards=[
+        port_forward(8081, 8081, name='health'),
+        port_forward(8443, 8443, name='metrics'),
+        port_forward(3333, 3333, name='webhook'),
```
FourFifthsCode, line 104: Love the port forward setup!

Tiltfile
```diff
@@ -0,0 +1,98 @@
...
+
+# Tiltfile for gitops-promoter
+
+# Allow using a custom Kubernetes context
+allow_k8s_contexts(k8s_context())
+
+# Configuration
+config.define_string('registry', args=True, usage='Container registry to push to (e.g., ttl.sh, localhost:5000)')
+cfg = config.parse()
+registry = cfg.get('registry', 'ttl.sh')
```
FourFifthsCode: this config is new to me, I'm curious to what it does! Could you explain it a bit?

emirot: This is not needed it is already setup by Tilt https://docs.tilt.dev/personal_registry.html

Tiltfile
```diff
@@ -0,0 +1,96 @@
+# -*- mode: Python -*-
+
```
FourFifthsCode, line 2: should add:
```
load('ext://restart_process', 'docker_build_with_restart')
```

Tiltfile
```diff
@@ -0,0 +1,96 @@
...
+    resource_deps=['build-dashboard'],
+    labels=['build'],
+)
+
+# =============================================================================
+# Kubernetes Deployment (Controller runs in K8s)
+# =============================================================================
+
+# Build Docker image for K8s deployment
+docker_build(
```
FourFifthsCode: should be:
```
docker_build_with_restart(
```

PR #894, opened 2026-01-19
chore(version): update promoter version

emirot:
## What

~Update version to latest~

~I don't know how this is used and why it is so old, if it matters or not but it looks wrong.~

Removing version pin


config/release/kustomization.yaml
```diff
@@ -4,6 +4,6 @@ resources:
 images:
 - name: quay.io/argoprojlabs/gitops-promoter
   newName: quay.io/argoprojlabs/gitops-promoter
-  newTag: v0.12.0
+  newTag: v0.20.2
```
zachaller: I think now that we push latests this should actually be latest now or just remove the newTag, newName?

emirot: @zachaller I have updated to latest, I believe newTag and newName can be removed but how to be sure?

emirot: But I guess if we remove it this folder become useless right ? 

So maybe it should be https://github.com/argoproj-labs/gitops-promoter/pull/913

zachaller (approved):

PR #922, opened 2026-01-21
fix(ui): page crashing

emirot:
## What

Fix page crashing

### Before

Screenshot 2026-01-21 at 8 47 22 AM

### After

Screenshot 2026-01-21 at 8 45 56 AM

## Why

I believe this was caused when we updated dependencies, it looks like `Without dedupe, Vite bundles both copies of React` from what I could find


ui/dashboard/vite.config.ts
```diff
@@ -32,5 +32,6 @@ export default defineConfig({
       '@lib': resolve(__dirname, '../components-lib/src'),
       '@shared': resolve(__dirname, '../shared/src'),
     },
+    dedupe: ['react', 'react-dom'],
```
jwinters01, line 35: I believe there is a way to specify a dependency as "external" so that it doesn't get built into the bundle but instead is borrowed from the consuming app, might be worth looking into doing that instead as that is our intent

PR #928, opened 2026-01-22
feat: auto configure timed commit status checks

zachaller:


zachaller:
There is a race with this setup

PR #930, opened 2026-01-22
docs: argocd multi instances

emirot:
## What

Add information about argocd single vs multi instance

## Why 

It is a recurring question

### DIFF
--- a/docs/faqs.md
+++ b/docs/faqs.md
@@ -27,6 +27,10 @@ applied to all environments.
 By focusing on the promotion part and leaving manifest manipulation to other tools, GitOps Promoter is able to reliably
 handle whatever manifest structure your organization prefers.
 
+## Do I need Argo CD to use GitOps Promoter?
+No, GitOps Promoter works with any tools that can provide hydrated manifests in Git.
+Check the [Custom Hydration](./custom-hydrator.md) documentation for more details.
+
 ## Why Should I use Webhooks?
 
 Webhooks allow GitOps Promoter to react to events from your Git repository in real-time. This means that when you push


docs/faqs.md
```diff
@@ -10,6 +10,11 @@ The problems with using branches for environments are all related to the user ex
 When using GitOps promoter, the user only pushes to a single branch. Hydrated manifests are automatically managed in
 environment-specific branches, but the user never has to interact with them directly.
 
+## Does Gitops-Promoter support multi Argo CD installations?
```
crenshaw-dev:
```suggestion
## Does GitOps Promoter support multi Argo CD installations?
```

Branding! 😁

crenshaw-dev:
Should we also note that ArgoCDCommitStatus can monitor Argo CD instances across multiple clusters? Or is that too in the weeds?

emirot:
> Should we also note that ArgoCDCommitStatus can monitor Argo CD instances across multiple clusters? Or is that too in the weeds?

Maybe it belongs more here https://gitops-promoter.readthedocs.io/en/latest/crd-specs/#argocdcommitstatus

crenshaw-dev:
That's fair... maybe jist a quick mention thay we support multiple instances _even if theyre hosted on multiple clusters_? Because lots of folks host multiple instances on one cluster,  so we want to be clear that we're super flexible.

zachaller:
Should we mention that all apps must exist on one instance aka you can't have prod apps split instances.

crenshaw-dev:
Well you can split prod across instances, but not if you want all the prod apps to be delivered as a single "environment." But that might be more than we can explain in the FAQ. I think most people will be doing something that works fine.

emirot:
> Should we mention that all apps must exist on one instance aka you can't have prod apps split instances.

No sure I understand that

emirot:
> LGTM, will wait for Zach to approve in case he wants the clarity on envs split across instances.

I don't understand what would be the use case

prod-argocd1-east  ( app1 )
prod-argocd2-west  ( app1) 


It cannot deploy app 1 across those 2 clusters ?

crenshaw-dev:
> It cannot deploy app 1 across those 2 clusters ?

Not if you're using Argo CD Source Hydrator, because Source Hydrator can't coordinate a single commit across two Argo CD instances.

zachaller:
> Well you can split prod across instances, but not if you want all the prod apps to be delivered as a single "environment." But that might be more than we can explain in the FAQ. I think most people will be doing something that works fine.

I agree that the most common setups will just naturally work so tldr is no need to document it @emirot

PR #935, opened 2026-01-23
fix: run test-e2e locally

emirot:
## What

In CI it will create a kind cluster, but locally it will fails if a cluster is not created with that particular name


zachaller:
Not sure this is even worth doing until we have real e2e tests?

PR #937, opened 2026-01-24
feat(ui): add commit status description

crenshaw-dev:
image

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -16,4 +16,5 @@ ignore:
   - 'dist'
   - 'hack'
   - 'internal/scms/mock'
+  - 'internal/controller/testutils'
   - '**/zz_generated.*'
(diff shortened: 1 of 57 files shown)


ui/components-lib/src/components/HealthSummary.tsx
```diff
@@ -58,6 +94,24 @@ const HealthSummary: React.FC<HealthSummaryProps> = ({ checks, title, status, he
...
+                }}
+                onMouseEnter={(e) => handleMouseEnter(index, e)}
+                onMouseLeave={handleMouseLeave}
+              >
+                <div className="health-tooltip">
+                  {check.description}
+                </div>
+              </div>
+            )
+          )}
```
jwinters01: It is generally better to build a Tooltip component that wraps the component you want it to target, for the sake of better understanding how everything relates to each other. You can probably ask the robots to use our implementation in [argo-ui](https://github.com/argoproj/argo-ui/blob/master/v2/components/tooltip/tooltip.tsx) as a reference to build a Tooltip component in this repo. That would also help clean up some of that mouse movement code as well.

PR #942, opened 2026-01-26
Fix linting on switch case

emirot:
## What

It looks like in some cases the linter will fail  intermittently https://github.com/argoproj-labs/gitops-promoter/actions/runs/21372990575/job/61522099710 latest CI run on Main branch

I have seen that happening already and  after restarting the job it worked.

I could not reproduce locally, but this lint exception is not needed so it can be removed.


emirot:
Also https://github.com/argoproj-labs/gitops-promoter/actions/runs/21372990575/job/61522099710

emirot:
And also https://github.com/argoproj-labs/gitops-promoter/actions/runs/21372990575/job/61522099710 why we see that now and not locally interesting ...

zachaller:
I want to try and keep the code, and fix the linting issue instead.

emirot:
> I want to try and keep the code, and fix the linting issue instead.

I don't believe it is a code issue, most likely a cache issue 

@zachaller  would you prefer to try https://github.com/argoproj-labs/gitops-promoter/pull/944

emirot:
It is intermittent https://github.com/argoproj-labs/gitops-promoter/actions/runs/21372990575/job/61537577731 when I relaunch it sometimes it passes, I believe this is the best fix for now?

PR #943, opened 2026-01-26
fix: helm templating

emirot:
## What

Commenting out example in yaml because it gets interpreted by helm and will make `helm install` fail.

## Why

Otherwise those commands will fail, because it will interpret the helm template

```
Error: INSTALLATION FAILED: template: promoter/templates/crd/argocdcommitstatuses.promoter.argoproj.io.yaml:146:91: executing "promoter/templates/crd/argocdcommitstatuses.promoter.argoproj.io.yaml" at <.ArgoCDCommitStatus.Spec.ApplicationSelector.MatchLabels>: nil pointer evaluating interface {}.Spec
```

With the change
```
./kubebuilder edit --plugins=helm/v2-alpha --output-dir=charts   
helm install promoter ./charts/chart --namespace promoter-system --create-namespace  --set crds.enabled=true
NAME: promoter
LAST DEPLOYED: Mon Jan 26 13:35:12 2026
NAMESPACE: promoter-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

### DIFF
--- a/dist/install.yaml
+++ b/dist/install.yaml
@@ -148,19 +148,18 @@ spec:
 
                       Example:
 
-                        {{- $baseURL := "https://dev.argocd.local" -}}
-                        {{- if eq .Environment "environment/development" -}}
-                        {{- $baseURL = "https://dev.argocd.local" -}}
-                        {{- else if eq .Environment "environment/staging" -}}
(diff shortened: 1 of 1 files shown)


emirot:
@crenshaw-dev sounds good I will close the PR

PR #985, opened 2026-02-02
fix: do not require Application CRD on startup

crenshaw-dev:


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -138,12 +137,25 @@ func (r *ArgoCDCommitStatusReconciler) Reconcile(ctx context.Context, req mcreco
...
 	}
+
+	// Try to add the local cluster - it will only be available if the ControllerConfigurationReconciler
+	// has engaged it based on the WatchLocalApplications configuration
+	localClusterName := LocalProviderName + "#" + mcmanager.LocalCluster
+	_, err = r.Manager.GetCluster(ctx, localClusterName)
+	if err == nil {
+		// Local cluster is available, add it to the list
+		clusters = append(clusters, localClusterName)
+	}
```
FourFifthsCode, line 156: Seems straight forward enough

internal/controller/controllerconfiguration_controller.go
```diff
@@ -19,47 +19,227 @@ package controller
...
+	defer r.lock.Unlock()
+
+	if !r.watchLocalApplicationsEngaged {
+		return // Not engaged
+	}
+
+	// Cancel the context to trigger cleanup in multicluster-runtime
+	if r.cancelFunc != nil {
+		r.cancelFunc()
+	}
```
FourFifthsCode, line 189: looks good as its the only current way to disengage a cluster

FourFifthsCode (approved):
The overall approach looks sound. I wish we had more configuration built into multi-cluster framework, but this looks to be the best way forward given the current constraints. Tests look good to ensure we are properly engaging and disengaging. PR looks good and seems pretty close to mergable as is.

Issue #989, opened 2026-02-03
Removing an environment from a PromotionStrategy doesn't update promoter-previous-env CommitStatus in related CTPs


sym-stiller:
My initial PromotionStrategy looked like this:

```
apiVersion: promoter.argoproj.io/v1alpha1
kind: PromotionStrategy
metadata:
  name: demo
activeCommitStatuses:
    - key: argocd-health
environments:
    - autoMerge: true
      branch: environments/dev
    - autoMerge: false
      branch: environments/staging
    - autoMerge: false
      branch: environments/prod
```

Then I removed the first environment (`environments/dev`) from the list. The `ChangeTransferPolicy` related to the staging environment wasn't updated, and was still waiting for the previous environment to become healthy (because of the built-in default `promoter-previous-env-*` CommitStatus gate). There is no previous environment anymore, so I'm blocked from promoting any changes.

My workaround was to delete and re-apply the PromotionStrategy. After reconciliation finished, everything worked fine again.

crenshaw-dev:
What version are you running?

sym-stiller:
I'm on v0.21.1

crenshaw-dev:
We tested manually, and I added an e2e test to try to reproduce the issue, but we've been unable. Is it possible that, after deleting the first env, the PromotionStrategy failed to reconcile for some reason?

crenshaw-dev:
Closing for now since we've been unable to reproduce the issue, but I'll still try to get the test case merged so we avoid regessions.

sym-stiller:
Hey! Sorry, I missed your last response. I can see a few log messages regarding failed reconciliations of the PromotionStrategy, but these look like concurrency issues that only affect concurrent updates to the `status` of the PS. These appear regularly and usually go away during the next reconciliation. The described issue was persistent for several hours, so I think there is only a slim chance these messages are related to the issue. Besides that, everything works quite well.

I'll try to test this scenario again this week. If I can reproduce the issue, I'll reopen this.

Thanks for your investigation and effort! We already love gitops-promoter and are excited for the next cool features.

sym-stiller:
I have found the root cause of this issue, and it wasn't related to gitops-promoter at all. Hydration was blocked by multiple issues that I outlined [in this issue comment](https://github.com/argoproj/argo-cd/issues/26570#issuecomment-3973309840).

PR #990, opened 2026-02-03
test(ps): cleanup after removing first env (#989)

crenshaw-dev:
Relates to #989


crenshaw-dev:
Closing since the reported issue was unrelated to GitOps Promoter.

PR #1002, opened 2026-02-05
fix: github cs create new when going from success to pending

zachaller:

PR #1008, opened 2026-02-06
fix: make controllers resilient to out-of-date status fields

ZeBidule:
Fixes reconciliation loops in PullRequest and CommitStatus controllers that occur when status fields in Kubernetes resources are out of sync with the actual state in SCM providers.

### Issue 1: PullRequest Reconciliation Loop
- When a merge request is successfully merged in GitLab/GitHub/Forgejo, the `status.state` field is sometimes not updated in the K8S resource
- Controller repeatedly attempts to merge an already-merged PR/MR
- Results in **405** (GitLab) or similar errors from SCM APIs
- Triggers unnecessary pipeline executions on each retry

Example error:
```
Reconciliation failed: failed to merge pull request: PUT https://gitlab_hostname/api/v4/projects/2673/merge_requests/1003/merge: 405 {message: 405 Method Not Allowed}
```

### Issue 2: CommitStatus Reconciliation Loop
- Controller attempts to set commit status even when already set
- GitLab's state machine rejects duplicate status transitions
- Results in **400** errors

Example error:
```
Commit status reconciliation failed: failed to set CommitStatus state: failed to create status: POST https://gitlab_hostname/api/v4/projects/2673/statuses/...: 400 {message: Cannot transition status via :enqueue from :pending}
```

### 1. PullRequest Providers (GitLab, GitHub, Forgejo)
- Check actual PR/MR state from SCM before attempting merge/close operations
- If already merged/closed, log and return success (no-op) instead of making redundant API calls
- Prevents reconciliation errors and unnecessary pipeline triggers

### 2. CommitStatus Controller
- Compare `status.sha` and `status.phase` with `spec` values before making API calls
- Skip redundant status updates when already correctly set
- Prevents GitLab state transition errors

## Impact

✅ Eliminates reconciliation loops caused by stale status fields
✅ Reduces unnecessary SCM API calls
✅ Prevents pipeline spam from repeated reconciliation attempts
✅ Improves resilience to transient status update failures

## Testing

- Code compiles successfully
- Changes tested against the reported scenario (GitOps Promoter v0.21.1 + GitLab v18.8.2)

## Related

This fix addresses the issue reported in the community discussion about promotion blocking due to out-of-sync status fields.

### DIFF
--- a/internal/controller/commitstatus_controller.go
+++ b/internal/controller/commitstatus_controller.go
@@ -101,6 +101,13 @@ func (r *CommitStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	// Check if the status is already set correctly to avoid redundant API calls
+	// This prevents GitLab state transition errors when status is already set
+	if cs.Status.Sha == cs.Spec.Sha && cs.Status.Phase == cs.Spec.Phase {
+		logger.Info("Commit status already set correctly, skipping API call", "sha", cs.Spec.Sha, "phase", cs.Spec.Phase)
+		return ctrl.Result{}, nil
+	}
+
 	commitStatusProvider, err := r.getCommitStatusProvider(ctx, cs)
 	if err != nil || commitStatusProvider == nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get CommitStatus provider: %w", err)
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -62,6 +62,20 @@ make run-dashboard     # Run dashboard locally
 - Use controller-runtime patterns for Kubernetes interactions
 - Avoid naked returns and follow error wrapping conventions
 
+### Linting Workflow
+Before committing any Go code changes, always run:
+```bash
+# Check for common issues
+go vet ./...
+
+# Verify formatting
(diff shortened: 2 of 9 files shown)


internal/controller/commitstatus_controller.go
```diff
@@ -101,6 +101,13 @@ func (r *CommitStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	// Check if the status is already set correctly to avoid redundant API calls
+	// This prevents GitLab state transition errors when status is already set
+	if cs.Status.Sha == cs.Spec.Sha && cs.Status.Phase == cs.Spec.Phase {
+		logger.Info("Commit status already set correctly, skipping API call", "sha", cs.Spec.Sha, "phase", cs.Spec.Phase)
+		return ctrl.Result{}, nil
+	}
```
crenshaw-dev, line 109: This change would make it so that other changes (like the description changing) wouldn't be written to the SCM. 

I think instead of making the change in commitstatus_controller (i.e. for all providers), we should make it on a per-provider basis. This github-specific fix is a good example: https://github.com/argoproj-labs/gitops-promoter/pull/1002

crenshaw-dev:
Thanks for the PR! Superseded by https://github.com/argoproj-labs/gitops-promoter/pull/1106

PR #1037, opened 2026-02-10
fix(ui): use SCM-specific commit URL paths

julianderks:
## Summary

The `getCommitUrl()` function in the ArgoCD extension UI was hardcoded to use `/commit/{sha}` for all SCM providers. This works for GitHub, Gitea, Forgejo, and Azure DevOps, but produces broken links for:

- **Bitbucket Cloud** — expects `/commits/{sha}` (plural)
- **GitLab** — expects `/-/commit/{sha}`

## Fix

Added a domain-based lookup table inside `getCommitUrl()` that maps SCM domains to their correct commit URL path pattern, with `/commit/` as the default fallback. Adding support for a new SCM is a single line addition.

## Images

Hyperlink

before_1

Wrong URL for Bitbucket Cloud

before_2


zachaller:
I like they hybrid option but I dont want to store that in the PS, I think we need some form of aggregation API etc to get all resources related to a PromotionStrategy.

PR #1038, opened 2026-02-10
fix(ui): gracefully handle missing PromotionStrategy in AppView extension

julianderks:
## Summary

The AppView extension crashes with an unhandled error when opened on ArgoCD Applications that don't have a PromotionStrategy resource in their resource tree.

**Error:** `Expected exactly one PromotionStrategy resource in Gitops Promoter Extension, found 0`

This happens because the component throws before React hooks (`useState`, `useEffect`) are called.

## Fix

- Move `useState` and `useEffect` hooks above the conditional checks
- Add an early return guard inside `useEffect` when the count is not exactly 1
- Replace the `throw` with graceful `return` messages after hooks, so the extension renders a message instead of crashing ArgoCD

## Before

Clicking the PromotionStrategy tab on any app without promotion CRDs crashes with `Something went wrong!` and a stacktrace.

image

## After

Shows a simple message: "No PromotionStrategy resource found for this application."

image

### DIFF
--- a/ui/extension/AppViewExtension.tsx
+++ b/ui/extension/AppViewExtension.tsx
@@ -8,17 +8,12 @@ const AppViewExtension = ({ tree, application }: AppViewComponentProps) => {
     () => tree.nodes.filter((node) => node.kind === "PromotionStrategy"),
     [tree.nodes]
   );
-  if (promotionStrategyNodes.length !== 1) {
-    throw new Error(
-      `Expected exactly one PromotionStrategy resource in Gitops Promoter Extension, found ${promotionStrategyNodes.length}`
-    );
-  }
-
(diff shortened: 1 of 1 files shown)


crenshaw-dev:
Thanks for the PR! Is this running Argo CD < 3.3? I think we have a fix in for > 3.3, but this would improve our version support.

julianderks:
@crenshaw-dev I was running [9.2.3](https://artifacthub.io/packages/helm/argo/argo-cd/9.2.3) on this setup. --> v3.2.3

crenshaw-dev:
Gotcha, that makes sense. PR looks good to me, but @jwinters01 will give it one extra look.

crenshaw-dev (approved):
Thank you!

PR #1050, opened 2026-02-14
fix: recover from lost status update on merge pullrequest

emirot:


internal/controller/pullrequest_controller.go
```diff
@@ -412,8 +412,24 @@ func (r *PullRequestReconciler) updatePullRequest(ctx context.Context, pr promot
 }
 
 func (r *PullRequestReconciler) mergePullRequest(ctx context.Context, pr *promoterv1alpha1.PullRequest, provider scms.PullRequestProvider) error {
+	logger := log.FromContext(ctx)
 	mergedTime := metav1.Now()
 
+	// Check if PR is still open before attempting merge
+	found, _, _, err := provider.FindOpen(ctx, *pr)
```
zachaller, line 419: I don't love calling FindOpen again due to API limits, why does re-reconciling not catch and fix this when we return the error.

PR #1069, opened 2026-02-17
chore: refactor context in tests

emirot:


internal/controller/pullrequest_controller_test.go
zachaller, line 62: I actually think this covers doing this already.

zachaller: I think this whole file is done correctly maybe we do this style for all the tests?

PR #1096, opened 2026-02-24
fix: don't send error with requeue

zachaller:
this is just for correctness err can not be non nil


zachaller:
> Just to make sure to understand it is because it should either requeue with no err, or just err not do both what it is doing currently

Correct and in this case err will always be nil

PR #1107, opened 2026-02-26
chore(ci): add tilt ci

emirot:
## What 

Add Task to test Tilt

## Why

If we remove some rbac for instance and do not update tilt or vice versa it will break and we won't know so adding it in ci


FourFifthsCode:
Lets apply argo-cd CRD dependencies after kind cluster is created.

emirot:
@FourFifthsCode I have updated the PR and tested on my fork https://github.com/emirot/gitops-promoter/actions/runs/22457389115/job/65042261647

PR #1147, opened 2026-03-04
chore(acdcs): improve error message when 0 apps are found

crenshaw-dev:


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -176,6 +176,10 @@ func (r *ArgoCDCommitStatusReconciler) Reconcile(ctx context.Context, req mcreco
 	}
 	logger.V(4).Info("Found Applications", "appCount", appCount)
 
+	if appCount == 0 {
+		return ctrl.Result{}, fmt.Errorf("no Argo CD Applications found matching selector %v - check that the clusters successfully engaged", ls)
```
emirot, line 180: Engaged, this means up and running?

crenshaw-dev: yep!

PR #1161, opened 2026-03-06
feat: Scm auth webrequest

zachaller:


api/v1alpha1/httpauth_types.go
```diff
@@ -59,6 +62,13 @@ type HTTPAuthentication struct {
 	// Requires a secret containing the client certificate and private key.
 	// +optional
 	TLS *TLSAuth `json:"tls,omitempty"`
+
+	// ScmAuth specifies authentication using credentials from the SCM provider.
+	// This uses the credentials configured in the ScmProvider referenced by the PromotionStrategy,
+	// applying the appropriate authentication method based on the SCM provider type
+	// (GitHub App, GitLab token, Azure DevOps PAT, etc.).
```
crenshaw-dev, line 69:
```suggestion
	// (GitHub App, GitLab token, Azure DevOps PAT, etc.).
	// To use this auth type, just set it to an empty object, i.e. scm: {}.
```

api/v1alpha1/httpauth_types.go
```diff
@@ -174,3 +184,17 @@ type TLSAuth struct {
...
+// This authentication method uses the credentials from the ScmProvider referenced
+// by the PromotionStrategy. The controller retrieves the SCM provider and secret
+// and applies the appropriate authentication based on the provider type:
+//
+//   - GitHub: GitHub App (JWT / installation transport)
+//   - GitLab: Personal access token via PRIVATE-TOKEN header
+//   - Azure DevOps: PAT as Basic auth
+//   - Bitbucket Cloud: Token as Bearer
+//   - Forgejo/Gitea: Personal access token via Authorization: token header or Basic auth
+//   - Fake: No authentication
```
crenshaw-dev:
```suggestion
// and applies the appropriate authentication based on the provider type.
```

Too much dupe detail that will drift from implementations.

docs/commit-status-controllers/web-request.md
```diff
@@ -375,6 +375,60 @@ data:
...
+metadata:
+  name: github-required-statuses
+spec:
+  promotionStrategyRef:
+    name: my-promotion-strategy
+  key: github-required-statuses
+  reportOn: proposed
+  descriptionTemplate: "GitHub required statuses: {{ .ReportedSha }}"
+  httpRequest:
+    urlTemplate: "https://api.github.com/repos/my-org/my-repo/commits/{{ .ReportedSha }}/check-runs?filter=latest"
```
crenshaw-dev, line 405: When using SCM auth, can we validate that the URL we're hitting is the URL of the selected (Cluster)ScmProvider?

I think our tenancy model is:

* ScmProvider: "bring your own token," i.e. you already have it
* ClusterScmProvider: "you're allowed to use this token"

Even so, there's a difference between making it easy to use a token and allowing the user to yeet the plaintext credential at an arbitrary domain.

If we add this limitation, we should document it both here and in the tenancy docs.

docs/commit-status-controllers/web-request.md
```diff
@@ -375,6 +375,60 @@ data:
...
+
+Set `authentication.scmAuth: {}` and the controller will use the credentials from the ScmProvider referenced by the PromotionStrategy's repository. The authentication method is applied automatically based on the SCM provider type (GitHub App, GitLab token, Azure DevOps PAT, etc.).
+
+| SCM Provider   | Authentication method              | Applied as                          |
+|----------------|------------------------------------|-------------------------------------|
+| GitHub         | GitHub App (JWT / installation)    | Custom HTTP client (installation transport) |
+| GitLab         | Access token                       | `PRIVATE-TOKEN` header               |
+| Azure DevOps   | Personal Access Token (PAT)        | Basic auth (empty username, PAT as password) |
+| Bitbucket Cloud| Repository token                   | Bearer token header                 |
+| Forgejo / Gitea| Token or basic auth                | `Authorization: token <token>` or Basic auth (username/password) |
```
crenshaw-dev: Excessive detail, this info already exists in the ScmProvider docs.

internal/scms/github/git_operations.go
```diff
@@ -199,3 +200,25 @@ func GetClient(ctx context.Context, scmProvider v1alpha1.GenericScmProvider, sec
...
+// that need to call the GitHub API with the same credentials as the rest of the promoter.
+// gitRepo is used to resolve the installation ID from the repo owner when InstallationID is not set on the ScmProvider.
+func GetHTTPTransport(ctx context.Context, scmProvider v1alpha1.GenericScmProvider, secret v1.Secret, gitRepo *v1alpha1.GitRepository) (http.RoundTripper, error) {
+	org := ""
+	if scmProvider.GetSpec().GitHub.InstallationID == 0 && gitRepo != nil && gitRepo.Spec.GitHub != nil {
+		org = gitRepo.Spec.GitHub.Owner
+	}
+	if scmProvider.GetSpec().GitHub.InstallationID == 0 && org == "" {
+		return nil, errors.New("gitRepo (owner) is required when GitHub InstallationID is not set")
+	}
```
crenshaw-dev: Seems like there's a lot of unnecessary conditionals here. `gitRepo` should never be nil (we should probably just short-circuit and return an error if it is). and `gitRepo.Spec.GitHub.Owner is a required field. If we're hitting this function, I think it's safe to assume that `Spec.GitHub` is non-nil, so it's also safe to assume `Spec.GitHub.Owner` is non-nil.

internal/controller/webrequestcommitstatus_controller.go
```diff
@@ -698,27 +698,36 @@ func (r *WebRequestCommitStatusReconciler) makeHTTPRequest(ctx context.Context,
...
+// applyAuthentication configures the request (or client) with the auth from spec: Basic, Bearer, OAuth2, TLS, or ScmAuth.
+// For Basic/Bearer/OAuth2 it mutates the request and returns nil. For TLS or ScmAuth (GitHub) it builds and returns
+// a custom http.Client. Credentials are read from the referenced Secrets or from the PromotionStrategy's SCM provider (ScmAuth).
+func (r *WebRequestCommitStatusReconciler) applyAuthentication(ctx context.Context, wrcs *promoterv1alpha1.WebRequestCommitStatus, ps *promoterv1alpha1.PromotionStrategy, req *http.Request) (*http.Client, error) {
 	auth := wrcs.Spec.HTTPRequest.Authentication
 
 	if auth.Basic != nil {
+		if auth.Basic.SecretRef == nil {
+			return nil, errors.New("authentication.basic.secretRef is required")
+		}
```
crenshaw-dev: I have mixed feelings about nil-checking fields that are already required by CRD validation. Might make the code harder to read, because the reader will be led to believe that the field actually can be nil.

docs/commit-status-controllers/web-request.md
zachaller: I like it from a security perspective maybe add it in a follow up PR?

internal/utils/httpauth/httpauth.go
```diff
@@ -405,3 +408,106 @@ func getSecret(ctx context.Context, reader client.Reader, namespace, name string
...
+// Authorization: token, or Basic) and (nil, nil) is returned.
+func ApplySCMAuth(ctx context.Context, scmProvider promoterv1alpha1.GenericScmProvider, secret *corev1.Secret, req *http.Request, gitRepo *promoterv1alpha1.GitRepository) (*http.Client, error) {
+	logger := log.FromContext(ctx)
+	spec := scmProvider.GetSpec()
+
+	switch {
+	case spec.GitHub != nil:
+		if secret == nil {
+			return nil, errors.New("secret required for GitHub SCM auth")
+		}
```
crenshaw-dev: Maybe do the nil checks in the provider-specifig GetHTTPTransport function? That keeps business logic out of this function (which is basically a simple toggle) and close to the code that requires the secret.

internal/utils/httpauth/httpauth.go
```diff
@@ -405,3 +408,106 @@ func getSecret(ctx context.Context, reader client.Reader, namespace, name string
...
+		transport, err := github.GetHTTPTransport(ctx, scmProvider, *secret, gitRepo)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build GitHub auth transport: %w", err)
+		}
+		logger.V(4).Info("Applied SCM authentication", "provider", "GitHub", "scmProvider", scmProvider.GetName())
+		return &http.Client{Transport: transport}, nil
+
+	case spec.GitLab != nil:
```
crenshaw-dev, line 442: The business logic for GitHub is pulled into a github-specific package. Do the same for the rest of the providers to keep parallel implimenetations?

internal/utils/httpauth/httpauth.go
```diff
@@ -405,3 +408,106 @@ func getSecret(ctx context.Context, reader client.Reader, namespace, name string
...
+			return nil, nil
+		}
+		username, userErr := GetSecretValue(secret, UsernameKey)
+		password, passErr := GetSecretValue(secret, PasswordKey)
+		if userErr == nil && passErr == nil && username != "" && password != "" {
+			credentials := base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
+			req.Header.Set("Authorization", "Basic "+credentials)
+			logger.V(4).Info("Applied SCM authentication (Basic)", "scmProvider", scmProvider.GetName())
+			return nil, nil
+		}
```
crenshaw-dev: We're silently swallowing some errors here. Should probably explicitly throw.

docs/commit-status-controllers/web-request.md
crenshaw-dev: Ehhh, would rather not, just in case we discover dragons while implementing the security check. It's also easy to forget.

docs/commit-status-controllers/web-request.md
```diff
@@ -375,6 +375,53 @@ data:
...
+metadata:
+  name: github-required-statuses
+spec:
+  promotionStrategyRef:
+    name: my-promotion-strategy
+  key: github-required-statuses
+  reportOn: proposed
+  descriptionTemplate: "GitHub required statuses: {{ .ReportedSha }}"
+  httpRequest:
+    urlTemplate: "https://api.github.com/repos/my-org/my-repo/commits/{{ .ReportedSha }}/check-runs?filter=latest"
```
crenshaw-dev, line 405: Small thing, but this example looks slightly off. Doesn't it usually have a `v3` in the path?

docs/commit-status-controllers/web-request.md
```diff
@@ -375,6 +375,53 @@ data:
...
+      Accept: "application/vnd.github+json"
+      X-GitHub-Api-Version: "2022-11-28"
+    authentication:
+      scmAuth: {}
+  success:
+    when:
+      expression: |
+        Response.StatusCode == 200 &&
+        len(Response.Body.check_runs) > 0 &&
+        all(Response.Body.check_runs, # r, r.status == "completed" && (r.conclusion == "success" || r.conclusion == "skipped"))
```
crenshaw-dev, line 417: This example is for check runs. Should we have another example for the legacy commit statuses?

internal/scms/github/git_operations.go
```diff
@@ -199,3 +200,25 @@ func GetClient(ctx context.Context, scmProvider v1alpha1.GenericScmProvider, sec
...
+// ApplyHTTPAuth returns an http.RoundTripper that authenticates requests using the GitHub App
+// credentials from the ScmProvider and secret. It is used for arbitrary HTTP requests (e.g. WebRequestCommitStatus)
+// that need to call the GitHub API with the same credentials as the rest of the promoter.
+// gitRepo is used to resolve the installation ID from the repo owner when InstallationID is not set on the ScmProvider.
+func ApplyHTTPAuth(ctx context.Context, scmProvider v1alpha1.GenericScmProvider, secret v1.Secret, gitRepo *v1alpha1.GitRepository) (http.RoundTripper, error) {
+	org := ""
+	if scmProvider.GetSpec().GitHub.InstallationID == 0 {
+		if gitRepo == nil {
+			return nil, errors.New("gitRepo is required when GitHub InstallationID is not set")
+		}
```
crenshaw-dev: nil gitRepo should be impossible

docs/commit-status-controllers/web-request.md
zachaller: no, here is anothe rworking example:

https://api.github.com/repos/{{ index (splitList "/" (trimPrefix
      "https://github.com" .Environment.PullRequest.Url)) 1 }}/{{ index
      (splitList "/" (trimPrefix "https://github.com"
      .Environment.PullRequest.Url)) 2 }}/pulls/{{ .Environment.PullRequest.ID
      }}

docs/commit-status-controllers/web-request.md
zachaller: I went with no, because actions check use check runs, and the promoter uses check runs

crenshaw-dev (approved):

Issue #1165, opened 2026-03-07
Feature: Reflect SCM's native checks into CommitStatus CRs


Angivare:
## Summary

Introduce a new controller that watches SCM for commit statuses and check runs, and reflects them as CommitStatus CRs in the cluster. This enables native SCM checks (e.g. GitHub Actions workflow runs) to participate in PromotionStrategy gating without requiring users to build external bridges or duplicate statuses.

## Motivation

See existing discussion https://github.com/argoproj-labs/gitops-promoter/discussions/422

### Problem

The existing controller that syncs SCM commit statuses and CommitStatus CRs is unidirectional: it takes CommitStatus CRs and pushes them to the SCM. There is no built-in way to go the other direction.
This creates a gap for teams that want to use/already have CI pipelines (e.g. GitHub Actions) producing check runs or commit statuses on their SCM. To use those as promotion gates, users have two choices:

- Build a custom webhook receiver or CI step that creates CommitStatus CRs in the cluster.
  - This requires end-users to maintain extra glue logic.
  - This will cause the existing controller to push those statuses back to the SCM, possibly duplicating the commit status or producing errors because of ownership conflicts.
- Use required checks on the promotion PRs directly, bypassing CommitStatus CRs entirely.
  - This makes the promotion state opaque from the cluster's perspective — inspecting CommitStatus resources gives no visibility into why a PR isn't being merged.

## Goals

- Reflect SCM-native commit statuses and check runs into CommitStatus CRs.
- Allow reflected statuses to participate in PromotionStrategy promotion gating.
- Avoid writing reflected statuses back to the SCM (preventing duplication and permission errors).
- Support filtering which SCM statuses/checks get reflected (not everything is relevant for promotion).

## Implementation proposal

- Support a `promoter.argoproj.io/reconcile` label on `CommitStatus` that, when set to `"false"`, will prevent the controller from pushing back the status to the SCM
  - By itself, this step should be enough to allow end-users to write their own sync logic without the controller messing up their SCM state
- Add a `ScmCommitStatus` CRD and controller that will fetch commit statuses from SCMs into `CommitStatus` CRs
- Listen on proper webhook events to trigger `ScmCommitStatus` reconciliation

### Changes to CommitStatus controller

CommitStatus CRs with `promoter.argoproj.io/reconcile=false` label must not be reconciled.

### ScmCommitStatus CRD

```yaml
apiVersion: promoter.argoproj.io/v1alpha1
[...]

PR #1167, opened 2026-03-08
feat: Allow CS label to prevent SCM reconciliation

Angivare:
Support a label that prevents CommitStatus to be pushed back into SCM

First step for https://github.com/argoproj-labs/gitops-promoter/issues/1165 proposal

### DIFF
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -4,6 +4,9 @@ package v1alpha1
 // PromotionStrategy CR
 const CommitStatusLabel = "promoter.argoproj.io/commit-status"
 
+// ReconcileLabel is the label used to identify whether a resource should be reconciled by its controller
+const ReconcileLabel = "promoter.argoproj.io/reconcile"
+
 // PreviousEnvProposedCommitPrefixNameLabel is the prefix name for copied proposed commits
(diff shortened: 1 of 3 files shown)


crenshaw-dev:
Maybe this should just be a spec field instead of a label. @zachaller?

zachaller:
@crenshaw-dev maybe, I see you already commented on the authors original issue and reason for this PR. As for this PR it is an interesting idea we have talked about a TriggerEventing kind although it would probably be quite similar to the WebRequestCommitStatus just without the creation of the CommitStatus, I am not sure adding a skip field for the CS today without a real use case makes sense. If we do decide to say let the webrequestcommitstatus manager to disable commitstatus creation it might make sense to still create the CommitStatus kind but disable the Scm call instead which case this PR would be nice but agree I would probably do it as a spec.

crenshaw-dev:
@Angivare is this still something you actively need, or are you pursuing an alternative design with WebRequestCommitStatus?

crenshaw-dev:
Closing for now, can revisit later.

Angivare:
No this is not needed anymore - we're going to move forward with WebRequestCommitStatus. Might work later on a PR to enable triggering the refresh with a webhook, but other than that we're good

PR #1177, opened 2026-03-09
chore: improve security scorecard

crenshaw-dev:
Turned Cursor loose on our OpenSSF Security Scorecard ouptut.


.github/workflows/bump-docs-manifests.yml
```diff
@@ -9,11 +9,13 @@ on:
         type: string
 
 permissions:
-  contents: write  # for creating PRs and pushing changes
```
emirot, line 12: Don't we need write, as we update versions in multiple files?

crenshaw-dev: Yeah, the scorecard check just wants it scoped to the specific job (which in this case is the only job, so it's kind of a silly validation).

PR #1184, opened 2026-03-11
fix: use ssa for status to avoid conflicts

zachaller:


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -664,3 +664,26 @@ func calculateApplicationPhase(app *argocd.Application) promoterv1alpha1.CommitS
 	}
 	return promoterv1alpha1.CommitPhasePending
 }
+
+func (r *ArgoCDCommitStatusReconciler) buildStatusApplyConfiguration(v *promoterv1alpha1.ArgoCDCommitStatus) any {
```
crenshaw-dev, line 669: Could this just be added as a public method on the `ArgoCDCommitStatus` type so that it satisfies an interface? I think that would let us avoid explicitly passing a function to `HandleReconciliationResult`.

Either way, I don't think this should be a receiver, since `r` is unused.

api/v1alpha1/argocdcommitstatus_types.go
```diff
@@ -92,6 +92,10 @@ type ArgoCDCommitStatusStatus struct {
 	// This field is sorted by environment (same order as the referenced PromotionStrategy), then namespace, then name.
 	ApplicationsSelected []ApplicationsSelected `json:"applicationsSelected,omitempty"`
 
+	// ObservedGeneration is the generation of the resource that was last reconciled.
+	// +kubebuilder:validation:Optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
crenshaw-dev, line 97: If this can be unset, shouldn't it be a pointer? I think `0` is a valid generation number.

But also, shouldn't it be required? Even if everything else goes wrong during reconciliation, we should be able to get the observed generation.

internal/utils/status_apply.go
```diff
@@ -0,0 +1,158 @@
...
+			WithLastTransitionTime(c.LastTransitionTime).
+			WithReason(c.Reason).
+			WithMessage(c.Message)
+		result = append(result, ac)
+	}
+	return result
+}
+
+// CommitShaStateToApply converts a CommitShaState to its apply configuration.
+func CommitShaStateToApply(s promoterv1alpha1.CommitShaState) *acv1alpha1.CommitShaStateApplyConfiguration {
```
crenshaw-dev: Do these CRD-specific functions really belong in a one-off utils package? Why not in their respective _types.go files?

zachaller: sure they could go there I tend to not like putting a bunch of stuff in the types file that not on the actuall type but we could

internal/controller/argocdcommitstatus_controller.go
zachaller: I tried sooo hard to turn this into an interface we get cyclic imports

internal/utils/status_apply.go
crenshaw-dev: I don't like business logic in _types, but this seems more type related. Either way, seems better in some type-specific packages instead of utils.

api/v1alpha1/argocdcommitstatus_types.go
zachaller: Upstream uses int64 https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/apps/types.go#L514

api/v1alpha1/timedcommitstatus_types.go
```diff
@@ -79,7 +83,7 @@ type TimedCommitStatusEnvironmentsStatus struct {
 	// Supports both SHA-1 (40 chars) and SHA-256 (64 chars) Git hash formats.
 	// +required
 	// +kubebuilder:validation:MaxLength=64
-	// +kubebuilder:validation:Pattern=`^([a-f0-9]{40}|[a-f0-9]{64})$`
+	// +kubebuilder:validation:Pattern=`^([a-f0-9]{40}|[a-f0-9]{64})?$`
 	Sha string `json:"sha"`
```
crenshaw-dev, line 87: Switch to optional and add omitempty?

api/v1alpha1/gitcommitstatus_types.go
```diff
@@ -135,9 +139,9 @@ type GitCommitStatusEnvironmentStatus struct {
 	// May be empty if the PromotionStrategy hasn't reconciled yet.
 	// Supports both SHA-1 (40 chars) and SHA-256 (64 chars) Git hash formats.
 	// +required
-	// +kubebuilder:validation:MinLength=40
+	// +kubebuilder:validation:MinLength=0
 	// +kubebuilder:validation:MaxLength=64
-	// +kubebuilder:validation:Pattern=`^([a-f0-9]{40}|[a-f0-9]{64})$`
+	// +kubebuilder:validation:Pattern=`^([a-f0-9]{40}|[a-f0-9]{64})?$`
 	ProposedHydratedSha string `json:"proposedHydratedSha"`
```
crenshaw-dev, line 145: Make it optional and remove the redundant MinLength=0 annotation?

crenshaw-dev: Also add omitempty

PR #1185, opened 2026-03-11
feat(ui): allow label to enable conditional rendering of argo-cd extension 

jwinters01:
* Add prettier config to UI directories
* App level annotation `promoter.argoproj.io/promotion-strategy` can now be used to indicate to the UI extension where to find the promotion strategy
* Add unit tests to test the conditional rendering logic
* Switch to Vitest (a more modern test runner)

### DIFF
--- a/docs/argocd-integrations/index.md
+++ b/docs/argocd-integrations/index.md
@@ -100,7 +100,7 @@ The following table shows the compatibility between GitOps Promoter versions and
 The extension tab appears on an Argo CD Application when either of the following is true:
 
 1. **Auto-detection (default):** The extension checks the Application's top-level resource tree for `PromotionStrategy` resources. If at least one is found, the tab is shown.
-2. **Label-based:** The Application has the label `promoter.argoproj.io/has-promotionstrategy` set to any truthy value (e.g. `"true"`). Use this when the PromotionStrategy is not top-level, i.e. it is a child of another resource.
+2. **Label-based:** The Application has the label `promoter.argoproj.io/has-promotionstrategy` set to `"true"`. Use this when the PromotionStrategy is not top-level, i.e. it is a child of another resource.
 
 Use the label if your `PromotionStrategy` resources are nested or if you want explicit control over which Applications display the extension tab:
 


ui/extension/hasOnePromotionStrategy.ts
```diff
@@ -0,0 +1,9 @@
+import type { Application } from "@shared/types/extension";
+
+export const ANNOTATION = "promoter.argoproj.io/promotion-strategy";
+
+export const hasOnePromotionStrategy = (application: Application): boolean => {
```
crenshaw-dev: Rename something more generic like `showExtension`?

ui/components-lib/package.json
```diff
@@ -6,7 +6,9 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "build": "tsc",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --noEmit",
+    "format": "prettier --write .",
+    "format:check": "prettier --check ."
```
crenshaw-dev, line 11: Should we add this to the CI checks?

ui/extension/AppViewExtension.tsx
```diff
@@ -3,50 +3,72 @@ import Card from "@components-lib/components/Card";
...
 
+const ANNOTATION = "promoter.argoproj.io/promotion-strategy";
+const GROUP = "promoter.argoproj.io";
+const KIND = "PromotionStrategy";
+
 const AppViewExtension = ({ tree, application }: AppViewComponentProps) => {
   const promotionStrategyNodes = useMemo(
-    () => tree.nodes.filter((node) => node.kind === "PromotionStrategy"),
-    [tree.nodes]
+    () => tree.nodes.filter((node) => node.kind === KIND),
```
crenshaw-dev: Just noticed that we should probably be checking the group too.

ui/extension/AppViewExtension.tsx
```diff
@@ -3,50 +3,72 @@ import Card from "@components-lib/components/Card";
...
-  const [isLoading, setIsLoading] = useState(false);
+  const [promotionStrategy, setPromotionStrategy] = useState<PromotionStrategy>();
+  const [fetchError, setFetchError] = useState<string | null>(null);
+
+  const annotationValue = application.metadata.annotations?.[ANNOTATION];
+
+  const fetchParams = useMemo(() => {
+    if (annotationValue) {
+      const [namespace, name] = annotationValue.split("/");
+      return { name, namespace, version: "v1alpha1" };
```
crenshaw-dev: Instead of putting this info on the annotation, what if we just had the annotation indicate that the app contains at least one PromotionStrategy? We could then pull the PromotionStrategy info from the resource tree, like we are today.

This wouldn't solve the problem of showing multiple PromotionStrategies in a single app, but I think that's a problem we can solve in a separate PR (i.e. have a dropdown to select the PS to view).

jwinters01: I tend to prefer non-boolean data when I can, imo adding the actual resource communicates the same thing + more, but if we expect to be able to handle multiple promotion strategies in one app at some point, I see your point 👍

ui/components-lib/package.json
jwinters01: yeah could be good, Ill update the workflow as well

ui/extension/AppViewExtension.tsx
jwinters01: what do you mean by group?

ui/extension/AppViewExtension.tsx
crenshaw-dev: Yeah definitely want to be able to handle multiple. +1 on not liking bools, but I'm not sure what else to put in that annotation.

ui/extension/AppViewExtension.tsx
crenshaw-dev: `promoter.argoproj.io` - it's super unlikely that we'll have PromotionStrategy resources from another group in an app, but being more specific is just a bit safer.

ui/extension/AppViewExtension.tsx
zachaller: Yea we already have multiple at intuit and we should support it somehow

jwinters01: so I understand we will eventually want to support multiple (maybe in a followup PR), but for the time being, what do we want the experience to be if there are multiple in this PR? Naively I would just choose the first one that gets returned as the one the extension displays, do you all have a preference?

jwinters01:
On adding the prettier config to CI, I needed to run the formatter on the whole repo, so lots of formatting changes were generated. Important files to look at for this PR:
1. `ui/extension/showExtension.ts` — Extracted showExtension function with label-based discovery (promoter.argoproj.io/extension-enabled) plus resource tree fallback (now >= 1 instead of === 1).
  2. `ui/extension/showExtension.test.ts` — Tests for the above.
  3. `ui/extension/index.tsx` — Uses the new showExtension instead of the old inline hasOnePromotionStrategy.
  4. `ui/extension/AppViewExtension.tsx` — Filters nodes by group + kind, handles >= 1 nodes (picks first), adds fallback URL when no node is in the tree, replaces isLoading with fetchError error handling.
  5. `ui/shared/src/types/extension.ts` — Adds labels? and annotations? to the Application metadata type.
  6. `ui/extension/vitest.config.ts` — Vitest config for the new test suite.
  7. `ui/extension/package.json` — Adds vitest, prettier, eslint-config-prettier deps and test script.
  8. `ui/extension/tsconfig.json` — Expands include to cover test files and AppViewExtension.tsx.
  9. `ui/extension/eslint.config.js` — Adds eslint-config-prettier integration.
  10. `Makefile` — Adds format:check to lint targets.

ui/extension/showExtension.ts
```diff
@@ -0,0 +1,11 @@
+import type { Application } from '@shared/types/extension';
+
+export const LABEL = 'promoter.argoproj.io/extension-enabled';
```
crenshaw-dev: How about `promoter.argoproj.io/has-promotionstrategy`? I think that's a bit nicer for folks who are filtering Argo CD apps by label in the Argo UI.

crenshaw-dev: Oh, and can you add a section to the Argo CD UI Extension docs to describe the label and when to use it?

docs/argocd-integrations/index.md
```diff
@@ -95,6 +95,23 @@ The following table shows the compatibility between GitOps Promoter versions and
 | v0.18.3+                | v2.3+           |
 | v0.18.2 and earlier     | v2.2.x and earlier |
 
+### Extension Discovery
+
+The extension tab appears on an Argo CD Application when either of the following is true:
+
+1. **Label-based (recommended):** The Application has the label `promoter.argoproj.io/has-promotionstrategy` set to any truthy value (e.g. `"true"`).
+2. **Auto-detection (default):** The extension checks the Application's top-level resource tree for `PromotionStrategy` resources. If at least one is found, the tab is shown.
```
crenshaw-dev:
```suggestion
1. **Auto-detection (default):** The extension checks the Application's top-level resource tree for `PromotionStrategy` resources. If at least one is found, the tab is shown.
2. **Label-based:** The Application has the label `promoter.argoproj.io/has-promotionstrategy` set to any truthy value (e.g. `"true"`). Use this when the PromotionStrategy is not top-level, i.e. it is a child of another resource.
```

ui/extension/showExtension.ts
```diff
@@ -0,0 +1,11 @@
+import type { Application } from '@shared/types/extension';
+
+export const LABEL = 'promoter.argoproj.io/has-promotionstrategy';
+
+export const showExtension = (application: Application): boolean => {
+  if (application.metadata.labels?.[LABEL]) return true;
```
crenshaw-dev: I think we should actually just go ahead and be super opinionated, i.e. the value must be lowercase `true`. As it is, `false` would return true.

docs/argocd-integrations/index.md
```diff
@@ -95,6 +95,23 @@ The following table shows the compatibility between GitOps Promoter versions and
 | v0.18.3+                | v2.3+           |
 | v0.18.2 and earlier     | v2.2.x and earlier |
 
+### Extension Discovery
+
+The extension tab appears on an Argo CD Application when either of the following is true:
+
+1. **Auto-detection (default):** The extension checks the Application's top-level resource tree for `PromotionStrategy` resources. If at least one is found, the tab is shown.
+2. **Label-based:** The Application has the label `promoter.argoproj.io/has-promotionstrategy` set to any truthy value (e.g. `"true"`). Use this when the PromotionStrategy is not top-level, i.e. it is a child of another resource.
```
crenshaw-dev:
```suggestion
2. **Label-based:** The Application has the label `promoter.argoproj.io/has-promotionstrategy` set to `"true"`. Use this when the PromotionStrategy is not top-level, i.e. it is a child of another resource.
```

PR #1193, opened 2026-03-13
chore(kubebuilder): bump version

emirot:
## What

Bumping kubebuilder version

### DIFF
--- a/.github/workflows/kubebuilder_helm.yaml
+++ b/.github/workflows/kubebuilder_helm.yaml
@@ -50,6 +50,7 @@ jobs:
 
       - name: Lint Helm chart
         run: |
+          # we set these to true to match the defaults from the Helm chart's values.yaml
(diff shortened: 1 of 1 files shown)


.github/workflows/kubebuilder_helm.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       - name: Lint Helm chart
         run: |
-          helm lint --debug ./tmp_chart/chart
+          helm lint --debug ./tmp_chart/chart  --set crd.enable=true --set webhook.enable=true
```
crenshaw-dev, line 54:
```suggestion
          # we set these to true to match the defaults from the Helm chart's values.yaml
          helm lint --debug ./tmp_chart/chart  --set crd.enable=true --set webhook.enable=true
```

emirot:
@crenshaw-dev thanks! good idea to add the comment

PR #1195, opened 2026-03-13
feat add metrics check

emirot:


zachaller:
I feel this should be a golang test not?

emirot:
> I feel this should be a golang test not?

This is testing our install, meaning our kustomize files that's why it s not in golang.
It made me realize that we don't have a service account in our config that will make our metrics work for prometheus so are we expecting users to create it ?

Issue #804, opened 2025-12-17
argocd ui extension not loading after 0.18.3


seanl-circle:
I had the ui extension installed in argocd-server as described in https://argo-cd.readthedocs.io/en/stable/developer-guide/extensions/ui-extensions/#ui-extensions

In 0.18.2 it works, and there's a promotionstrategy tab on promotionstrategy resources
In 0.18.3 / 0.19.0 it just doesn't show up at all. I couldn't find any errors or logs from argocd-server or in the browser console logs.

zachaller:
I think we just moved it, you can click the little branching like icon: 

<img width="324" height="160" alt="Image" src="https://github.com/user-attachments/assets/a6ba0993-3480-4f35-99b9-2b0a2bb08082" />

seanl-circle:
Hmm, that might be a bit of a problem for us.
We have repos that can can hold either a single or multiple services, and each may have one or more instances deployed independently.

Our plan was to use the an appset with an scm generator to map 1 repo to 1 root argocd application, which would for each service instance in that repo, template out a promotionstrategy + appset with a cluster generator.
So each root argocd application can hold multiple promotion strategies

Moving promotion strategy ui there it looks like it only supports one promotionstrategy per app?
The application screen crashes with:

```
Something went wrong!

Consider submitting an issue [here](https://github.com/argoproj/argo-cd/issues/new?labels=bug&template=bug_report.md).


Stacktrace:

Error: Expected exactly one PromotionStrategy resource in Gitops Promoter Extension, found 2
    at window.extensionsAPI.registerAppViewExtension.t (https://argocd-staging.example.com/extensions.js:2:42499)
    at Qo (https://argocd-staging.example.com/main.b8f3b898a503e41b9e16.js:2:563114)
    at Es (https://argocd-staging.example.com/main.b8f3b898a503e41b9e16.js:2:609072)
    at kl (https://argocd-staging.example.com/main.b8f3b898a503e41b9e16.js:2:601856)
    at xl (https://argocd-staging.example.com/main.b8f3b898a503e41b9e16.js:2:601781)
    at hl (https://argocd-staging.example.com/main.b8f3b898a503e41b9e16.js:2:598814)
    at https://argocd-staging.example.com/main.b8f3b898a503e41b9e16.js:2:550534
    at t.unstable_runWithPriority (https://argocd-staging.example.com/main.b8f3b898a503e41b9e16.js:2:625639)
[...]

PR #1213, opened 2026-03-18
feat(ui): support multiple promotionstrategies in ui extension

seanl-circle:
For the argocd ui extension, this adds a dropdown menu in the style of the namespace selector in the dashboard ui.
The promotion strategy is also selectable via the url query parameter `promotionstrategy=<name>`.

For #804

### DIFF
--- a/ui/extension/AppViewExtension.tsx
+++ b/ui/extension/AppViewExtension.tsx
@@ -41,7 +41,22 @@ const AppViewExtension = ({ application }: AppViewComponentProps) => {
 
     setFetchError(null);
     fetch(url)
-      .then((response) => response.json())
+      .then(async (response) => {
+        if (!response.ok) {
+          let errorText = '';
+          try {
+            errorText = await response.text();
+          } catch {
+            // ignore errors while reading error body
+          }
+          const messageParts = [
+            `Request failed with status ${response.status} ${response.statusText}`,
+            errorText && `body: ${errorText}`,
+          ].filter(Boolean);
+          throw new Error(messageParts.join(' - '));
+        }
+        return response.json();
+      })
       .then((data) => {
         if (!data.items || data.items.length === 0) {
           throw new Error('No PromotionStrategy resources found');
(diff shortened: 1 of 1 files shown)


ui/extension/StrategyDropdown.scss
```diff
@@ -0,0 +1,92 @@
...
+  padding: 10px 12px;
+}
+
+.strategy-dropdown__option--is-selected {
+  background: #d1eaff;
+  color: #0d8ecf;
+}
+
+.strategy-dropdown__control {
+  width: 400px !important;
```
crenshaw-dev, line 21: This looks really heavy on the `!important`s, are they really necessary?

crenshaw-dev:
> Changes will increase total bundle size by 88.84kB (212.11%)

I guess react-select is kinda bulky.

ui/extension/StrategyDropdown.scss
seanl-circle: Not sure, it's basically a copy of https://github.com/argoproj-labs/gitops-promoter/blob/09a228ff026ddbcc82abdfa7e5cc4ccea947c127/ui/dashboard/src/components/NamespaceDropdown.scss

crenshaw-dev: Oof didn't know that was there. No harm done duplicating then.

jwinters01:
> > Changes will increase total bundle size by 88.84kB (212.11%)
> 
> I guess react-select is kinda bulky.

A good alternative to these sometimes bulky packages is to use [shadcn](https://ui.shadcn.com/docs/components/base/select), works a bit differently in that you "install" these components through an npx script that directly adds the code to your repo, which means you own it, and also means minimal impact on bundle size (aside from the shadcn primitives that are needed for the components)

crenshaw-dev:
https://github.com/user-attachments/assets/6e11704f-0aa8-49e9-8ca3-fda29fabf439

PR #1215, opened 2026-03-18
feat(ui): History view improvements

jwinters01:
* More Prominent Grayed out background in history view
* Separates active checks and Proposed checks snapshots
* Adds descriptive tooltips describing when these snapshots were made

https://github.com/user-attachments/assets/a34d24e3-47e5-44b7-89c6-ae0f30d3fe2a

_note:_ This also includes more enforcement of prettier linting rules (previous integration was missing the shared library), so a bunch of these file changes are just formatting changes 🙏


crenshaw-dev:
Man Copilot is chatty.

PR #1232, opened 2026-03-24
Feat/add automerge and branches for pull request template

emirot:
## What

Improve PR template default configuration

## Test

Screenshot 2026-03-23 at 4 52 56 PM

### DIFF
--- a/internal/utils/template.go
+++ b/internal/utils/template.go
@@ -16,6 +16,15 @@ func init() {
 	delete(sanitizedSprigFuncMap, "expandenv")
 	delete(sanitizedSprigFuncMap, "getHostByName")
 	sanitizedSprigFuncMap["urlQueryEscape"] = url.QueryEscape
+	sanitizedSprigFuncMap["deref"] = derefBool
+}
+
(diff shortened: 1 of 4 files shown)


internal/utils/template.go
```diff
@@ -16,6 +16,15 @@ func init() {
 	delete(sanitizedSprigFuncMap, "expandenv")
 	delete(sanitizedSprigFuncMap, "getHostByName")
 	sanitizedSprigFuncMap["urlQueryEscape"] = url.QueryEscape
+	sanitizedSprigFuncMap["deref"] = derefBool
```
zachaller, line 19: I don't think this is needed maybe `expr.AsBool()`

emirot: @zachaller from what I saw online expr.AsBool() is not compatible with go templates.

crenshaw-dev: Could you just wrap the nilable value in an if block?

emirot: Yes but I will still need the deref right @crenshaw-dev ?

PR #1259, opened 2026-04-02
feat: support promotionstrategy context for WRCS

zachaller:


api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -340,19 +370,69 @@ type HTTPRequestSpec struct {
...
+	// When PhasePerBranch is set, Phase is used as the default for any branch not listed in PhasePerBranch.
+	// +kubebuilder:validation:Enum=pending;success;failure
+	// +required
+	Phase CommitStatusPhase `json:"phase"`
+
+	// PhasePerBranch holds per-branch phases when the success expression returned an object with per-branch overrides.
+	// Key is branch name, value is "pending", "success", or "failure". When set, each environment's CommitStatus
+	// uses this phase; branches not in the map use Phase.
+	// +optional
+	PhasePerBranch map[string]CommitStatusPhase `json:"phasePerBranch,omitempty"`
```
crenshaw-dev: List with branch key instead? Just in case one way we wanna do something silly like have per-env urls or descriptions.

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -340,19 +370,69 @@ type HTTPRequestSpec struct {
...
+
+	// LastRequestTime is when the last HTTP request was made.
+	// +optional
+	LastRequestTime *metav1.Time `json:"lastRequestTime,omitempty"`
+
+	// LastResponseStatusCode is the HTTP status code from the last request.
+	// +optional
+	LastResponseStatusCode *int `json:"lastResponseStatusCode,omitempty"`
```
crenshaw-dev, line 410: Have a `response` block with children `StatusCode` and `Output`?

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -340,19 +370,69 @@ type HTTPRequestSpec struct {
...
+
+	// TriggerOutput stores the map returned by spec.mode.trigger.when.output.expression.
+	// +optional
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:validation:Type=object
+	// +kubebuilder:pruning:PreserveUnknownFields
+	TriggerOutput *apiextensionsv1.JSON `json:"triggerOutput,omitempty"`
```
crenshaw-dev: Any reason to nest `trigger.output`?

docs/commit-status-controllers/web-request.md
```diff
@@ -8,15 +8,17 @@ The WebRequestCommitStatus controller provides flexible validation by calling ex
 
 ### How It Works
 
-For each environment configured via the PromotionStrategy:
+Behavior depends on **`spec.mode.context`** (see [Request scope](#request-scope-environments-vs-promotionstrategy) below). By default (`environments`), the controller runs **one HTTP request per applicable environment**. With `promotionstrategy`, it runs **one HTTP request per reconcile** and maps the result to every applicable environment’s `CommitStatus`.
```
crenshaw-dev: A couple places talk about "one HTTP request per reconcile" but that's not quite accurate. It's one request per WebRequestCommitStatus. A given reconcile may or may not trigger a request.

docs/commit-status-controllers/web-request.md
```diff
@@ -56,6 +58,142 @@ Uses expressions to dynamically control when HTTP requests are made. Powerful fo
...
+| Value | HTTP requests per reconcile (when the trigger allows) | `CommitStatus` resources |
+|-------|------------------------------------------------------|---------------------------|
+| `environments` (default) | One per applicable environment | One per environment; each request uses that environment’s templates and SHA |
+| `promotionstrategy` | **One** for the whole `WebRequestCommitStatus` | Still one per environment, but all are updated from the **same** response; each row uses that environment’s `reportOn` SHA |
+
+The controller reconciles when the `WebRequestCommitStatus` or the referenced **`PromotionStrategy`** changes (for example environment SHAs moving), so promotion-strategy context stays in sync with the strategy.
+
+### When to use `promotionstrategy` context
+
+Use it when a **single** external API call represents validation for the whole pipeline (or a subset of environments that share one backend), and you do not want N identical HTTP calls for N environments. Examples:
```
crenshaw-dev:
```suggestion
Use it when a **single** external API call represents validation for the whole PromotionStrategy (or a subset of environments that share one backend), and you do not want N identical HTTP calls for N environments. Examples:
```

pipeline is a bad word

internal/controller/webrequestcommitstatus_controller.go
```diff
@@ -237,132 +249,62 @@ func (r *WebRequestCommitStatusReconciler) SetupWithManager(ctx context.Context,
...
-	for i := range previousStatus.Environments {
-		wrcsEnvStatusMap[previousStatus.Environments[i].Branch] = &previousStatus.Environments[i]
-	}
-
-	// Build a map of environments from PromotionStrategy for efficient lookup
-	psEnvStatusMap := make(map[string]*promoterv1alpha1.EnvironmentStatus, len(ps.Status.Environments))
-	for i := range ps.Status.Environments {
-		psEnvStatusMap[ps.Status.Environments[i].Branch] = &ps.Status.Environments[i]
+	// Clear the promotionstrategy-context status; this path uses per-environment status instead.
+	wrcs.Status.PromotionStrategyContext = nil
```
crenshaw-dev, line 257: If we're gonna allow switching contexts (i.e. context is a mutable field), we should probably have a test to confirm that it works as expected.

internal/controller/webrequestcommitstatus_expressions.go
```diff
@@ -192,34 +184,135 @@ func (r *WebRequestCommitStatusReconciler) evaluateValidationExpression(ctx cont
...
+func parsePhaseString(phaseStr string, defaultPhase promoterv1alpha1.CommitStatusPhase) promoterv1alpha1.CommitStatusPhase {
+	switch phaseStr {
+	case "success":
+		return promoterv1alpha1.CommitPhaseSuccess
+	case "pending":
+		return promoterv1alpha1.CommitPhasePending
+	case "failure":
+		return promoterv1alpha1.CommitPhaseFailure
+	default:
+		return defaultPhase
```
crenshaw-dev: Log the invalid phase provided?

crenshaw-dev (changes requested):

internal/controller/webrequestcommitstatus_expressions.go
```diff
@@ -192,34 +184,135 @@ func (r *WebRequestCommitStatusReconciler) evaluateValidationExpression(ctx cont
...
+		}
+		if branch == "" {
+			return promoterv1alpha1.CommitPhasePending, nil, fmt.Errorf("validation expression environments[%d]: branch is required", i)
+		}
+		phaseStr, err := getString(entry, "phase")
+		if err != nil {
+			return promoterv1alpha1.CommitPhasePending, nil, fmt.Errorf("validation expression environments[%d].phase: %w", i, err)
+		}
+		m[branch] = parsePhaseString(phaseStr, defaultPhase)
+	}
```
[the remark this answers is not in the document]
zachaller: this is a bit more strict now and a bunch of validation is in parsePerBranchPhases

api/v1alpha1/webrequestcommitstatus_types.go
zachaller: would have to update the other mode as well this keeps it consistent with that mode

api/v1alpha1/webrequestcommitstatus_types.go
zachaller: consitent with other type WebRequestCommitStatusEnvironmentStatus

api/v1alpha1/webrequestcommitstatus_types.go
zachaller: consistent with : WebRequestCommitStatusEnvironmentStatus

api/v1alpha1/webrequestcommitstatus_types.go
zachaller: they are already templated from the webreq

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -132,6 +142,17 @@ type ModeSpec struct {
 	// The controller will evaluate the expression to determine when to make HTTP requests.
 	// +optional
 	Trigger *TriggerModeSpec `json:"trigger,omitempty"`
+
+	// Context controls whether the controller makes one HTTP request per environment or one per PromotionStrategy.
+	// - "environments" (default): one HTTP request per environment; each environment gets its own phase and status.
+	// - "promotionstrategy": at most one HTTP request per WebRequestCommitStatus; the same phase is applied to CommitStatuses for all
+	//   environments, each still reporting on that environment's reportOn SHA. When context is promotionstrategy,
+	//   per-environment template variables (Environment, ReportedSha, LastSuccessfulSha) are not set; do not use
+	//   them in URL, body, description, or trigger templates.
```
crenshaw-dev: > the same phase is applied to CommitStatuses for all environments

That's not exactly true, is it? Can't the user specify per-environment statuses if they want?

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -146,16 +167,25 @@ type PollingModeSpec struct {
 
 // SuccessSpec defines when the HTTP response is considered successful (commit status phase success).
 type SuccessSpec struct {
-	// When holds the boolean expression evaluated against the HTTP response (expr library).
+	// When holds the expression evaluated against the HTTP response (expr library).
 	// Available variables: Response.StatusCode (int), Response.Body (parsed JSON as map[string]any, or raw string if not JSON), Response.Headers (map[string][]string).
-	// The expression must return true for validation to pass. Example: "Response.StatusCode == 200 && Response.Body.approved == true"
+	//
+	// Return type:
+	// - When mode.context is "environments": must return a boolean. true sets phase Success, false sets Pending.
```
crenshaw-dev: Not necessary for this PR, but it's odd that the `promotionstrategy` mode can set any status (i.e. failure), but `environments` cannot.

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -146,16 +167,25 @@ type PollingModeSpec struct {
...
+	// - When mode.context is "promotionstrategy": may return:
+	//   - A boolean: same phase (Success or Pending) for all environments.
+	//   - An object with defaultPhase (optional) and environments (optional array):
+	//     - defaultPhase: "success", "pending", or "failure" — defaults to "pending" when omitted. Used for all when environments is empty, or for branches not listed in environments.
+	//     - environments: optional array of { branch (string), phase (string) }. Example: { defaultPhase: "pending", environments: [{ branch: "dev", phase: "success" }, { branch: "staging", phase: "pending" }] }.
```
crenshaw-dev: Should most of this doc string just be moved to the `Expression` field? Not sure why it's a level up.

api/v1alpha1/webrequestcommitstatus_types.go
crenshaw-dev: Yeah but that ends up being a global value, right? Can't be customized per environment.

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -340,19 +370,69 @@ type HTTPRequestSpec struct {
...
+	// LastRequestTime is when the last HTTP request was made.
+	// +optional
+	LastRequestTime *metav1.Time `json:"lastRequestTime,omitempty"`
+
+	// LastResponseStatusCode is the HTTP status code from the last request.
+	// +optional
+	LastResponseStatusCode *int `json:"lastResponseStatusCode,omitempty"`
+
+	// TriggerOutput stores the map returned by spec.mode.trigger.when.output.expression.
+	// +optional
```
crenshaw-dev: Does it have to be a map? If we dropped `// +kubebuilder:validation:Type=object` it seems like this could be anything that can marshal to JSON.

api/v1alpha1/webrequestcommitstatus_types.go
```diff
@@ -340,19 +370,69 @@ type HTTPRequestSpec struct {
...
+	// +kubebuilder:validation:Type=object
+	// +kubebuilder:pruning:PreserveUnknownFields
+	ResponseOutput *apiextensionsv1.JSON `json:"responseOutput,omitempty"`
+
+	// LastSuccessfulShas tracks the last SHA that achieved success for each branch.
+	// Used with reportOn "proposed" + polling to skip HTTP requests when all environments
+	// have already succeeded for their current SHAs.
+	// Key is branch name, value is the SHA that last succeeded.
+	// +optional
+	LastSuccessfulShas map[string]string `json:"lastSuccessfulShas,omitempty"`
```
crenshaw-dev: I'm allergic to maps, but can't think of any reason we'd add additional fields. It would let us validate the branch and sha fields though.

docs/commit-status-controllers/web-request.md
```diff
@@ -56,6 +58,142 @@ Uses expressions to dynamically control when HTTP requests are made. Powerful fo
...
+- `{{ .ReportedSha }}` / `ReportedSha`  
+- `{{ .LastSuccessfulSha }}` / `LastSuccessfulSha`  
+
+They are unset (empty / nil) because there is no single “current environment” for that one request.
+
+**You can use:** `PromotionStrategy` (full spec and status), `Phase` (previous overall phase in promotion-strategy context), `TriggerOutput`, `ResponseOutput` (trigger mode), and `NamespaceMetadata` labels and annotations — same as in environment context, except for the per-env fields above.
+
+When rendering **CommitStatus** `descriptionTemplate` and `urlTemplate`, the controller sets **`{{ .Phase }}`** to **that environment’s** resolved phase (`success`, `pending`, or `failure`). It still does **not** set `ReportedSha` or `Environment` in promotion-strategy context. If you need branch- or SHA-specific text in the SCM description or URL, either **walk `{{ .PromotionStrategy }}`** in the Go template (for example, `range` over `.PromotionStrategy.Status.Environments` and match on `.Branch`) or use **the same wording for every environment** (a generic message or link that does not try to substitute `{{ .ReportedSha }}` or `{{ .Environment.Branch }}`).
```
crenshaw-dev, line 93: If it's not already there, we should document this on the `urlTemplate` and `descriptionTemplate` fields.

api/v1alpha1/webrequestcommitstatus_types.go
crenshaw-dev: But not persisted on the CR status?

docs/commit-status-controllers/web-request.md
```diff
@@ -56,6 +58,142 @@ Uses expressions to dynamically control when HTTP requests are made. Powerful fo
...
+### Status shape (`promotionstrategy` context)
+
+Observed state is stored under **`status.promotionStrategyContext`**, not under **`status.environments`** (that slice is used only for `environments` context).
+
+Notable fields:
+
+- **`phase`** — default phase for branches not overridden in `phasePerBranch`
+- **`phasePerBranch`** — map branch → phase when the success expression returned the object form
+- **`lastSuccessfulShas`** — branch → last SHA that reached **success** for that branch (used for optimizations below)
+- **`lastRequestTime`**, **`lastResponseStatusCode`**, **`triggerOutput`**, **`responseOutput`** — same roles as per-environment status in the default context, but stored once for the shared request
```
crenshaw-dev, line 120: Don't forget to add new stuff to the WebRequestCommitStatus example CR.

internal/controller/webrequestcommitstatus_controller.go
```diff
@@ -237,132 +249,62 @@ func (r *WebRequestCommitStatusReconciler) SetupWithManager(ctx context.Context,
...
+
+	// Snapshot status from the last reconcile before we rebuild it. Used for transition detection,
+	// polling skip optimization, and carrying forward state when the trigger doesn't fire.
+	lastReconciledStatus := wrcs.Status.DeepCopy()
+	if lastReconciledStatus == nil {
+		lastReconciledStatus = &promoterv1alpha1.WebRequestCommitStatusStatus{}
+	}
+	wrcsEnvStatusMap := make(map[string]*promoterv1alpha1.WebRequestCommitStatusEnvironmentStatus, len(lastReconciledStatus.Environments))
```
crenshaw-dev: Maybe rename var `statusByEnv`?

internal/controller/webrequestcommitstatus_controller_test.go
```diff
@@ -1889,3 +1889,809 @@ var _ = Describe("WebRequestCommitStatus Controller - SCM Host Validation", func
...
+	Describe("Polling Mode - Boolean Success Expression", func() {
+		var wrcs *promoterv1alpha1.WebRequestCommitStatus
+
+		BeforeEach(func() {
+			By("Creating a test HTTP server that returns approved=true")
+			testServer = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusOK)
+				_ = json.NewEncoder(w).Encode(map[string]any{
+					"approved": true,
```
crenshaw-dev, line 1955: Would be interesting to set up the polling tests where the server initially returns false and then goes to true. You'd want a short polling interval.

internal/controller/webrequestcommitstatus_controller.go
```diff
@@ -371,167 +313,461 @@ func (r *WebRequestCommitStatusReconciler) processEnvironments(ctx context.Conte
...
+	if decision.ShouldFire {
+		return r.handleHTTPRequestAndValidation(ctx, wrcs, td)
+	}
+	return httpValidationResult{
+		Phase:                  phaseOrDefault(lastState.Phase),
+		PhasePerBranch:         lastState.PhasePerBranch,
+		LastRequestTime:        lastState.LastRequestTime,
+		LastResponseStatusCode: lastState.LastResponseStatusCode,
+		ResponseDataJSON:       lastState.ResponseOutput,
+	}, nil
```
crenshaw-dev, line 696: This is only used in two places and is kinda trivial. Maybe just inline it for clarity?

The only real logic here that would be annoying to duplicate is the re-construction of `httpValidationResult` so you get the function call `phaseOrDefault(lastState.Phase)`.

But do we really need that? Is the lastState.Phase not defaulted in the same place that all the other fields are set? Or could it be defaulted where the field is consumed?

crenshaw-dev:
We should probably test the "zero applicable environments" case.

PR #1260, opened 2026-04-02
feat(metrics): resource count gauge

crenshaw-dev:


zachaller:
Instead of calling `client.List()` for every kind (which deep-copies every object just to count them), you can get the count directly from the informer's store — zero allocations, zero copies.

### Why this matters

`client.List()` on the manager's cached client still:
1. Calls into the informer store to get the raw objects
2. **Deep-copies every single object** via `DeepCopyObject()` so the caller can't mutate the cache
3. Appends each deep-copied object into the typed `list.Items` slice

For 1,000 `PullRequest` resources, that's 1,000 full deep copies — just to throw them away after `len(list.Items)`. Repeated every 30 seconds for all 13 kinds.

### Suggested approach: read the informer store length directly

The informers are already running for all these types (the controllers watch them). `cache.GetInformer()` returns a `cache.SharedIndexInformer` under the hood (confirmed in controller-runtime's `internal/informers.go`), which exposes `GetStore().List()` — a `[]interface{}` of pointers to objects **already in memory**. Just `len()` it.

This also eliminates the 130-line switch statement and Copilot's concern about drift between `promoterResourceKinds` and the switch cases — single source of truth.

```go
type promoterResource struct {
	kind string
	obj  client.Object
}

var promoterResources = []promoterResource{
	{"ArgoCDCommitStatus", &promoterv1alpha1.ArgoCDCommitStatus{}},
	{"ChangeTransferPolicy", &promoterv1alpha1.ChangeTransferPolicy{}},
	{"ClusterScmProvider", &promoterv1alpha1.ClusterScmProvider{}},
	{"CommitStatus", &promoterv1alpha1.CommitStatus{}},
	{"ControllerConfiguration", &promoterv1alpha1.ControllerConfiguration{}},
	{"GitCommitStatus", &promoterv1alpha1.GitCommitStatus{}},
	{"GitRepository", &promoterv1alpha1.GitRepository{}},
	{"PromotionStrategy", &promoterv1alpha1.PromotionStrategy{}},
	{"PullRequest", &promoterv1alpha1.PullRequest{}},
	{"RevertCommit", &promoterv1alpha1.RevertCommit{}},
	{"ScmProvider", &promoterv1alpha1.ScmProvider{}},
	{"TimedCommitStatus", &promoterv1alpha1.TimedCommitStatus{}},
	{"WebRequestCommitStatus", &promoterv1alpha1.WebRequestCommitStatus{}},
}

func countFromInformer(ctx context.Context, c crcache.Cache, obj client.Object) (int, error) {
	informer, err := c.GetInformer(ctx, obj)
	if err != nil {
		return 0, fmt.Errorf("getting informer: %w", err)
	}
	si, ok := informer.(toolscache.SharedIndexInformer)
	if !ok {
		return 0, fmt.Errorf("informer does not implement SharedIndexInformer")
	}
	return len(si.GetStore().List()), nil
}

func refreshKubernetesResourceCounts(ctx context.Context, c crcache.Cache, log logr.Logger) {
	for _, r := range promoterResources {
		n, err := countFromInformer(ctx, c, r.obj)
		if err != nil {
			log.Error(err, "counting resources for promoter_kubernetes_resources metric", "kind", r.kind)
			kubernetesResources.WithLabelValues(r.kind).Set(0)
			continue
		}
		kubernetesResources.WithLabelValues(r.kind).Set(float64(n))
	}
}
```

Call site changes from `localManager.GetClient()` to `localManager.GetCache()`:

```go
if err := localManager.Add(metrics.NewResourceCountRunnable(localManager.GetCache())); err != nil {
```

The type assertion to `SharedIndexInformer` is safe — controller-runtime's internal `Cache` struct stores that exact type (`Informer cache.SharedIndexInformer` on line 110 of `internal/informers.go` in v0.23.3).

PR #1303, opened 2026-04-07
chore(ci): upgrade to newer goreleaser format

crenshaw-dev:
In response to this log line from goreleaser:

>     • dockers and docker_manifests are being phased out and will eventually be replaced by dockers_v2, check https://goreleaser.com/deprecations#dockers for more info


crenshaw-dev:
Will worry about this later

PR #1319, opened 2026-04-10
fix: no-op dry SHA bypasses health checks on previous environments

zachaller:
Fixes a bug where a newer no-op dry commit (one that doesn't change manifests for any environment) caused isPreviousEnvironmentPending to skip health checks on all preceding environments, allowing staging and production to promote before earlier environments were healthy.
Adds active commit status health verification before recursing past no-op environments in the previous-environment check.

## Problem
When a new dry commit arrives that is a no-op for all environments (only git notes updated, no new hydrated commits), every environment satisfies the "no-op with no pending changes" condition:

Note.DrySha (newer) != Proposed.Dry.Sha (older) → classified as no-op
Active.Dry.Sha == Proposed.Dry.Sha (real changes already merged) → no pending changes
The function recursed past every environment without checking any Active.CommitStatuses, hit the base case (empty list), and returned "allow promotion." This caused:

Staging to promote before dev was verified healthy
Production to promote while staging's argocd-health was still pending (0 healthy, 0 degraded, 2 pending)

## Fix
Before recursing past a no-op environment, verify its active commit statuses are passing. An environment can be a no-op for the newest dry SHA, but its health from the last real deployment still matters for gating downstream promotions.


crenshaw-dev:
I think one way to look at this change is that we don't currently have a way to determine whether a previous environment's unhealthy state is relevant or not to the promotion of the current environment, so we err on the safe side.

Example scenario (A' is a no-op):

```
env/dev: A (healthy) -> A' (unhealthy)
env/stg: A (healthy) -> A' (healthy)
env/prd: A (healthy) -> B (blocked by env/dev)
```

In this case, it would be intuitive to say that, since A went healthy in env/dev, there's no reason to believe its current un-healthiness is related to change B which only affected env/prd.

But since we don't look at historical data, and since for the purpose of determining whether a change is promotion-worthy A' is identical to A, we block promotion of B to env/prd in case there was something wrong with A.

Does that make sense?

zachaller:
Yea, another way of saying it and I guess this gets a bit tricky as well is do previous environments need to be healthy (now including no-ops) no matter what the sha is in order to promote, but that could also be seen as not ideal behavior as well?

crenshaw-dev:
Agreed, I think it's maybe too aggressive. But for the sake of safety, I think we should start there. We can add safe optimizations over time.

zachaller:
One possible option is also adding a time element, today we check health if `envMergedTarget := envStatus.Active.Dry.Sha == targetDrySha`: https://github.com/zachaller/gitops-promoter/blob/3dbddfce97641f1bfc3407f6ff915860b2f983b8/internal/controller/promotionstrategy_controller.go#L746-L758 

When we check for the no-op case we could add a check and only do it if say the actually dry sha is older than x time. A pro of moving the previous commit status to a new controller is that would could make that time more configurable in the new spec.

crenshaw-dev (approved):

PR #1320, opened 2026-04-10
fix(ctp): clear CTP finalizer from PRs before CTP is deleted

crenshaw-dev:
The CTP needs to make sure its finalizers are removed from all owned PRs so that the PRs can be cleaned up.

controller-runtime uses background deletion by default which doesn't block deletion of the CTP on the owned PRs, even though block owner deletion is true. So we use a finalizer to ensure cleanup.


api/v1alpha1/constants.go
```diff
@@ -36,6 +36,11 @@ const PullRequestFinalizer = "pullrequest.promoter.argoproj.io/finalizer"
 // ChangeTransferPolicyPullRequestFinalizer prevents deletion of PullRequest until ChangeTransferPolicy copies its status
 const ChangeTransferPolicyPullRequestFinalizer = "changetransferpolicy.promoter.argoproj.io/pullrequest-finalizer"
 
+// ChangeTransferPolicyPullRequestCleanupFinalizer is set on ChangeTransferPolicy while PullRequests may carry
+// ChangeTransferPolicyPullRequestFinalizer. On ChangeTransferPolicy deletion, the controller removes the PR finalizers
+// and clears this finalizer so the policy can finish deleting without waiting for per-PR status propagation.
+const ChangeTransferPolicyPullRequestCleanupFinalizer = "changetransferpolicy.promoter.argoproj.io/pullrequest-finalizer-cleanup"
```
zachaller: I don't think we should get into having multiple finalizer labels per controller kind, I think we should just call this changetransferpolicy.promoter.argoproj.io/finalizer?

crenshaw-dev: Good call, changing

internal/controller/changetransferpolicy_controller.go
```diff
@@ -871,6 +870,49 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+
+	for i := range prList.Items {
+		prItem := &prList.Items[i]
+
+		// Remove the CTP's PR finalizer so the PR can be deleted.
+		if controllerutil.ContainsFinalizer(prItem, promoterv1alpha1.ChangeTransferPolicyPullRequestFinalizer) {
+			logger.Info("Removing ChangeTransferPolicy PullRequest finalizer during ChangeTransferPolicy deletion",
+				"pullRequest", prItem.Name)
+			controllerutil.RemoveFinalizer(prItem, promoterv1alpha1.ChangeTransferPolicyPullRequestFinalizer)
+			if err := r.Update(ctx, prItem); err != nil {
```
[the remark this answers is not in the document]
zachaller: This should probably be a SSA? I think the standard for controller modifing other resoruces today is SSA

api/v1alpha1/constants.go
crenshaw-dev: Changed, added docs

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1207,6 +1249,49 @@ func (r *ChangeTransferPolicyReconciler) gitMergeStrategyOurs(ctx context.Contex
...
+	}
+
+	// Remove CTP finalizers from any PRs. The CTP is being deleted, we don't care about monitoring those PRs anymore.
+	err := r.handleCTPCleanupOnDelete(ctx, ctp)
+	if err != nil {
+		return fmt.Errorf("failed to clean up PullRequest finalizers for deleted ChangeTransferPolicy: %w", err)
+	}
+
+	controllerutil.RemoveFinalizer(ctp, finalizer)
+	if err := r.Update(ctx, ctp); err != nil {
```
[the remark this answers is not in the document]
zachaller: This should use retryonconflict

internal/controller/changetransferpolicy_controller.go
```diff
@@ -115,6 +116,18 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
...
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to handle finalizer: %w", err)
+	}
+
+	// If the CTP is being deleted, stop here. handleFinalizer has already performed cleanup
+	// (removing the CTP's PR finalizer from owned PullRequests). We must not continue and
+	// re-add those finalizers via creatOrUpdatePullRequest or mergePullRequests.
+	if !ctp.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	}
```
zachaller: In other controllers I am pretty sure this is in handleFinalizer, which does make you need to return a deleted var like in PR controller, this is generally only nice when you are actually deleting some external resource that might need to know about being deleted not sure if we want to keep the pattern the same or not, all our current handleFinalizer do it the same way?

crenshaw-dev: Yeah I thought about following the existing pattern... there's basically no flow that escapes `handleFinalizer` that should permit further processing, so a boolean seemed redundant. This way just read more cleanly to me.

zachaller: Yea if we are going to make the change, do we want to do it for all controllers then, that don't need to know about cleanup logic returned? I think consistency wins either way? I personally like keeping that logic in the function because it seems a bit more finalizer related than reconciliation related but meh as long as we do the same thing in places I think it's fine.

crenshaw-dev: Yeah I'd like it more in the function but if the only path is `return early`, I wanna avoid the boolean.... will make sure it's consistent

crenshaw-dev: I went ahead and changed it to match the others. I don't like that `err` and `deleted` effectively communicate the same thing, but meh.

crenshaw-dev:
@zachaller I feel like copilot js just wrong about how server side apply will handle the unset finalizers field?

zachaller:
@copilot resolve the merge conflicts in this pull request

crenshaw-dev:
@copilot resolve the merge conflicts in this pull request

zachaller:
@copilot resolve the merge conflicts in this pull request

zachaller (approved):

PR #1345, opened 2026-04-17
fix: use SSA to update status to avoid conflicts

zachaller:


docs/contributing/updating-status.md
```diff
@@ -0,0 +1,99 @@
...
+2. During reconciliation, the controller mutates `obj.Status` in memory. **Do not call
+   `r.Status().Update` or `r.Status().Patch` yourself.**
+3. At the end of the reconcile, the deferred helper sets the `Ready` condition based on
+   whether `*err` is nil, stamps `status.observedGeneration = metadata.generation`, then
+   applies the whole status subresource via **Server-Side Apply** under the
+   per-controller `FieldOwner` with `ForceOwnership`.
+4. If the full apply is rejected (for example, OpenAPI schema or CEL validation on some
+   status field), the helper retries with a conditions-only SSA so the `Ready=False`
+   condition describing the failure still reaches the user. The next successful reconcile
+   naturally re-owns all fields.
```
crenshaw-dev: Does setting only conditions mean it'll wipe the rest of the status field? Seems maybe excessive and would make the UI experience problematic. Should we instead just restore everything that was already set in the status and just know that the failed condition means the status is stale?

zachaller: hm... yea I like that we even have an observered genreation in the condition we can update with the new generation but we can keep the top level one old

zachaller: done

internal/utils/utils.go
```diff
@@ -322,6 +337,14 @@ func HandleReconciliationResult(
...
 	conditions := obj.GetConditions() // GetConditions() is guaranteed to be non-nil for our CRDs.
 	readyCondition := meta.FindStatusCondition(*conditions, string(promoterConditions.Ready))
+	// Only this helper writes Reason=ReconciliationError. Finding one here means it's a stale
+	// leftover from a previous failed reconcile; discard it so a successful reconcile can
+	// install a fresh success condition below. Without this, SSA would keep re-applying the
+	// stale condition forever because the deferred helper reads conditions back from the
+	// in-memory object (which still carries whatever was persisted by the previous reconcile).
+	if readyCondition != nil && readyCondition.Reason == string(promoterConditions.ReconciliationError) {
+		readyCondition = nil
+	}
```
crenshaw-dev: Isn't this made unnecessary by the `Remove` utility called at the beginning of each controller's Reconcile?

zachaller: It is, I dind't mind keeping it but also don't mind removing it

mkdocs.yml
```diff
@@ -37,6 +37,7 @@ nav:
       - Continuous Integration: contributing/ci.md
       - Cutting a Release: contributing/cutting-a-release.md
       - Using Finalizers: contributing/using-finalizers.md
+      - Updating Status: contributing/updating-status.md
```
crenshaw-dev: Consider renaming to "maintaining resource status"

crenshaw-dev (approved):
🥳

PR #1365, opened 2026-04-24
feat: Simulator for WebRequestCommitStatus

### LINKED ISSUE #1371: Improve expression cache to avoid memory leak
The `Evaluator` component caches all compiled expressions and doesn't do any garbage collection.

We should switch to something like a LRU or LFU cache to avoid memory leaking.

zachaller:
also fixes: https://github.com/argoproj-labs/gitops-promoter/issues/1371


internal/webrequest/wrcs_process_types.go
```diff
@@ -0,0 +1,80 @@
...
+// (per-environment context and promotionstrategy context).
+type ProcessWebRequestCommitStatusInput struct {
+	Evaluator              *Evaluator
+	HttpExec               HTTPEXecutor
+	WebRequestCommitStatus *promoterv1alpha1.WebRequestCommitStatus
+	PromotionStrategy      *promoterv1alpha1.PromotionStrategy
+	// NamespaceMeta is passed into TemplateData for template rendering.
+	NamespaceMeta         NamespaceMetadata
+	CommitEmitter         CommitStatusEmitter
+	RenderedHTTPCollector RenderedHTTPCollector // optional; when non-nil, CollectRenderedHTTP is called after a successful template render when the trigger fires
```
crenshaw-dev: Why do we need this in addition to the HTTPExecutor? Couldn't the executor collect the data?

internal/webrequest/evaluator.go
crenshaw-dev: Might be worth renaming to ExpressionEvaluator to differentiate from other components.

zachaller: even with it under the webrequest package?

webrequestsimulator/internal/simulate/simulate.go
```diff
@@ -0,0 +1,195 @@
...
+	*c.out = append(*c.out, simulatortypes.RenderedRequest{
+		Branch: r.Branch, Method: r.Method, URL: r.URL, Headers: r.Headers, Body: r.Body,
+	})
+}
+
+// Simulate runs one WebRequestCommitStatus reconcile against args, using
+// args.HTTPResponses in place of any real HTTP call. The returned Result.Status
+// matches what the controller would write to WebRequestCommitStatus.Status.
+//
+// Safe for concurrent use: a fresh Evaluator is created per call.
```
crenshaw-dev: Isn't Evaluator already thread safe?

internal/webrequest/wrcs_process_types.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"context"
+
+	promoterv1alpha1 "github.com/argoproj-labs/gitops-promoter/api/v1alpha1"
+)
+
+// ProcessWebRequestCommitStatusInput carries shared dependencies for WRCS processing
+// (per-environment context and promotionstrategy context).
+type ProcessWebRequestCommitStatusInput struct {
+	Evaluator              *Evaluator
```
crenshaw-dev: We can simplify this struct by making the Evaluator cache a package singleton. It already is effectively a singleton, because we initialize it only once for the whole controller.

internal/webrequest/evaluator.go
crenshaw-dev: Yeah, webrequest has a lot of stuff going on now. Even in that context "Evaluator" doesn't mean much.

crenshaw-dev (approved):

PR #1393, opened 2026-05-04
feat(ui): persist strategy dropdown selection

jwinters01:
Enable deep-linking from a GitHub PR check's "Details" button to a specific
`PromotionStrategy` in the ArgoCD UI extension when an Application owns more
than one strategy.

- **Controller:** add `.PromotionStrategy` to `URLTemplateData` so the
  `ArgoCDCommitStatus` `url.template` can encode which strategy owns a status
  (e.g. `?promotionstrategy={{ .PromotionStrategy.Namespace }}/{{ .PromotionStrategy.Name }}`).
  Docs and a render test updated.
- **UI extension:** persist the selected strategy in `localStorage` (scoped
  per-app) as a fallback under the URL query param, since ArgoCD strips the
  query string when navigating between application views. Resolution order
  is URL → localStorage → first strategy; the URL remains the authoritative
  deep-link target.

### DIFF
--- a/ui/extension/AppViewExtension.test.tsx
+++ b/ui/extension/AppViewExtension.test.tsx
@@ -189,7 +189,7 @@ describe('AppViewExtension', () => {
         .dispatchEvent(new MouseEvent('mousedown', { bubbles: true, cancelable: true }));
       await wait();
 
-      const options = Array.from(container.querySelectorAll('.strategy-dropdown__option'));
+      const options = Array.from(document.querySelectorAll('.strategy-dropdown__option'));
       const optionTexts = options.map((o) => o.textContent);
       expect(optionTexts).toContain('strategy-1');
       expect(optionTexts).toContain('strategy-2');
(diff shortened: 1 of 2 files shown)


docs/commit-status-controllers/argocd.md
```diff
@@ -70,6 +70,7 @@ The following variables are available in the template:
 
 - `.Environment` - string holding the environment name (i.e. environment branch name) for the group of Applications the URL is being generated for.
 - `.ArgoCDCommitStatus` - holds the whole [CR](../crd-specs.md#argocdcommitstatus) in its current state
+- `.PromotionStrategy` - holds the [PromotionStrategy CR](../crd-specs.md#promotionstrategy) that owns this commit status. Useful for deep-linking the Argo CD UI extension to a specific strategy, for example `?promotionstrategy={{ .PromotionStrategy.Namespace }}/{{ .PromotionStrategy.Name }}`.
```
crenshaw-dev: I think adding `PromotionStrategy` as a variable would be fine, but I don't think we actually need it for this feature.

The `spec.promotionStrategyRef.name` field gets you the name, and the PromotionStrategy is always in the same namespace as the ArgoCDCommitStatus, so you can get the namespace from `metadata.namespace`.

zachaller (approved):

PR #1396, opened 2026-05-06
fix(ctp): old field manager preserving stale dry sha

crenshaw-dev:
When we switched from client-side to server-side apply, we left the old field manager.

For ChangeTransferPolicies (and PromotionStrategies) used with GitOps Promoter < 0.27.0, the `Note` field is owned by the old field manager, while the `Note.DrySha` field is owned by the new field manager.

This means that, when the controller processes a new hydrated commit that doesn't yet have a git note, it'll leave the `note` field unmodified, and the old note will persist.

Having an old `note.drySha` can allow the PromotionStrategy controller to get confused about what commit is actually up for promotion. A new commit can be promoted before it's actually validated, because the controller thinks that it's actually an older, already-validated commit.

This change is the smallest possible change I could find to fix the bug. By making the `note` field atomic, the new field manager will take ownership even if the field isn't yet populated.

Two longer-term fixes we should consider:

1) Adding logic to completely clean up the legacy field manager
2) Making more (all?) fields in the statuses atomic, since we currently use only one controller per status field


crenshaw-dev:
The correct tests failed before the fix:

> 2026-05-06T14:10:18.1581479Z [38;5;9m[1mSummarizing 2 Failures:[0m
> 2026-05-06T14:10:18.1584784Z   [38;5;9m[FAIL][0m [0mSSA migration bug e2e (CTP controller produces stale Note.DrySha after legacy Update) [38;5;9m[1m[It] after a new dry SHA push, Note.DrySha must NOT remain stuck at the legacy Update's old value[0m
> 2026-05-06T14:10:18.1588201Z   [38;5;243m/home/runner/work/gitops-promoter/gitops-promoter/internal/controller/changetransferpolicy_controller_test.go:1050[0m
> 2026-05-06T14:10:18.1592484Z   [38;5;9m[FAIL][0m [0mSSA migration bug e2e (PromotionStrategy gates downstream env on stale Note.DrySha) [38;5;9m[1m[It] when a new dry SHA arrives on a CTP carrying legacy gitops-promoter Update ownership, the previous-environment gate must NOT fire success[0m
> 2026-05-06T14:10:18.1595470Z   [38;5;243m/home/runner/work/gitops-promoter/gitops-promoter/internal/controller/promotionstrategy_controller_test.go:5144[0m

PR #1408, opened 2026-05-07
Fix activePath issues: branch naming, CommitStatus scoping, FindOpen filter, PR titles

patjlm:
## Summary

Fixes found while testing the `activePath` monorepo mode from #1337 with a [5-component × 6-environment example repo](https://github.com/patjlm/gitops-promoter-example). Details in [discussion #1385](https://github.com/argoproj-labs/gitops-promoter/discussions/1385).

### 1. Git ref conflict: proposed branch names conflict with active branch names (Blocker)

The current naming generates proposed branches as `<active-branch>/<activePath>-next`, e.g. `environment/development/apps/app-a-next`. Git cannot create this ref because `environment/development` already exists as a branch ref.

**Fix** (`766d3e0`): Append activePath to the existing `-next` convention — `environment/development-next/apps/app-a`.

### 2. Shared commit status keys collide in activePath mode (Blocker)

When multiple PromotionStrategies share the same active branch, they share the same commit SHA. If they use the same commit status keys (e.g. `ci-check`), the CTP controller finds multiple CommitStatus CRs with the same `(sha, key)` and errors with "too many matching SHAs". This blocks promotion for all components.

**Fix** (`59014dc`, `404a8ff`): Add an `ActivePathLabel` to CommitStatus CRs and filter on it in `setCommitStatusState`. The label must be set everywhere CommitStatus CRs are created: WebRequestCommitStatus controller, TimedCommitStatus controller, and `createOrUpdatePreviousEnvironmentCommitStatus` in the PromotionStrategy controller. Without the label on `promoter-previous-environment` CRs, promotion beyond the first environment is blocked.

### 3. PR title and description should include activePath (Enhancement)

With shared active branches, all PRs are titled `Promote <sha> to environment/development` — indistinguishable.

**Fix** (`5e13fa3`): Update the default PR title template to conditionally include activePath.

### 4. FindOpen returns wrong PR when multiple PRs target the same base branch (Medium)

Not activePath-specific, but exposed by it. `FindOpen` in the GitHub SCM provider passes `pullRequest.Spec.SourceBranch` directly as the `Head` filter to GitHub's List Pull Requests API. GitHub requires `owner:branch` format — without the `owner:` prefix, GitHub silently ignores the filter and returns all open PRs for the base branch. `FindOpen` then picks `pullRequests[0]` and the subsequent `Update` overwrites the wrong PR's title.

**Fix** (`23a6558`): Prefix `Head` with `gitRepo.Spec.GitHub.Owner`.

## Test plan

- [x] Tested with [gitops-promoter-example](https://github.com/patjlm/gitops-promoter-example): 5 components × 6 environments, shared active branches, same commit status keys
- [x] Full promotion pipeline verified: development → integration → stage (with 2min soak timer) → prod (manual merge)
- [x] Single-component change (app-b only) promotes independently without affecting other components
- [x] PRs created with correct titles and source branches (no cross-contamination)
- [x] CommitStatus CRs don't collide across components sharing the same active branch
- [ ] Existing unit tests pass

### DIFF
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -22,6 +22,10 @@ const TimedCommitStatusLabel = "promoter.argoproj.io/timed-commit-status"
 // WebRequestCommitStatusLabel the web request commit status which the commit status is associated with.
 const WebRequestCommitStatusLabel = "promoter.argoproj.io/web-request-commit-status"
 
+// ActivePathLabel scopes CommitStatus CRs to a specific activePath when multiple
+// PromotionStrategies share the same environment branches.
+const ActivePathLabel = "promoter.argoproj.io/active-path"
+
 // PreviousEnvironmentCommitStatusKey the commit status key name used to indicate the previous environment health
 const PreviousEnvironmentCommitStatusKey = "promoter-previous-environment"
 
--- a/internal/scms/github/pullrequest.go
+++ b/internal/scms/github/pullrequest.go
@@ -203,7 +203,7 @@ func (pr *PullRequest) FindOpen(ctx context.Context, pullRequest v1alpha1.PullRe
 	pullRequests, response, err := pr.client.PullRequests.List(
 		ctx, gitRepo.Spec.GitHub.Owner,
 		gitRepo.Spec.GitHub.Name,
-		&github.PullRequestListOptions{Base: pullRequest.Spec.TargetBranch, Head: pullRequest.Spec.SourceBranch, State: "open"})
+		&github.PullRequestListOptions{Base: pullRequest.Spec.TargetBranch, Head: fmt.Sprintf("%s:%s", gitRepo.Spec.GitHub.Owner, pullRequest.Spec.SourceBranch), State: "open"})
 	if response != nil {
 		metrics.RecordSCMCall(ctx, gitRepo, metrics.SCMAPIPullRequest, metrics.SCMOperationList, response.StatusCode, time.Since(start), getRateLimitMetrics(response.Rate))
 	}
--- a/config/config/controllerconfiguration.yaml
+++ b/config/config/controllerconfiguration.yaml
@@ -47,8 +47,8 @@ spec:
               slowDelay: "5m"
               maxFastAttempts: 3
     template:
-      title: "Promote {{ trunc 5 .ChangeTransferPolicy.Status.Proposed.Dry.Sha }} to `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}`"
-      description: "This PR is promoting the environment branch `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}` which is currently on dry sha {{ .ChangeTransferPolicy.Status.Active.Dry.Sha }} to dry sha {{ .ChangeTransferPolicy.Status.Proposed.Dry.Sha }}."
+      title: "Promote {{ with .ChangeTransferPolicy.Spec.ActivePath }}`{{ . }}` {{ end }}({{ trunc 5 .ChangeTransferPolicy.Status.Proposed.Dry.Sha }}) to `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}`"
+      description: "This PR is promoting {{ with .ChangeTransferPolicy.Spec.ActivePath }}`{{ . }}` on {{ end }}the environment branch `{{ .ChangeTransferPolicy.Spec.ActiveBranch }}` from dry sha {{ .ChangeTransferPolicy.Status.Active.Dry.Sha }} to {{ .ChangeTransferPolicy.Status.Proposed.Dry.Sha }}."
   commitStatus:
     workQueue:
       maxConcurrentReconciles: 10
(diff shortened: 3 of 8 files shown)


internal/scms/github/pullrequest.go
```diff
@@ -203,7 +203,7 @@ func (pr *PullRequest) FindOpen(ctx context.Context, pullRequest v1alpha1.PullRe
 	pullRequests, response, err := pr.client.PullRequests.List(
 		ctx, gitRepo.Spec.GitHub.Owner,
 		gitRepo.Spec.GitHub.Name,
-		&github.PullRequestListOptions{Base: pullRequest.Spec.TargetBranch, Head: pullRequest.Spec.SourceBranch, State: "open"})
+		&github.PullRequestListOptions{Base: pullRequest.Spec.TargetBranch, Head: fmt.Sprintf("%s:%s", gitRepo.Spec.GitHub.Owner, pullRequest.Spec.SourceBranch), State: "open"})
```
crenshaw-dev, line 206: This is super interesting, it was previously loading all PRs and not just the ones within the same fork?

crenshaw-dev: I wonder if the problem applies to other SCMs too.

api/v1alpha1/constants.go
```diff
@@ -22,6 +22,10 @@ const TimedCommitStatusLabel = "promoter.argoproj.io/timed-commit-status"
 // WebRequestCommitStatusLabel the web request commit status which the commit status is associated with.
 const WebRequestCommitStatusLabel = "promoter.argoproj.io/web-request-commit-status"
 
+// ActivePathLabel scopes CommitStatus CRs to a specific activePath when multiple
+// PromotionStrategies share the same environment branches.
+const ActivePathLabel = "promoter.argoproj.io/active-path"
```
crenshaw-dev, line 27: I'm hesitant about this... it basically makes `key` + label the composite key, which means "key" is no longer really the "key". What were the exact problems with appending the active path to the existing key field?

Something we'll have to deal with either way is allowed character constraints. We're pretty limited in the allowed character set, and we need to make sure that any normalization we apply doesn't allow key collisions (which could allow premature promotion).

internal/scms/github/pullrequest.go
crenshaw-dev: Would you mind opening this as an independent fix? Nice easy fix. :-)

patjlm: Sure: https://github.com/argoproj-labs/gitops-promoter/pull/1417

I included some reference docs I found on the topic:
- [GitHub REST API — List pull requests](https://docs.github.com/en/rest/pulls/pulls#list-pull-requests): *"Filter pulls by head user or head organization and branch name in the format of `user:ref-name` or `organization:ref-name`."*
- [GitHub Community Discussion #24354](https://github.com/orgs/community/discussions/24354): confirms the silent-ignore behavior when the owner prefix is omitted.

api/v1alpha1/constants.go
patjlm: The base PR #1337 had no `activePath` scoping in `CommitStatus` lookups — the collision is a gap I found during testing. There are two approaches to fix it (that I can think of):

Option A: Append `activePath` to the key value (your suggestion)
- The `CommitStatus` label becomes `commit-status=ci-check-apps-app-a`
- Pros: simpler, no new label, key remains the sole lookup dimension
- Cons: creates a mismatch between the user-facing key (`key: ci-check` in `PromotionStrategy`/`WebRequestCommitStatus` specs) and the internal label value (`ci-check-apps-app-a`). Also, `KubeSafeLabel` truncates from the beginning (preserves the suffix) — so a long combo could drop the key prefix (`ci-check` / `deploy`), which is the part that distinguishes different gates. The `activePath` suffix would survive, but two different keys with the same `activePath` could collide

Option B: Separate label (my current implementation)
- Adds `active-path=apps-app-a` as an independent label
- Pros: key stays exactly as defined in the spec — no mismatch. `Key` and `activePath` are independently constrained (each under 63 chars, no combined truncation risk)
- Cons: adds a new label to the API surface, makes the lookup composite

Both approaches require the same controllers (`WebRequestCommitStatus`, `TimedCommitStatus`, `PromotionStrategy`) to be `activePath`-aware when creating `CommitStatus` CRs — the scope of changes is the same either way.

I went with option B because the key/spec mismatch and truncation behavior concerned me, but happy to rework if you prefer option A.

crenshaw-dev:
Note to self/us: need to make sure the `activePath` implementation sufficiently guards against path traversal

api/v1alpha1/constants.go
crenshaw-dev: I discussed with @zachaller, and we both ended up on Option C: leave it to the user to disambiguate. It's a bit more work for them to understand what's happening and set the values. But it's less magic and allows more direct control.

ArgoCDCommitStatus and TimedCommitStatus don't yet support setting a `spec.key`, so I've opened a PR to add that: https://github.com/argoproj-labs/gitops-promoter/pull/1446

crenshaw-dev:
@patjlm with https://github.com/argoproj-labs/gitops-promoter/pull/1446 merged, we can push this forward! Could you open a new PR against `main` with all the changes together? I'll close the Copilot PR and we can just work off your new PR.

patjlm:
@crenshaw-dev  I opened https://github.com/argoproj-labs/gitops-promoter/pull/1486

Still performing some tests, but the initial ones look good. Closing this

PR #1426, opened 2026-05-13
test: fix flaky CommitStatus spec/status read test

crenshaw-dev:


internal/controller/changetransferpolicy_controller_test.go
```diff
@@ -544,6 +547,21 @@ var _ = Describe("ChangeTransferPolicy Controller", func() {
 				Expect(commitStatus.Spec.Phase).To(Equal(promoterv1alpha1.CommitPhaseSuccess), "spec should be success")
 				Expect(commitStatus.Status.Phase).To(Equal(promoterv1alpha1.CommitPhasePending), "status should be pending")
 
+				By("Bumping a CTP annotation to trigger reconcile")
```
zachaller: We should probably use: CTPEnqueueFunc

zachaller (approved):

PR #1446, opened 2026-05-21
feat(acdcs/tcs): add optional spec.key field

crenshaw-dev:
GitCommitStatus and WebRequestCommitStatus both support setting a custom `key` per resource. ArgoCDCommitStatus and TimedCommitStatus both hard-code a static key.

I'm adding `spec.key` to ArgoCDCommitStatus and TimedCommitStatus to 1) match the other two resources and 2) support multiple *CommitStatus resources targeting the same active branch when we add `activePath` support. This solves point 2 in https://github.com/argoproj-labs/gitops-promoter/pull/1408 by leaving it to the user to make sure they disambiguate *CommitStatus resources by key when they share a single active branch.

For backwards-compatibility, the new key fields are optional and default to the current value. I've added docs describing the plan to make them required so that they match the other CRs.


internal/controller/argocdcommitstatus_controller.go
```diff
@@ -460,16 +469,8 @@ func lookupArgoCDCommitStatusFromArgoCDApplication(mgr mcmanager.Manager) mchand
...
-			// fetch the ArgoCDApplication from the cluster
-			if err := cl.GetClient().Get(ctx, client.ObjectKeyFromObject(argoCDApplication), application, &client.GetOptions{}); err != nil {
-				if k8s_errors.IsNotFound(err) {
-					return nil
-				}
-				logger.Error(err, "failed to get ArgoCDApplication")
-				return nil
-			}
+			// Use labels from the watch event object so Application deletions still enqueue reconciliation.
+			appLabels := labels.Set(argoCDApplication.GetLabels())
```
crenshaw-dev, line 474: It threw in this optimization for free. Not sure why we didn't use this in the first place.

zachaller:
Follow-up to the panic concern on `CommitStatusResourceName` / `CommitStatusGateLabelKeyForParent`: rather than `panic` on empty `TypeMeta.Kind` (or making callers remember to set `ResourceTypeMeta(...)`), you can resolve the Kind from the runtime scheme via `apiutil.GVKForObject`, which is already used in this repo (`internal/metrics/resource_count_test.go`).

```go
import "sigs.k8s.io/controller-runtime/pkg/client/apiutil"

func CommitStatusResourceName(ctx context.Context, scheme *runtime.Scheme, parent client.Object, branch string) (string, error) {
    kind := parent.GetObjectKind().GroupVersionKind().Kind
    if kind == \"\" {
        gvk, err := apiutil.GVKForObject(parent, scheme)
        if err != nil {
            return \"\", fmt.Errorf(\"resolve kind for %T: %w\", parent, err)
        }
        kind = gvk.Kind
    }
    partialKind := strings.ToLower(strings.TrimSuffix(kind, \"CommitStatus\"))
    return KubeSafeUniqueName(ctx, parent.GetName()+\"-\"+branch+\"-\"+partialKind), nil
}
```

Same shape for \`CommitStatusGateLabelKeyForParent\`. Reconcilers already have a scheme via \`r.Client.Scheme()\`, so call sites become \`utils.CommitStatusResourceName(ctx, r.Client.Scheme(), &gcs, branch)\` — no new plumbing. The webrequest simulator constructs its own scheme already, so it can pass that through.

Benefits:
- No \`panic\` in a reconciler hot path.
- The two controller-runtime issues (\`#1870\`, \`#3302\`) about empty \`TypeMeta.Kind\` on typed \`Get\` results stop mattering.
- Test helpers (\`argocdCommitStatusResourceName\`, \`gitCommitStatusResourceName\`, and the suggested \`timedCommitStatusResourceName\`) can drop the \`ResourceTypeMeta(...)\` boilerplate and just pass an \`ObjectMeta\`-only object plus the suite's scheme.
- \`apiutil.GVKForObject\` errors when the type isn't registered or matches multiple GVKs — for the four \`*CommitStatus\` types that's a non-issue, but the error return surfaces real misconfigurations cleanly instead of crashing the pod.

Tradeoff is a signature change (extra \`scheme\` param + \`error\` return) across ~5 call sites, all mechanical. Strictly better than the current panic + mandatory \`TypeMeta\` contract.

internal/controller/argocdcommitstatus_controller.go
crenshaw-dev: Actually this fixed a bug where we missed reconciles when the Application was being deleted. There's supposedly a test that'll prevent regressions, I haven't verified yet.

zachaller (approved):

PR #1447, opened 2026-05-21
feat(acdcs): resolve root-relative URL templates via argocd-cm

jwinters01:
ArgoCDCommitStatus.URL.Template now accepts root-relative paths ("/applications?labels=...") in addition to absolute http(s) URLs. Root-relative output is resolved at render time to an absolute URL, so CommitStatus.spec.url stays SCM-safe (GitHub/GitLab/etc. require http(s) for details_url / target_url).

Base URL resolution precedence:
  1. ArgoCDCommitStatus.spec.argocdBaseURL (new, per-resource override)
  2. argocd-cm.url ConfigMap on the local cluster
  3. None — CommitStatus.spec.url is omitted and a Ready condition is set with Reason=ArgoCDBaseURLUnset so the degraded state is visible via kubectl describe.

Absolute http(s) templates continue to pass through unchanged.

Adds a narrow ConfigMap get permission scoped to resourceNames=argocd-cm.

### DIFF
--- a/api/v1alpha1/argocdcommitstatus_types.go
+++ b/api/v1alpha1/argocdcommitstatus_types.go
@@ -34,8 +34,25 @@ type ArgoCDCommitStatusSpec struct {
 	ApplicationSelector *metav1.LabelSelector `json:"applicationSelector,omitempty"`
 
 	// URL generates the URL to use in the CommitStatus, for example a link to the Argo CD UI.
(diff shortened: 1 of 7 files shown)


crenshaw-dev:
Closing since we found a better way. We're gonna use a CMP to inject the instance domain into the ArgoCDCommitStatus so we can construct the full URL. No need to make it relative.

PR #1468, opened 2026-05-22
fix(ui): NaN days ago (#1467)

### LINKED ISSUE #1467: "NaN days ago" in the UI
Image

crenshaw-dev:
Fixes: #1467 

After:

image

### DIFF
--- a/ui/shared/src/utils/PSData.ts
+++ b/ui/shared/src/utils/PSData.ts
@@ -1,4 +1,4 @@
-import { getCommitUrl, extractNameOnly, extractBodyPreTrailer, formatDate, timeAgo } from './util';
+import { getCommitUrl, extractNameOnly, extractBodyPreTrailer, timeAgo } from './util';
 import { getEnvironmentStatus, getHealthStatus } from './getStatus';
 import type {
   CommitStatus,
@@ -9,6 +9,7 @@ import type {
   EnrichedEnvDetails,
   PromotionPhase,
   ReferenceCommit,
+  RelativeTimeAgo,
 } from '../types/promotion';
 
 function getChecks(commitStatuses: CommitStatus[]): Check[] {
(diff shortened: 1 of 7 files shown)


jwinters01 (approved):

jwinters01:
I am actually confused about where the actual fix here is, taking another look 😅  looks like we are just making a bunch of new type declarations and passing empty string default values?

ui/shared/src/utils/PSData.ts
```diff
@@ -45,7 +46,8 @@ function extractReferenceCommitData(dryCommit: Commit): null | ReferenceCommit {
   const subject = referenceCommit.subject || '-';
   const body = referenceCommit.body || '-';
 
-  const date = referenceCommit.date ? formatDate(referenceCommit.date) : '-';
```
jwinters01, line 48: oh here, looks like we no longer use format date, the type you are inheriting automatically formats the date I guess? Does this mean we can clean up formatDate entirely?

ui/shared/src/utils/PSData.ts
```diff
@@ -45,7 +46,8 @@ function extractReferenceCommitData(dryCommit: Commit): null | ReferenceCommit {
...
+  // Pass RFC 3339 through for TimeAgo; do not formatDate here.
+  const date = referenceCommit.date;
```
crenshaw-dev, line 50: @jwinters01 this is the actual fix (repeated several places). We were calling `formatDate`, which produced something like "Jul 5 2025, 12:15pm EDT". But then the value was getting passed through `timeAgo`, which didn't support that date format. So it became NaN.

ui/shared/src/utils/PSData.ts
crenshaw-dev: > the type you are inheriting automatically formats the date

Yeah, the underlying data comes from the CR, which always formats the date as RFC3339

jwinters01 (approved):

zachaller (approved):

PR #1472, opened 2026-05-22
feat: don't reconcile on stale state

zachaller:


crenshaw-dev (approved):
approved, while very much looking forward to this: https://github.com/kubernetes-sigs/controller-runtime/pull/3473

PR #1486, opened 2026-05-27
feat: activePath support for monorepo environments

patjlm:
## Summary

Adds `spec.activePath` to `PromotionStrategy` and `ChangeTransferPolicy`, enabling monorepo setups where multiple applications share per-environment active branches. Each app occupies a distinct subdirectory (`activePath`), and the controller manages promotion independently per app.

This is a consolidation of #1337 with the following fixes on top, built against the current `main` (which includes #1417 and #1446).

## What activePath enables

Without `activePath`, each `PromotionStrategy` owns dedicated branches per environment. With `activePath`, multiple strategies can share the same active branch (e.g. `environment/prod`) — one per app path:

```
environment/prod              ← shared active branch
environment/prod-next/apps/payments   ← proposed branch for payments app
environment/prod-next/apps/frontend   ← proposed branch for frontend app
```

Each CTP reads/writes only its own subdirectory's `hydrator.metadata` and merges only its own path's changes into the shared active branch.

## Fixes included beyond #1337

**Fix 1 — Proposed branch naming** (`<env>-next/<activePath>` not `<env>/<activePath>-next`)  
The original naming scheme produced refs that collided with the active branch when `activePath` was a suffix. The correct scheme places `-next` on the environment segment before the path.

**Fix 2 — Path-aware merge conflict resolution** (`MergeWithOursStrategyForPath`)  
When the proposed branch contains changes from multiple apps (e.g. both `apps/payments` and `apps/frontend` diverged from active), the merge must preserve only `activePath`'s changes from the proposed branch and keep the active branch content for all other paths. A naive `ours` strategy would overwrite other apps' live state.

**Fix 3 — PR title includes activePath**  
Template updated to `Promote \`apps/payments\` (abc12) to \`environment/prod\`` when `activePath` is set, so PRs are clearly scoped.

**Fix 4 — Per-CTP git clone isolation**  
With `activePath`, multiple CTPs share the same active branch. Previously they shared one git clone (keyed by `activeBranch`), causing concurrent `git fetch` ref lock collisions under high reconcile concurrency. Each CTP now gets its own isolated clone (`activeBranch/activePath` key).

Note: commit-status label scoping (originally Fix 2 in #1408) is intentionally omitted — users set distinct `spec.key` on `ArgoCDCommitStatus`/`TimedCommitStatus` to differentiate per-app statuses (enabled by #1446).

## Changes

- **API**: `spec.activePath` on `PromotionStrategy` and `ChangeTransferPolicy` with CEL validation (no absolute paths, no `..` traversal)
- **Branch naming**: proposed branch is `<env-branch>-next/<activePath>`
- **Hydrator metadata**: reads from `<activePath>/hydrator.metadata` when set
- **Merge**: `MergeWithOursStrategyForPath` preserves only the activePath subtree from proposed branch
- **CRDs + apply configs + install.yaml**: updated
- **Docs**: new `repository-structure.md` covering single-branch-per-app vs shared active branch patterns; updates to `custom-hydrator.md` and `getting-started.md`
- **Tests**: new integration test contexts for single-strategy and multi-strategy activePath scenarios; git unit tests for metadata reading and path-aware merge

## Test plan

- [x] Unit tests: `go test ./internal/git/...` — passes
- [x] Controller unit tests: `go test ./internal/controller/... -run TestControllers` — passes (envtest suites excluded, pre-existing gap)
- [x] Manual end-to-end: 5 components × 6 environments, shared active branches, GitHub Actions custom hydrator — promotion flows work correctly per app path

### DIFF
--- a/api/v1alpha1/promotionstrategy_types.go
+++ b/api/v1alpha1/promotionstrategy_types.go
@@ -59,6 +59,13 @@ type PromotionStrategySpec struct {
 	// +listType:=map
 	// +listMapKey=branch
 	Environments []Environment `json:"environments"`
+
+	// ActivePath is the default repository subpath for this strategy's active state.
+	// When set, proposed branches are created as <environment-branch>-next/<activePath>.
+	// Individual environments can override this value via their own activePath field.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	ActivePath string `json:"activePath,omitempty"`
 }
 
 // Environment defines a single environment in the promotion sequence.
@@ -91,6 +98,12 @@ type Environment struct {
 	// +listType:=map
 	// +listMapKey=key
 	ProposedCommitStatuses []CommitStatusSelector `json:"proposedCommitStatuses,omitempty"`
+
+	// ActivePath optionally overrides the strategy-level activePath for this environment.
+	// When set, this environment's CTP uses this path instead of spec.activePath.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	ActivePath string `json:"activePath,omitempty"`
 }
 
 // GetAutoMerge returns the value of the AutoMerge field, defaulting to true if the field is nil.
--- a/internal/controller/changetransferpolicy_controller.go
+++ b/internal/controller/changetransferpolicy_controller.go
@@ -225,7 +225,7 @@ func (r *ChangeTransferPolicyReconciler) calculateHistory(ctx context.Context, c
 
 	history := make([]promoterv1alpha1.History, 0, len(shaListActive))
 	for _, sha := range shaListActive {
-		historyEntry, shouldInclude, err := r.buildHistoryEntry(ctx, sha, gitOperations)
+		historyEntry, shouldInclude, err := r.buildHistoryEntry(ctx, sha, ctp.Spec.ActivePath, gitOperations)
 		if err != nil {
 			logger.V(4).Info("failed to build history entry", "sha", sha, "err", err)
 			continue
@@ -240,7 +240,7 @@ func (r *ChangeTransferPolicyReconciler) calculateHistory(ctx context.Context, c
 }
 
 // buildHistoryEntry creates a single history entry for the given SHA
-func (r *ChangeTransferPolicyReconciler) buildHistoryEntry(ctx context.Context, sha string, gitOperations *git.EnvironmentOperations) (promoterv1alpha1.History, bool, error) {
+func (r *ChangeTransferPolicyReconciler) buildHistoryEntry(ctx context.Context, sha, activePath string, gitOperations *git.EnvironmentOperations) (promoterv1alpha1.History, bool, error) {
 	activeTrailers, err := gitOperations.GetTrailers(ctx, sha)
 	if err != nil {
 		return promoterv1alpha1.History{}, false, fmt.Errorf("failed to get trailers for SHA %q: %w", sha, err)
@@ -252,7 +252,7 @@ func (r *ChangeTransferPolicyReconciler) buildHistoryEntry(ctx context.Context, 
 		PullRequest: &promoterv1alpha1.PullRequestCommonStatus{},
 	}
 
-	r.populateActiveMetadata(ctx, &historyEntry, sha, gitOperations)
+	r.populateActiveMetadata(ctx, &historyEntry, sha, activePath, gitOperations)
 	r.populateProposedMetadata(ctx, &historyEntry, activeTrailers, gitOperations)
 	r.populatePullRequestMetadata(ctx, &historyEntry, activeTrailers)
 	r.populateCommitStatuses(ctx, &historyEntry, activeTrailers)
@@ -269,7 +269,7 @@ func getFirstTrailerValue(trailers map[string][]string, key string) string {
 }
 
 // populateActiveMetadata populates the active metadata for a history entry
-func (r *ChangeTransferPolicyReconciler) populateActiveMetadata(ctx context.Context, h *promoterv1alpha1.History, sha string, gitOperations *git.EnvironmentOperations) {
+func (r *ChangeTransferPolicyReconciler) populateActiveMetadata(ctx context.Context, h *promoterv1alpha1.History, sha, activePath string, gitOperations *git.EnvironmentOperations) {
 	logger := log.FromContext(ctx)
 	activeHydrated, err := gitOperations.GetShaMetadataFromGit(ctx, sha)
 	if err != nil {
@@ -278,7 +278,7 @@ func (r *ChangeTransferPolicyReconciler) populateActiveMetadata(ctx context.Cont
 	h.Active.Hydrated = activeHydrated
 	h.Active.Hydrated.Body = removeKnownTrailers(h.Active.Hydrated.Body)
 
-	activeDry, err := gitOperations.GetShaMetadataFromFile(ctx, sha)
+	activeDry, err := gitOperations.GetShaMetadataFromFile(ctx, sha, activePath)
 	if err != nil {
 		logger.V(4).Info("failed to get active historic metadata from file", "sha", sha, "error", err)
 	}
@@ -538,7 +538,7 @@ func (r *ChangeTransferPolicyReconciler) calculateStatus(ctx context.Context, ct
 	// to be made concurrency-safe for a single identity first; today its EnvironmentOperations methods share one
 	// on-disk clone and must be called sequentially (see the internal/git package documentation).
 
-	proposedShas, err := gitOperations.GetBranchShas(ctx, ctp.Spec.ProposedBranch)
+	proposedShas, err := gitOperations.GetBranchShas(ctx, ctp.Spec.ProposedBranch, ctp.Spec.ActivePath)
 	if err != nil {
 		// If the proposed branch doesn't exist, it's likely because the hydrator hasn't run yet
 		if strings.Contains(err.Error(), "couldn't find remote ref") {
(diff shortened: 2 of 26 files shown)


internal/controller/changetransferpolicy_controller.go
```diff
@@ -142,7 +142,11 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get GitRepository: %w", err)
 	}
-	gitOperations := git.NewEnvironmentOperations(gitRepo, gitAuthProvider, ctp.Spec.ActiveBranch)
+	cloneKey := ctp.Spec.ActiveBranch
+	if ctp.Spec.ActivePath != "" {
+		cloneKey = ctp.Spec.ActiveBranch + "/" + ctp.Spec.ActivePath
+	}
+	gitOperations := git.NewEnvironmentOperations(gitRepo, gitAuthProvider, cloneKey)
```
crenshaw-dev: I'd use a struct instead of a string for the key. Strings are a recipe for key collissions.

mkdocs.yml
```diff
@@ -81,4 +82,5 @@ plugins:
   - gh-admonitions
   - redirects:
       redirect_maps:
-        'metrics.md': 'monitoring/metrics.md' # Page was moved to a subdirectory.
\ No newline at end of file
+        'metrics.md': 'monitoring/metrics.md' # Page was moved to a subdirectory.
+        'contributing.md': 'contributing/index.md' # Page was moved to a subdirectory.
```
crenshaw-dev: We can drop this. The redirect ended up being unnecessary and causing an infinite loop.

internal/git/git.go
```diff
@@ -111,8 +112,15 @@ type BranchShas struct {
 	Hydrated string
 }
 
+func buildHydratorMetadataPath(activePath string) string {
+	if activePath == "" {
+		return "hydrator.metadata"
+	}
+	return path.Join(activePath, "hydrator.metadata")
+}
```
crenshaw-dev, line 171: I'm trying to think of how we could secure this from path traversal. I think maybe we could use os.Root and attempt some kind of read operation before running the git command?

crenshaw-dev (changes requested):

internal/controller/changetransferpolicy_controller.go
patjlm: Done — switched to a `gitpaths.Key` struct with `RepoURL`, `ActiveBranch`, and `ActivePath` fields. Also simplified `NewEnvironmentOperations` to take `activeBranch` and `activePath` as separate parameters, so the git layer builds the key internally and no concatenation happens in the controller.

mkdocs.yml
patjlm: Dropped, thanks.

internal/git/git.go
patjlm: Added a runtime validation in `buildHydratorMetadataPath` that mirrors the CEL rules: it rejects any `activePath` whose `path.Clean` result is absolute or starts with `..'`. The CRD CEL validation is the first line of defense; this adds defense-in-depth at the git layer.

On `os.Root`: since `activePath` is used as a git object tree path in `git show origin/branch:<path>` commands (not as a local filesystem path), `os.Root` wouldn't provide additional protection here — git accesses the object tree, not the disk. That said, happy to explore it if you see a gap I'm missing.

internal/git/git.go
```diff
@@ -111,10 +125,23 @@ type BranchShas struct {
...
+func buildHydratorMetadataPath(activePath string) (string, error) {
+	if activePath == "" {
+		return "hydrator.metadata", nil
+	}
+	// Validate that the cleaned path stays within the repository root.
+	// This mirrors the CEL validation on the CRD and adds defense-in-depth at the git layer.
+	cleaned := path.Clean(activePath)
+	if path.IsAbs(cleaned) || strings.HasPrefix(cleaned, "..") {
+		return "", fmt.Errorf("activePath %q is invalid: must not be absolute or traverse above the repository root", activePath)
+	}
```
crenshaw-dev: My experience doing path validation/sanitization with the `path` library has been really, really bad.

I've looked into it a bit more, and here's what I'd suggest: let's rely on git to protect us from the worst possible misuses. 

The two commands we're dealing with are `git show`, which accepts a "gitrevision"; and `git checkout`, which accepts a pathspec.

In both cases, `activePath` will be interpreted as a "tree path," not a plan OS filesystem path. Tree paths don't treat `..` as traversal, it's treated as a literal part of the path.

And in both cases, git constrains reads and writes within the confines of the git repo, which is probably our trust boundary anyway. Multiple teams share a single git branch, but all their tooling has full write access to that branch, so even if someone broke out of their `activePath`, they wouldn't really get somewhere that wasn't relatively accessible anyway.

I've got one more mitigation I'll comment below. But I wanna backtrack and suggest that we drop all CEL and `path.*` validation and instead rely on git. Validating paths is fraught and, I think, not really necessary in this case.

internal/git/git.go
```diff
@@ -480,11 +515,64 @@ func (g *EnvironmentOperations) MergeWithOursStrategy(ctx context.Context, propo
...
+		return fmt.Errorf("failed to merge branch %q into %q: %w (stdout: %s, stderr: %s)", activeBranch, proposedBranch, err, stdout, stderr)
+	}
+
+	_, stderr, err = g.runCmd(ctx, gitPath, "checkout", "origin/"+activeBranch, "--", ".")
+	if err != nil {
+		logger.Error(err, "Failed to checkout active branch content", "activeBranch", activeBranch, "stderr", stderr)
+		return fmt.Errorf("failed to checkout active branch content from %q: %w (stderr: %s)", activeBranch, err, stderr)
+	}
+
+	_, stderr, err = g.runCmd(ctx, gitPath, "checkout", "origin/"+proposedBranch, "--", activePath)
```
crenshaw-dev: git pathspecs are weirdly powerful. They accept globs and other similar magic. This prefix should prevent someone from accidentally (or purposefully) stepping into that magic:


```suggestion
	_, stderr, err = g.runCmd(ctx, gitPath, "checkout", "origin/"+proposedBranch, "--", ":(literal)" + activePath)
```

internal/git/git.go
crenshaw-dev: I ended up backtracking on this. See comment below. 🙂

api/v1alpha1/promotionstrategy_types.go
```diff
@@ -59,6 +59,13 @@ type PromotionStrategySpec struct {
 	// +listType:=map
 	// +listMapKey=branch
 	Environments []Environment `json:"environments"`
+
+	// ActivePath is an optional repository subpath for this strategy's active state.
+	// When set, proposed branches are created as <environment-branch>-next/<activePath>.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="!self.startsWith('/') && !self.endsWith('/') && !self.contains('/..') && !self.startsWith('..')",message="activePath must not be an absolute path, end with '/', or contain '..' path components"
+	ActivePath string `json:"activePath,omitempty"`
```
crenshaw-dev: Is it necessary to set this globally, or could it be per-environment?

internal/git/git.go
patjlm: Agreed — dropped the CEL `x-kubernetes-validations` on both CRD types and simplified `buildHydratorMetadataPath` back to a plain `path.Join`. While in the area I also fixed a related correctness bug (Copilot's comment below on line 547): `git checkout` doesn't remove files deleted in the proposed branch, so I now do a `git rm -rf --ignore-unmatch` on the activePath subtree before the checkout, backed by a regression test.

internal/git/git.go
patjlm: Applied, thanks!

internal/git/git.go
```diff
@@ -480,11 +515,64 @@ func (g *EnvironmentOperations) MergeWithOursStrategy(ctx context.Context, propo
...
+	if err != nil {
+		logger.Error(err, "Failed to checkout proposed activePath content", "proposedBranch", proposedBranch, "activePath", activePath, "stderr", stderr)
+		return fmt.Errorf("failed to checkout activePath %q from proposed branch %q: %w (stderr: %s)", activePath, proposedBranch, err, stderr)
+	}
```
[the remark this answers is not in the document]
patjlm: Fixed — added `git rm -r -f --ignore-unmatch -- :(literal)<activePath>` before the checkout so deletions in the proposed branch propagate correctly. Added a regression test.

internal/git/git.go
```diff
@@ -111,10 +125,23 @@ type BranchShas struct {
...
+	if activePath == "" {
+		return "hydrator.metadata", nil
+	}
+	// Validate that the cleaned path stays within the repository root.
+	// This mirrors the CEL validation on the CRD and adds defense-in-depth at the git layer.
+	cleaned := path.Clean(activePath)
+	if path.IsAbs(cleaned) || strings.HasPrefix(cleaned, "..") {
+		return "", fmt.Errorf("activePath %q is invalid: must not be absolute or traverse above the repository root", activePath)
+	}
+	return path.Join(activePath, "hydrator.metadata"), nil
```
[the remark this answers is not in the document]
patjlm: Dropped along with the CEL validation (see crenshaw-dev's comment above) — `buildHydratorMetadataPath` is now a plain `path.Join`, so nothing to validate here.

api/v1alpha1/promotionstrategy_types.go
patjlm: Good call — added per-environment override support. `Environment.ActivePath` now optionally overrides the strategy-level default. This covers cases where a hydrator generates files in different locations depending on the target environment. Non-overridden environments fall back to the global `spec.activePath`. Added a controller test covering the override behavior.

crenshaw-dev:
@patjlm I'm working on a few concurrency issues in other PRs to make sure we don't run into issues when this is merged. Coming back for another round of review ASAP. :-)

crenshaw-dev (approved):
Thanks so much for the awesome feature @patjlm!

PR #1488, opened 2026-05-28
fix(gitlab): use project-scoped ListProjectMergeRequests in FindOpen

ZeBidule:
## Summary

`FindOpen` in `internal/scms/gitlab/pullrequest.go` uses the global `ListMergeRequests` API endpoint (`GET /api/v4/merge_requests`) instead of the project-scoped `ListProjectMergeRequests` (`GET /api/v4/projects/:id/merge_requests`). This causes the function to return merge requests from **any project** on the GitLab instance that matches the source/target branch names.

## Impact

When multiple repositories use the same branch naming convention (e.g. `environment/validation-next` → `environment/validation`), the first MR returned by the global API pollutes all other repositories' PullRequest objects with the wrong MR IID. This causes:

- `ChangeTransferPolicy` resources to enter a degraded state (Ready=False)
- The controller to attempt updating MRs in the wrong projects (404 errors or silently updating the wrong MR)
- Promotion to be completely blocked for all affected repositories

In our deployment, 10 out of 11 addon repositories were broken because they all picked up MR #9 from an unrelated project (reloader). Only reloader itself worked because its own MR happened to be #9.

## Fix

Replace `ListMergeRequestsOptions` / `ListMergeRequests(options)` with `ListProjectMergeRequestsOptions` / `ListProjectMergeRequests(repo.Spec.GitLab.ProjectID, options)`.

Both option structs share the same fields (`SourceBranch`, `TargetBranch`, `State`), and `ListProjectMergeRequests` returns `[]*BasicMergeRequest` with the same `IID`/`CreatedAt` fields used downstream — so this is a drop-in replacement.

## Versions Affected

- v0.26.2 (confirmed)
- v0.30.0 (confirmed — code is identical at this location)

### DIFF
--- a/internal/scms/gitlab/pullrequest.go
+++ b/internal/scms/gitlab/pullrequest.go
@@ -240,14 +240,14 @@ func (pr *PullRequest) FindOpen(ctx context.Context, pullRequest v1alpha1.PullRe
 		return false, "", time.Time{}, fmt.Errorf("failed to get repo: %w", err)
 	}
 
-	options := &gitlab.ListMergeRequestsOptions{
+	options := &gitlab.ListProjectMergeRequestsOptions{
 		SourceBranch: gitlab.Ptr(pullRequest.Spec.SourceBranch),
 		TargetBranch: gitlab.Ptr(pullRequest.Spec.TargetBranch),
 		State:        gitlab.Ptr("opened"),
 	}
 
 	start := time.Now()
-	mrs, resp, err := pr.client.MergeRequests.ListMergeRequests(options)
+	mrs, resp, err := pr.client.MergeRequests.ListProjectMergeRequests(repo.Spec.GitLab.ProjectID, options)
 	if resp == nil {
 		statusCode := -1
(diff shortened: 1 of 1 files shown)


ZeBidule:
We've built and deployed a patched image with this fix on our clusters (2 clusters running gitops-promoter managing 11 addon repositories with identical branch naming). After applying the fix and cleaning up the stale PullRequest objects, all 10 previously-broken addons immediately recovered — the controller now correctly finds each project's own MR instead of picking up an unrelated one from the global API.

zachaller (approved):

crenshaw-dev (approved):
Thank you!

If you're able, add yourself to USERS.md! We love to know who's using the tool. :-)

ZeBidule:
Thanks a lot for your reactivity :)
I will be happy to add us in USERS.md if I have the approval of my management but for the moment they do not understand the concept so they refuse :(
We will continue to try to convince them and perhaps it will change...

PR #1493, opened 2026-05-28
feat: adding instanceID label to support n installs of gitops-promoter 

pbhatnagar-oss:
### Summary

  Adds a new `--instance-id` flag (also `GITOPS_PROMOTER_INSTANCE_ID` env var) to the Promoter controller. When set, the controller only reconciles resources carrying the matching `promoter.argoproj.io/instance-id` label. This enables multiple Promoter installs to share a single Kubernetes API server without stepping on each other — each install scopes to the resources it owns.

  When unset (the default), behavior is unchanged: the controller reconciles every resource in scope, exactly as before. **No existing deployment needs to change.**

  ### Motivation

  There are operational scenarios where multiple Promoter installs need to coexist on the same cluster — e.g. operators may want each install scoped to a distinct subset of `PromotionStrategies`, partitioned by team, environment tier, release cadence, or any other logical grouping. Each install runs in its own namespace, but they all watch the same Kubernetes API server. Without partitioning, every install receives watch events for every CR in the cluster and they all race to
  reconcile.

  Argo CD's application controller solves an analogous problem with hash-based sharding. Promoter's needs are simpler: operators declare the partition explicitly via a label; the controller filters via a controller-runtime predicate. No hashing, no rebalancing, no shard heartbeats.

  ### How it works

  The flag value flows through five layers:

  1. **`cmd/main.go`** reads `--instance-id` (or the env-var fallback) and threads the value into every reconciler struct and the webhook receiver.
  2. **Each reconciler's `SetupWithManager`** wraps the existing predicate with `predicate.And(existing, promoterpredicate.InstanceID(r.InstanceID))`. Events for resources with the wrong label are dropped before reaching the work queue.
  3. **Webhook receiver** (`internal/webhookreceiver/server.go`) adds a `client.MatchingLabels` selector to the `findChangeTransferPolicy` List calls. SCM webhooks only enqueue CTPs belonging to this install.
  4. **`ChangeTransferPolicyReconciler`'s `WatchesRawSource(Channel)`** — which bypasses standard predicates — has an in-closure label check at the channel sender to drop cross-instance enqueues.
  5. **Child-resource creation sites** propagate the label from parent to child via helpers in `internal/utils/labels.go`. This keeps the predicate working at every level of the resource graph.

  A new package `internal/predicate/` exports `InstanceID(id string) predicate.Predicate` — returns a pass-all predicate when `id == ""` (backwards-compat invariant) and a label-match predicate otherwise.

  ### Design decisions worth flagging for reviewers

  1. **Label key: `promoter.argoproj.io/instance-id`** — matches the existing `promoter.argoproj.io/*` convention used elsewhere in the repo. Exported from `api/v1alpha1/constants.go` for reuse.

  2. **Single string value, not a selector.** `--instance-id wave-0` style. A more flexible selector form can be added later without breaking the API; the predicate helper is structured to support it.

  3. **`ArgoCDCommitStatusReconciler.Watches(&argocd.Application{}, ...)` is intentionally NOT filtered.** Remote-cluster `Application` objects don't carry our label by design (cross-cluster label stamping is out of scope and would require coordination across other codebases). The InstanceID predicate is applied to the local `For(ArgoCDCommitStatus)` and `Watches(PromotionStrategy)` — sufficient for partitioning the local reconciler's work. A dedicated AST test
  (`TestArgoCDCommitStatusOptionB`) asserts this asymmetry stays in place against future refactors.

  4. **`ClusterScmProviderReconciler` is treated like the namespaced reconcilers.** Cluster-scoped resources can carry the label; the cluster scope makes no difference to label-based filtering.

  5. **Two label-copy helpers** in `internal/utils/labels.go`: `CopyInstanceIDLabel(parent, child client.Object)` and `CopyInstanceIDLabelToMap(parent client.Object, m map[string]string) map[string]string`. The second exists for Server-Side Apply call sites that take a raw map (which doesn't satisfy `client.Object`); composing main's existing `CommitStatusStandardLabels` with the map variant keeps the SSA call chain fluent.

  ### Backwards compatibility

  When `--instance-id` is unset (single-instance deployments):

  - `predicate.InstanceID("")` returns a uniform pass-all predicate for Create/Update/Delete/Generic events.
  - Both label-copy helpers are no-ops when the parent carries no instance-ID label.
  - The webhook receiver skips the `MatchingLabels` ListOption — the List remains cluster-wide-by-SHA, exactly as before.
  - The CTP channel-sender filter short-circuits.

  Result: zero behavioral diff between this PR and `main` for single-instance deployments. Verified at all five surface areas.

  ### Testing

  - **Predicate truth table** (`internal/predicate/instanceid_test.go`): empty-id allows everything across 5 label scenarios × 4 event types; non-empty-id matches correctly across the same matrix.
  - **Label-copy helpers** (`internal/utils/labels_test.go`): nil-labels, sibling preservation, stale-value overwrite, no-op cases.
  - **Per-reconciler `SetupWithManager` wiring** (`internal/controller/setup_predicate_chain_test.go`): AST-based assertions over all 13 reconcilers + 3 `Watches(PromotionStrategy)` sites: the predicate is chained via `And` (not `Or`); ArgoCDCommitStatus's `Watches(Application)` is intentionally unfiltered. Validated via mutation testing before commit — the assertions catch real regressions.

  ### Out of scope

  - **End-to-end two-instance envtest.** The existing test scaffolding hosts a single Promoter manager; building a two-manager envtest harness is a separate piece of work I'd be happy to follow up with as a dedicated PR. The combination of the predicate truth table + the AST chain-wiring proof gives strong confidence in the partitioning contract for Alpha.
  - **Deployment manifests for multi-install setups.** This PR adds the controller-side primitive; how operators package and deploy N installs is left to their existing Helm/Kustomize workflow.
  - **Metric labels.** Adding `instance_id` to Prometheus metrics is a separate small change.

  ### Checklist

  - [x] All commits DCO signed-off (`Signed-off-by:` trailer present, verified)
  - [x] All commits GPG signed
  - [x] `go test ./internal/predicate/... ./internal/utils/... ./internal/controller/...` passes locally
  - [x] `go vet` and `fieldalignment` clean on touched packages
  - [x] No new lint findings (existing `ui/web/embed.go: pattern static` reproduces on `main` and is unrelated)
  - [x] Commit messages follow Conventional Commits style

## Summary by CodeRabbit

* **New Features**
  * Added multi-install controller partitioning using the `promoter.argoproj.io/instance-id` label, including instance-aware caching.
  * Extended `ControllerConfiguration` with `spec.instanceID`, plus `status.observedGeneration`, `status.conditions`, and `status.instanceID`.
  * Promoter resources now expose `status.instanceID` (mirrored from the label) after successful reconciliation.

* **Bug Fixes**
  * Improved handling of optional `instanceID` values to avoid incorrect copying.

* **Documentation**
  * Expanded multi-install, labeling, and troubleshooting guides, including promotion/gating behavior details.

* **Tests**
  * Added coverage for instance-id propagation, migration, and cache scoping behavior.


crenshaw-dev:
Thanks @pbhatnagar-oss! Could you add user-facing docs about how to use the feature?

docs/multi-install.md
```diff
@@ -0,0 +1,179 @@
+# Multi-Install Deployments
+
+GitOps Promoter can be deployed as multiple independent installs on a single Kubernetes cluster, with each install scoped
+to a distinct subset of resources. This is useful when you want to partition reconciliation by team, environment tier,
+release cadence, or another logical grouping — without each install racing to reconcile every PromotionStrategy on the
+cluster.
+
+The mechanism is a single label-key contract: each Promoter install runs with an `--instance-id` flag (or the
+`GITOPS_PROMOTER_INSTANCE_ID` environment variable) and only reconciles resources carrying a matching
+`promoter.argoproj.io/instance-id` label.
```
crenshaw-dev: Let's add an `instanceID` field to the ControllerConfiguration CRD (same field validation as label values) instead of configuring it via a flag/env var.

docs/multi-install.md
```diff
@@ -0,0 +1,179 @@
...
+spec:
+  ...
+```
+
+The same label key is used on:
+
+* `PromotionStrategy`
+* `ScmProvider` and `ClusterScmProvider`
+* `GitRepository`
+* `CommitStatus` and its sub-variants (`GitCommitStatus`, `WebRequestCommitStatus`, `TimedCommitStatus`,
```
crenshaw-dev: @zachaller I think the CommitStatus controller should only _reconcile_ CommitStatus resources with the matching instance-id label.

But I think all controllers should _observe_ **all** CommitStatus resources when, for example, making gating decisions.

Does that make sense?

docs/multi-install.md
```diff
@@ -0,0 +1,179 @@
...
+  `ArgoCDCommitStatus`)
+* `ControllerConfiguration`
```
crenshaw-dev: I think we shouldn't support the instance ID label on ControllerConfiguration. Instead, the controller's ControllerConfiguration (as determined by being in the same namespace with the hard-coded config resource name) should be the source of truth for the controller's instance ID.

crenshaw-dev (changes requested):

docs/multi-install.md
zachaller: Yea we currently also don't support setting a ControllerConfiguration to use and so we force instance-id to be used with multiple namespace only.

docs/multi-install.md
zachaller: I think the should be fully independent it's awkward to mix each reconcile loops should check for instance id and only reoncile that ever list or get of child resources should filter with the id its what I would expect behavior to be

internal/utils/labels.go
```diff
@@ -0,0 +1,43 @@
...
+	v, ok := parent.GetLabels()[promoterv1alpha1.InstanceIDLabel]
+	if !ok || v == "" {
+		return labels
+	}
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels[promoterv1alpha1.InstanceIDLabel] = v
+	return labels
+}
```
zachaller, line 20: **[blocking] Add lookup-side helpers so every reconcile/List can enforce the instance-id contract**

`CopyInstanceIDLabel`/`CopyInstanceIDLabelToMap` cover the *write* side (label propagation), but there's nothing for the *read* side. The predicates only gate watch events; `Owns()` watches, `EnqueueRequestsFromMapFunc` handlers, and `WatchesRawSource` all enqueue requests that never pass through `For(...)`'s predicates, so each `Reconcile` and each `List`/`Get` of child resources needs its own check. Suggest adding to this file (the other review comments reference these helpers):

```go
// MatchesInstanceID reports whether obj belongs to the install identified by id.
// Empty id preserves single-install semantics: every object matches.
func MatchesInstanceID(obj client.Object, id string) bool {
	return id == "" || obj.GetLabels()[promoterv1alpha1.InstanceIDLabel] == id
}

// InstanceIDSelector returns a labels.Selector scoping List calls to this
// install's instance id, or labels.Everything() when id is empty.
// For use with client.ListOptions{LabelSelector: ...}.
func InstanceIDSelector(id string) labels.Selector {
	if id == "" {
		return labels.Everything()
	}
	return labels.SelectorFromSet(labels.Set{promoterv1alpha1.InstanceIDLabel: id})
}

// InstanceIDListOption is InstanceIDSelector as a variadic client.ListOption,
// for List(ctx, list, client.InNamespace(...), utils.InstanceIDListOption(id)) call sites.
func InstanceIDListOption(id string) client.ListOption {
	return client.MatchingLabelsSelector{Selector: InstanceIDSelector(id)}
}
```

(needs `"k8s.io/apimachinery/pkg/labels"` import)

The reconcile-time guard pattern is the same everywhere: store the already-fetched `instanceID` on the reconciler struct in `SetupWithManager` (`r.instanceID = instanceID`), then check after the `Get` in `Reconcile`. One subtlety: the deferred `utils.HandleReconciliationResult` would otherwise SSA-patch status on the foreign object — and since both installs use the **same field-manager name**, that's a cross-instance ownership fight. Zero the object before returning so the defer's existing `GetName()==""` early-exit kicks in.

zachaller (commented):
Reviewing against the multi-install partitioning contract: **(1) every reconcile must respect its instance ID, and (2) every list/get of child resources must filter by instance ID.**

Summary of the inline comments:

- **2 critical propagation bugs** — `TimedCommitStatus` and `WebRequestCommitStatus` don't propagate the instance-id label to the CommitStatuses they create, so in multi-install mode those gates silently never pass (this install's own CommitStatus predicate drops their events).
- **Reconcile-time guards missing** — predicates only gate watch events. `Owns()` watches, `EnqueueRequestsFromMapFunc` handlers, and `WatchesRawSource` channel sends enqueue requests that bypass the `For(...)` predicates, so each `Reconcile` needs a label check. (The comment at `changetransferpolicy_controller.go` L519–520 claims such a short-circuit exists — it doesn't.)
- **Unfiltered Lists** — the only Lists that gained an instance-id selector are the webhook receiver's CTP lookups. The map-handler Lists, CTP gating/PR lookups, CommitStatus→CTP fan-out, and orphan-cleanup Lists are all unfiltered.

📍 Some comments are anchored to nearby in-diff lines because their actual targets aren't part of this PR's hunks; each one starts with a pointer to the real location.

Two follow-ups not tied to a line:

1. **Test gap**: `setup_predicate_chain_test.go` proves `SetupWithManager` wiring only. Please also assert (a) every child-creation site composes `CopyInstanceIDLabel*` (would have caught both propagation bugs) and (b) every map-function List passes an instance-id list option.
2. **Docs**: `docs/multi-install.md` should describe the enforcement layers explicitly (predicates → reconcile guards → filtered Lists/Gets) and document the gating-semantics decision (do cross-instance CommitStatuses gate promotions?) once the open thread with @crenshaw-dev is resolved.

internal/controller/timedcommitstatus_controller.go
```diff
@@ -146,9 +147,20 @@ func (r *TimedCommitStatusReconciler) SetupWithManager(ctx context.Context, mgr
...
 
+	instanceID, err := r.SettingsMgr.GetInstanceIDDirect(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get InstanceID from ControllerConfiguration: %w", err)
+	}
+
 	err = ctrl.NewControllerManagedBy(mgr).
-		For(&promoterv1alpha1.TimedCommitStatus{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Watches(&promoterv1alpha1.PromotionStrategy{}, r.enqueueTimedCommitStatusForPromotionStrategy()).
+		For(&promoterv1alpha1.TimedCommitStatus{}, builder.WithPredicates(predicate.And(
```
zachaller: 📍 *Applies to `upsertCommitStatus` (L301–303), which isn't part of this PR's diff hunks — anchored here on the predicate wiring instead.*

**[blocking] Missing instance-id propagation — TimedCommitStatus gates silently break in multi-install mode**

This is one of two child-creation sites that didn't get `CopyInstanceIDLabelToMap` (the other is WebRequestCommitStatus). The CommitStatus created in `upsertCommitStatus` carries no instance-id label, so this install's own `CommitStatusReconciler` predicate **drops its events** — the status never reaches the SCM and the time gate never passes. This contradicts the PR description's "child resources automatically inherit the label" claim and `docs/multi-install.md`.

```diff
-	// Build the apply configuration
+	// Build the apply configuration. Compose main's standard-labels helper with
+	// our instance-id propagation (ARGO-3085).
+	commitStatusLabels := utils.CopyInstanceIDLabelToMap(
+		tcs,
+		utils.CommitStatusStandardLabels(tcs, branch, key),
+	)
 	commitStatusApply := acv1alpha1.CommitStatus(commitStatusName, tcs.Namespace).
-		WithLabels(utils.CommitStatusStandardLabels(tcs, branch, key)).
+		WithLabels(commitStatusLabels).
```

Please also extend `setup_predicate_chain_test.go` (or a sibling test) to assert all seven CommitStatus/child creation sites call `CopyInstanceIDLabel*` — the AST test only proves `SetupWithManager` wiring and cannot catch exactly this class of regression.

internal/controller/timedcommitstatus_controller.go
```diff
@@ -146,9 +147,20 @@ func (r *TimedCommitStatusReconciler) SetupWithManager(ctx context.Context, mgr
...
+			predicate.GenerationChangedPredicate{},
+			promoterpredicate.InstanceID(instanceID),
+		))).
+		Watches(&promoterv1alpha1.PromotionStrategy{},
+			r.enqueueTimedCommitStatusForPromotionStrategy(),
```
zachaller: 📍 *Applies to `Reconcile` (after the Get, ~L80–88) and `enqueueTimedCommitStatusForPromotionStrategy` (L374–377) — anchored here on the Watches wiring.*

**[blocking] Reconcile guard + unfiltered List in the PromotionStrategy map handler**

Two gaps in this controller:

1. Requests produced by `enqueueTimedCommitStatusForPromotionStrategy` **do not pass through the `For(...)` predicates** (map-function requests are direct queue inserts). The `Watches` predicate filters the *source* PS event, but the List inside the handler enqueues *every* TCS in the namespace — including foreign-instance ones:

```diff
 		// List all TimedCommitStatus resources in the same namespace
 		var tcsList promoterv1alpha1.TimedCommitStatusList
-		if err := r.List(ctx, &tcsList, client.InNamespace(ps.Namespace)); err != nil {
+		if err := r.List(ctx, &tcsList, client.InNamespace(ps.Namespace), utils.InstanceIDListOption(r.instanceID)); err != nil {
```

2. Defense-in-depth guard in `Reconcile` (requires an `instanceID string` field on the struct, set in `SetupWithManager`):

```diff
 	err = r.Get(ctx, req.NamespacedName, &tcs, &client.GetOptions{})
 	if err != nil {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&tcs, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", tcs.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		tcs = promoterv1alpha1.TimedCommitStatus{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

Optionally also verify the referenced PS at step 2 (`r.Get(ctx, psKey, &ps)`): a TCS labeled for instance A referencing a PS labeled for instance B is a misconfiguration worth surfacing as an error rather than silently acting on it.

internal/controller/webrequestcommitstatus_controller.go
```diff
@@ -226,9 +227,20 @@ func (r *WebRequestCommitStatusReconciler) SetupWithManager(ctx context.Context,
...
 
+	instanceID, err := r.SettingsMgr.GetInstanceIDDirect(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get InstanceID from ControllerConfiguration: %w", err)
+	}
+
 	err = ctrl.NewControllerManagedBy(mgr).
-		For(&promoterv1alpha1.WebRequestCommitStatus{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Watches(&promoterv1alpha1.PromotionStrategy{}, r.enqueueWebRequestCommitStatusForPromotionStrategy()).
+		For(&promoterv1alpha1.WebRequestCommitStatus{}, builder.WithPredicates(predicate.And(
```
zachaller: 📍 *Applies to `upsertCommitStatus` (L486–487), which isn't part of this PR's diff hunks — anchored here on the predicate wiring instead.*

**[blocking] Missing instance-id propagation — same bug as TimedCommitStatus**

`upsertCommitStatus` here doesn't propagate the label, so in multi-install mode web-request gate CommitStatuses are invisible to this install's CommitStatus reconciler and never reach the SCM:

```diff
+	// Compose main's standard-labels helper with our instance-id propagation (ARGO-3085).
+	commitStatusLabels := utils.CopyInstanceIDLabelToMap(
+		wrcs,
+		utils.CommitStatusStandardLabels(wrcs, branch, wrcs.Spec.Key),
+	)
 	commitStatusApply := acv1alpha1.CommitStatus(commitStatusName, wrcs.Namespace).
-		WithLabels(utils.CommitStatusStandardLabels(wrcs, branch, wrcs.Spec.Key)).
+		WithLabels(commitStatusLabels).
```

internal/controller/webrequestcommitstatus_controller.go
```diff
@@ -226,9 +227,20 @@ func (r *WebRequestCommitStatusReconciler) SetupWithManager(ctx context.Context,
...
+			predicate.GenerationChangedPredicate{},
+			promoterpredicate.InstanceID(instanceID),
+		))).
+		Watches(&promoterv1alpha1.PromotionStrategy{},
+			r.enqueueWebRequestCommitStatusForPromotionStrategy(),
```
zachaller: 📍 *Applies to `Reconcile` (~L140–148) and `enqueueWebRequestCommitStatusForPromotionStrategy` (L518–521) — anchored here on the Watches wiring.*

**[blocking] Reconcile guard + unfiltered List in map handler**

Same two gaps as TimedCommitStatus. This controller already has the `skipStatusWrite` flag, so the guard is cleaner here — no object-zeroing needed:

```diff
 	if r.rvTracker.IsCacheStale(req.NamespacedName, wrcs.ResourceVersion) {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&wrcs, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", wrcs.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		skipStatusWrite = true
+		return ctrl.Result{}, nil
+	}
```

And in `enqueueWebRequestCommitStatusForPromotionStrategy`:

```diff
-		if err := r.List(ctx, &wrcsList, client.InNamespace(ps.Namespace)); err != nil {
+		if err := r.List(ctx, &wrcsList, client.InNamespace(ps.Namespace), utils.InstanceIDListOption(r.instanceID)); err != nil {
```

(both require storing `r.instanceID` in `SetupWithManager`; the HTTP side effects this controller fires make the reconcile guard especially important — a cross-instance reconcile here doesn't just waste CPU, it can re-fire trigger webhooks)

internal/controller/gitcommitstatus_controller.go
```diff
@@ -148,9 +149,20 @@ func (r *GitCommitStatusReconciler) SetupWithManager(ctx context.Context, mgr ct
...
+
 	err = ctrl.NewControllerManagedBy(mgr).
-		For(&promoterv1alpha1.GitCommitStatus{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Watches(&promoterv1alpha1.PromotionStrategy{}, r.enqueueGitCommitStatusForPromotionStrategy()).
+		For(&promoterv1alpha1.GitCommitStatus{}, builder.WithPredicates(predicate.And(
+			predicate.GenerationChangedPredicate{},
+			promoterpredicate.InstanceID(instanceID),
+		))).
+		Watches(&promoterv1alpha1.PromotionStrategy{},
+			r.enqueueGitCommitStatusForPromotionStrategy(),
```
zachaller: 📍 *Applies to `Reconcile` (~L90–99) and `enqueueGitCommitStatusForPromotionStrategy` (L458–462) — anchored here on the Watches wiring.*

**[blocking] Reconcile guard + unfiltered List in map handler**

Same pattern as the other two gate controllers (the label propagation at L418 is already correct here 👍):

```diff
 	err = r.Get(ctx, req.NamespacedName, &gcs, &client.GetOptions{})
 	if err != nil {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&gcs, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", gcs.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		gcs = promoterv1alpha1.GitCommitStatus{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

```diff
 		// List all GitCommitStatus resources in the same namespace
 		var gcsList promoterv1alpha1.GitCommitStatusList
-		if err := r.List(ctx, &gcsList, client.InNamespace(ps.Namespace)); err != nil {
+		if err := r.List(ctx, &gcsList, client.InNamespace(ps.Namespace), utils.InstanceIDListOption(r.instanceID)); err != nil {
```

internal/controller/promotionstrategy_controller.go
```diff
@@ -176,8 +178,16 @@ func (r *PromotionStrategyReconciler) SetupWithManager(ctx context.Context, mgr
...
+		return fmt.Errorf("failed to get InstanceID from ControllerConfiguration: %w", err)
+	}
+
 	err = ctrl.NewControllerManagedBy(mgr).
-		For(&promoterv1alpha1.PromotionStrategy{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&promoterv1alpha1.PromotionStrategy{}, builder.WithPredicates(predicate.And(
+			predicate.GenerationChangedPredicate{},
+			promoterpredicate.InstanceID(instanceID),
+		))).
 		Owns(&promoterv1alpha1.ChangeTransferPolicy{}).
```
zachaller, line 200: **[blocking] `Owns(ChangeTransferPolicy)` bypasses the InstanceID predicate — guard needed; cleanup List unfiltered**

`builder.WithPredicates` inside `For(...)` applies only to that watch. This `Owns(&ChangeTransferPolicy{})` carries no InstanceID predicate, so any CTP event (e.g. a pre-migration unlabeled child) maps to its owner PS and enqueues it — and `Reconcile` will then upsert CTPs/CommitStatuses for a foreign PromotionStrategy. Guard after the Get (~L95–109, outside this diff):

```diff
 	// If the resource is being deleted, stop reconciling immediately without requeuing
 	if !ps.DeletionTimestamp.IsZero() {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&ps, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", ps.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		ps = promoterv1alpha1.PromotionStrategy{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

(Alternative: add `builder.WithPredicates(instanceIDPredicate)` to the `Owns(...)` clause — but the guard is still needed for full coverage, e.g. the cross-controller enqueue paths.)

In `cleanupOrphanedChangeTransferPolicies` (L299–306, outside this diff), scope the List to this PS's instance. Deriving the filter from the parent (rather than `r.instanceID`) keeps it correct even in single-install mode:

```diff
-	var ctpList promoterv1alpha1.ChangeTransferPolicyList
-	err := r.List(ctx, &ctpList, client.InNamespace(ps.Namespace), client.MatchingLabels{
-		promoterv1alpha1.PromotionStrategyLabel: utils.KubeSafeLabel(ps.Name),
-	})
+	ctpSelector := map[string]string{
+		promoterv1alpha1.PromotionStrategyLabel: utils.KubeSafeLabel(ps.Name),
+	}
+	ctpSelector = utils.CopyInstanceIDLabelToMap(ps, ctpSelector)
+	var ctpList promoterv1alpha1.ChangeTransferPolicyList
+	err := r.List(ctx, &ctpList, client.InNamespace(ps.Namespace), client.MatchingLabels(ctpSelector))
```

⚠️ Migration caveat worth documenting: children created *before* this feature carry no label, so a label-filtered cleanup List will skip them. The docs' migration section should tell operators that children inherit the label "on their next reconcile" — which only works if the parent reconciles them at least once before the filter is relied on.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -493,7 +499,36 @@ func (r *ChangeTransferPolicyReconciler) SetupWithManager(ctx context.Context, m
...
+		if instanceID != "" {
+			var existing promoterv1alpha1.ChangeTransferPolicy
+			err := r.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, &existing)
+			if err != nil {
+				if k8s_errors.IsNotFound(err) {
+					// CTP was deleted between the caller's decision and this send. Drop silently.
+					return
+				}
+				// Transient cache error — log and proceed without filtering rather than
+				// silently dropping. Reconcile will short-circuit if the label is wrong.
```
zachaller: **[blocking] This comment claims a short-circuit that doesn't exist**

> `// Transient cache error — log and proceed without filtering rather than silently dropping. Reconcile will short-circuit if the label is wrong.`

There is no label check anywhere in `Reconcile`. On a transient cache error this forwards an enqueue into a reconcile that runs **full git clone/fetch/merge plus PR creation** on a potentially foreign CTP. The `Owns(&PullRequest{})` watch similarly enqueues this CTP with no instance filtering. Add the guard that makes the comment true:

```diff
 	if deleted, err := r.handleFinalizer(ctx, &ctp); err != nil || deleted {
 		return ctrl.Result{}, err
 	}
+
+	if !utils.MatchesInstanceID(&ctp, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", ctp.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		ctp = promoterv1alpha1.ChangeTransferPolicy{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

(requires an `instanceID string` struct field set in `SetupWithManager`; note the guard goes *after* `handleFinalizer` deliberately — or before it, but then a mislabeled deleting CTP keeps its finalizer forever; worth an explicit decision + comment either way)

internal/controller/changetransferpolicy_controller.go
```diff
@@ -511,18 +546,23 @@ func (r *ChangeTransferPolicyReconciler) SetupWithManager(ctx context.Context, m
...
-				// TODO: use a custom predicate to only trigger on the specific annotation change.
-				predicate.AnnotationChangedPredicate{},
+			builder.WithPredicates(predicate.And(
+				predicate.Or(
+					predicate.GenerationChangedPredicate{},
+					// Webhooks trigger reconciliations by bumping an annotation.
+					// TODO: use a custom predicate to only trigger on the specific annotation change.
+					predicate.AnnotationChangedPredicate{},
+				),
+				promoterpredicate.InstanceID(instanceID),
```
zachaller: 📍 *Applies to `setCommitStatusState` (L727–737), which isn't part of this PR's diff hunks.*

**[blocking — design decision] CommitStatus gating List is unfiltered; this is the unresolved thread with @crenshaw-dev**

`setCommitStatusState` lists CommitStatuses by key + SHA to decide whether a promotion gate passes. As written, a CommitStatus created by *another* install (same key, same SHA) gates — or worse, trips the `TooManyMatchingShaError` path when two installs both emit a status for the same SHA, wedging promotion for both.

Per the "fully independent installs" position, derive the filter from the parent CTP:

```diff
 		var csList promoterv1alpha1.CommitStatusList
-		// Find all the replicasets that match the commit status configured name and the sha of the hydrated commit
+		// Find all the CommitStatuses that match the configured key, the sha of the hydrated
+		// commit, and this CTP's instance id (no-op selector in single-install mode).
+		csSelector := map[string]string{
+			promoterv1alpha1.CommitStatusLabel: utils.KubeSafeLabel(status.Key),
+		}
+		csSelector = utils.CopyInstanceIDLabelToMap(ctp, csSelector)
 		err := r.List(ctx, &csList, &client.ListOptions{
-			LabelSelector: labels.SelectorFromSet(map[string]string{
-				promoterv1alpha1.CommitStatusLabel: utils.KubeSafeLabel(status.Key),
-			}),
+			LabelSelector: labels.SelectorFromSet(csSelector),
 			FieldSelector: fields.SelectorFromSet(map[string]string{
 				".spec.sha": targetCommitBranchState.Hydrated.Sha,
 			}),
 		})
```

Whichever way this lands ("filter everything" vs. "observe all for gating"), it should be an explicit, documented decision in `docs/multi-install.md` — right now the PR does neither consistently (webhook CTP lookups filter, gating lookups don't).

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1141,12 +1181,14 @@ func (r *ChangeTransferPolicyReconciler) creatOrUpdatePullRequest(ctx context.Co
 	}
 
 	// Build the apply configuration
+	prLabels := map[string]string{
+		promoterv1alpha1.PromotionStrategyLabel:    utils.KubeSafeLabel(ctp.Labels[promoterv1alpha1.PromotionStrategyLabel]),
+		promoterv1alpha1.ChangeTransferPolicyLabel: utils.KubeSafeLabel(ctp.Name),
+		promoterv1alpha1.EnvironmentLabel:          utils.KubeSafeLabel(ctp.Spec.ActiveBranch),
+	}
+	prLabels = utils.CopyInstanceIDLabelToMap(ctp, prLabels)
```
zachaller: **[important] PullRequest child Lists don't filter by instance id**

`ctpPullRequestListOptions` (L933–943; used by `setPullRequestState`, `handlePRFinalizerRemoval`, `handleCTPCleanupOnDelete`) and the inline duplicate in `mergePullRequests` (L1246–1253) list child PRs by PS/CTP/env labels only. Since this line makes CTP→PullRequest propagation work 👍, the child's instance label is guaranteed for new PRs — derive the filter from the parent:

```diff
 func ctpPullRequestListOptions(ctp *promoterv1alpha1.ChangeTransferPolicy) *client.ListOptions {
+	prSelector := map[string]string{
+		promoterv1alpha1.PromotionStrategyLabel:    utils.KubeSafeLabel(ctp.Labels[promoterv1alpha1.PromotionStrategyLabel]),
+		promoterv1alpha1.ChangeTransferPolicyLabel: utils.KubeSafeLabel(ctp.Name),
+		promoterv1alpha1.EnvironmentLabel:          utils.KubeSafeLabel(ctp.Spec.ActiveBranch),
+	}
+	prSelector = utils.CopyInstanceIDLabelToMap(ctp, prSelector)
 	return &client.ListOptions{
 		Namespace: ctp.Namespace,
-		LabelSelector: labels.SelectorFromSet(map[string]string{
-			promoterv1alpha1.PromotionStrategyLabel:    utils.KubeSafeLabel(ctp.Labels[promoterv1alpha1.PromotionStrategyLabel]),
-			promoterv1alpha1.ChangeTransferPolicyLabel: utils.KubeSafeLabel(ctp.Name),
-			promoterv1alpha1.EnvironmentLabel:          utils.KubeSafeLabel(ctp.Spec.ActiveBranch),
-		}),
+		LabelSelector: labels.SelectorFromSet(prSelector),
 	}
 }
```

`mergePullRequests` should call `ctpPullRequestListOptions(ctp)` instead of rebuilding the same selector inline — that removes the duplication and picks up the filter for free.

Same migration caveat as the PS cleanup List: PRs created before the upgrade lack the label until their parent CTP re-applies them once.

internal/controller/commitstatus_controller.go
```diff
@@ -126,8 +127,16 @@ func (r *CommitStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 func (r *CommitStatusReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
-	err := ctrl.NewControllerManagedBy(mgr).
-		For(&promoterv1alpha1.CommitStatus{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+	instanceID, err := r.SettingsMgr.GetInstanceIDDirect(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get InstanceID from ControllerConfiguration: %w", err)
+	}
+
+	err = ctrl.NewControllerManagedBy(mgr).
+		For(&promoterv1alpha1.CommitStatus{}, builder.WithPredicates(predicate.And(
```
zachaller: 📍 *Applies to `Reconcile` (~L85–94) and `triggerReconcileChangeTransferPolicy` (L209–252) — anchored here on the predicate wiring.*

**[important] Reconcile guard + four unfiltered CTP Lists**

1. Guard (this Reconcile calls the SCM provider's `Set()` — a cross-instance reconcile writes statuses to the SCM twice from two installs):

```diff
 	err = r.Get(ctx, req.NamespacedName, &cs, &client.GetOptions{})
 	if err != nil {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&cs, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", cs.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		cs = promoterv1alpha1.CommitStatus{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

2. `triggerReconcileChangeTransferPolicy` does four cluster-wide CTP Lists by SHA with no label selector, then relies entirely on the enqueueFunc's sender-side gate to drop foreign CTPs. Filter at the source (apply to all four `ListOptions`):

```diff
 	err := r.List(ctx, ctpListActiveOldSha, &client.ListOptions{
 		FieldSelector: fields.SelectorFromSet(map[string]string{
 			".status.active.hydrated.sha": oldSha,
 		}),
+		LabelSelector: utils.InstanceIDSelector(r.instanceID),
 	})
```

This also cuts cache-listing cost in big multi-install clusters instead of fetching every install's CTPs on every CommitStatus reconcile.

internal/controller/argocdcommitstatus_controller.go
```diff
@@ -551,7 +559,7 @@ func (r *ArgoCDCommitStatusReconciler) SetupWithManager(ctx context.Context, mcM
 		Watches(&promoterv1alpha1.PromotionStrategy{}, enqueueArgoCDCommitStatusForPromotionStrategy(mcMgr),
```
zachaller, line 559: 📍 *Partially anchored — also applies to `lookupArgoCDCommitStatusFromArgoCDApplication` (L477–482) and `Reconcile` (L120–132), outside this diff's hunks.*

**[blocking] The unfiltered `Watches(Application)` funnels into an unfiltered cluster-wide List — instance A reconciles instance B's ArgoCDCommitStatus on every app health flap**

Design decision #3 (don't filter remote Application watches) is fine, but it makes the downstream filtering mandatory, and it's missing:

1. `lookupArgoCDCommitStatusFromArgoCDApplication` lists **all** ArgoCDCommitStatuses cluster-wide and enqueues the first selector match regardless of instance. Map-function requests bypass `For(...)` predicates entirely. Pass the instanceID in and filter:

```diff
-func lookupArgoCDCommitStatusFromArgoCDApplication(mgr mcmanager.Manager) mchandler.TypedEventHandlerFunc[client.Object, mcreconcile.Request] {
+func lookupArgoCDCommitStatusFromArgoCDApplication(mgr mcmanager.Manager, instanceID string) mchandler.TypedEventHandlerFunc[client.Object, mcreconcile.Request] {
 ...
-			if err := mgr.GetLocalManager().GetClient().List(ctx, &argoCDCommitStatusList, &client.ListOptions{}); err != nil {
+			if err := mgr.GetLocalManager().GetClient().List(ctx, &argoCDCommitStatusList, &client.ListOptions{
+				LabelSelector: utils.InstanceIDSelector(instanceID),
+			}); err != nil {
```

2. Same for `enqueueArgoCDCommitStatusForPromotionStrategy(mcMgr)` — add the `instanceID` param and `utils.InstanceIDListOption(instanceID)` to its List. Update both call sites in `SetupWithManager`.

3. Reconcile guard (add an `instanceID string` field next to `watchLocalApplications` and set it where the value is fetched):

```diff
 	err = r.localClient.Get(ctx, req.NamespacedName, &argoCDCommitStatus, &client.GetOptions{})
 	if err != nil {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&argoCDCommitStatus, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", argoCDCommitStatus.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		argoCDCommitStatus = promoterv1alpha1.ArgoCDCommitStatus{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

Without #3, every Application event from the unfiltered watch runs this reconcile's git `ls-remote` + CommitStatus SSA for foreign resources — the exact cross-instance race this PR exists to prevent. `TestArgoCDCommitStatusOptionB` should gain an assertion that the *handler* List is filtered, since the watch itself intentionally isn't.

internal/controller/pullrequest_controller.go
```diff
@@ -335,10 +336,18 @@ func (r *PullRequestReconciler) SetupWithManager(ctx context.Context, mgr ctrl.M
...
 	err = ctrl.NewControllerManagedBy(mgr).
-		For(&promoterv1alpha1.PullRequest{}, builder.WithPredicates(predicate.Or(
-			predicate.GenerationChangedPredicate{},
-			pullRequestDeletionFinalizerLengthChangedPredicate(),
+		For(&promoterv1alpha1.PullRequest{}, builder.WithPredicates(predicate.And(
+			predicate.Or(
+				predicate.GenerationChangedPredicate{},
+				pullRequestDeletionFinalizerLengthChangedPredicate(),
+			),
+			promoterpredicate.InstanceID(instanceID),
```
zachaller: 📍 *Applies to `Reconcile` (~L85–94), outside this diff's hunks — anchored here on the predicate wiring.*

**[important] Reconcile-time guard for the simple reconcilers (defense in depth)**

Predicates are an optimization, not a security boundary — anything that lands in the queue gets reconciled. For consistency with the guarded controllers, add the same check here (and identically in `revertcommit_controller.go`, `scmprovider_controller.go`, `clusterscmprovider_controller.go`, `gitrepository_controller.go`), storing `r.instanceID` in `SetupWithManager` where it's already fetched:

```diff
 	if err := r.Get(ctx, req.NamespacedName, &pr); err != nil {
 		if errors.IsNotFound(err) {
 			...
 		}
 		return ctrl.Result{}, fmt.Errorf("failed to get PullRequest: %w", err)
 	}
+
+	if !utils.MatchesInstanceID(&pr, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", pr.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		pr = promoterv1alpha1.PullRequest{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

PullRequest is the highest-stakes of the five: a cross-instance reconcile here merges/closes PRs on the SCM. The others are cheap reconciles but the guard keeps the invariant uniform and lets the AST test assert "every reconciler except ControllerConfiguration guards" as a single rule.

internal/utils/labels.go
zachaller, line 20: **[blocking] Add lookup-side helpers so every reconcile/List can enforce the instance-id contract**

`CopyInstanceIDLabel`/`CopyInstanceIDLabelToMap` cover the *write* side (label propagation), but there's nothing for the *read* side. The predicates only gate watch events; `Owns()` watches, `EnqueueRequestsFromMapFunc` handlers, and `WatchesRawSource` all enqueue requests that never pass through `For(...)`'s predicates, so each `Reconcile` and each `List`/`Get` of child resources needs its own check. Suggest adding to this file (the other review comments reference these helpers):

```go
// MatchesInstanceID reports whether obj belongs to the install identified by id.
// Empty id preserves single-install semantics: every object matches.
func MatchesInstanceID(obj client.Object, id string) bool {
	return id == "" || obj.GetLabels()[promoterv1alpha1.InstanceIDLabel] == id
}

// InstanceIDSelector returns a labels.Selector scoping List calls to this
// install's instance id, or labels.Everything() when id is empty.
// For use with client.ListOptions{LabelSelector: ...}.
func InstanceIDSelector(id string) labels.Selector {
	if id == "" {
		return labels.Everything()
	}
	return labels.SelectorFromSet(labels.Set{promoterv1alpha1.InstanceIDLabel: id})
}

// InstanceIDListOption is InstanceIDSelector as a variadic client.ListOption,
// for List(ctx, list, client.InNamespace(...), utils.InstanceIDListOption(id)) call sites.
func InstanceIDListOption(id string) client.ListOption {
	return client.MatchingLabelsSelector{Selector: InstanceIDSelector(id)}
}
```

(needs `"k8s.io/apimachinery/pkg/labels"` import)

The reconcile-time guard pattern is the same everywhere: store the already-fetched `instanceID` on the reconciler struct in `SetupWithManager` (`r.instanceID = instanceID`), then check after the `Get` in `Reconcile`. One subtlety: the deferred `utils.HandleReconciliationResult` would otherwise SSA-patch status on the foreign object — and since both installs use the **same field-manager name**, that's a cross-instance ownership fight. Zero the object before returning so the defer's existing `GetName()==""` early-exit kicks in.

internal/controller/timedcommitstatus_controller.go
zachaller: 📍 *Applies to `upsertCommitStatus` (L301–303), which isn't part of this PR's diff hunks — anchored here on the predicate wiring instead.*

**[blocking] Missing instance-id propagation — TimedCommitStatus gates silently break in multi-install mode**

This is one of two child-creation sites that didn't get `CopyInstanceIDLabelToMap` (the other is WebRequestCommitStatus). The CommitStatus created in `upsertCommitStatus` carries no instance-id label, so this install's own `CommitStatusReconciler` predicate **drops its events** — the status never reaches the SCM and the time gate never passes. This contradicts the PR description's "child resources automatically inherit the label" claim and `docs/multi-install.md`.

```diff
-	// Build the apply configuration
+	// Build the apply configuration. Compose main's standard-labels helper with
+	// our instance-id propagation (ARGO-3085).
+	commitStatusLabels := utils.CopyInstanceIDLabelToMap(
+		tcs,
+		utils.CommitStatusStandardLabels(tcs, branch, key),
+	)
 	commitStatusApply := acv1alpha1.CommitStatus(commitStatusName, tcs.Namespace).
-		WithLabels(utils.CommitStatusStandardLabels(tcs, branch, key)).
+		WithLabels(commitStatusLabels).
```

Please also extend `setup_predicate_chain_test.go` (or a sibling test) to assert all seven CommitStatus/child creation sites call `CopyInstanceIDLabel*` — the AST test only proves `SetupWithManager` wiring and cannot catch exactly this class of regression.

internal/controller/timedcommitstatus_controller.go
zachaller: 📍 *Applies to `Reconcile` (after the Get, ~L80–88) and `enqueueTimedCommitStatusForPromotionStrategy` (L374–377) — anchored here on the Watches wiring.*

**[blocking] Reconcile guard + unfiltered List in the PromotionStrategy map handler**

Two gaps in this controller:

1. Requests produced by `enqueueTimedCommitStatusForPromotionStrategy` **do not pass through the `For(...)` predicates** (map-function requests are direct queue inserts). The `Watches` predicate filters the *source* PS event, but the List inside the handler enqueues *every* TCS in the namespace — including foreign-instance ones:

```diff
 		// List all TimedCommitStatus resources in the same namespace
 		var tcsList promoterv1alpha1.TimedCommitStatusList
-		if err := r.List(ctx, &tcsList, client.InNamespace(ps.Namespace)); err != nil {
+		if err := r.List(ctx, &tcsList, client.InNamespace(ps.Namespace), utils.InstanceIDListOption(r.instanceID)); err != nil {
```

2. Defense-in-depth guard in `Reconcile` (requires an `instanceID string` field on the struct, set in `SetupWithManager`):

```diff
 	err = r.Get(ctx, req.NamespacedName, &tcs, &client.GetOptions{})
 	if err != nil {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&tcs, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", tcs.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		tcs = promoterv1alpha1.TimedCommitStatus{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

Optionally also verify the referenced PS at step 2 (`r.Get(ctx, psKey, &ps)`): a TCS labeled for instance A referencing a PS labeled for instance B is a misconfiguration worth surfacing as an error rather than silently acting on it.

internal/controller/webrequestcommitstatus_controller.go
zachaller: 📍 *Applies to `upsertCommitStatus` (L486–487), which isn't part of this PR's diff hunks — anchored here on the predicate wiring instead.*

**[blocking] Missing instance-id propagation — same bug as TimedCommitStatus**

`upsertCommitStatus` here doesn't propagate the label, so in multi-install mode web-request gate CommitStatuses are invisible to this install's CommitStatus reconciler and never reach the SCM:

```diff
+	// Compose main's standard-labels helper with our instance-id propagation (ARGO-3085).
+	commitStatusLabels := utils.CopyInstanceIDLabelToMap(
+		wrcs,
+		utils.CommitStatusStandardLabels(wrcs, branch, wrcs.Spec.Key),
+	)
 	commitStatusApply := acv1alpha1.CommitStatus(commitStatusName, wrcs.Namespace).
-		WithLabels(utils.CommitStatusStandardLabels(wrcs, branch, wrcs.Spec.Key)).
+		WithLabels(commitStatusLabels).
```

internal/controller/webrequestcommitstatus_controller.go
zachaller: 📍 *Applies to `Reconcile` (~L140–148) and `enqueueWebRequestCommitStatusForPromotionStrategy` (L518–521) — anchored here on the Watches wiring.*

**[blocking] Reconcile guard + unfiltered List in map handler**

Same two gaps as TimedCommitStatus. This controller already has the `skipStatusWrite` flag, so the guard is cleaner here — no object-zeroing needed:

```diff
 	if r.rvTracker.IsCacheStale(req.NamespacedName, wrcs.ResourceVersion) {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&wrcs, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", wrcs.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		skipStatusWrite = true
+		return ctrl.Result{}, nil
+	}
```

And in `enqueueWebRequestCommitStatusForPromotionStrategy`:

```diff
-		if err := r.List(ctx, &wrcsList, client.InNamespace(ps.Namespace)); err != nil {
+		if err := r.List(ctx, &wrcsList, client.InNamespace(ps.Namespace), utils.InstanceIDListOption(r.instanceID)); err != nil {
```

(both require storing `r.instanceID` in `SetupWithManager`; the HTTP side effects this controller fires make the reconcile guard especially important — a cross-instance reconcile here doesn't just waste CPU, it can re-fire trigger webhooks)

zachaller (commented):
Reviewing against the multi-install partitioning contract: **(1) every reconcile must respect its instance ID, and (2) every list/get of child resources must filter by instance ID.**

Summary of the inline comments:

- **2 critical propagation bugs** — `TimedCommitStatus` and `WebRequestCommitStatus` don't propagate the instance-id label to the CommitStatuses they create, so in multi-install mode those gates silently never pass (this install's own CommitStatus predicate drops their events).
- **Reconcile-time guards missing** — predicates only gate watch events. `Owns()` watches, `EnqueueRequestsFromMapFunc` handlers, and `WatchesRawSource` channel sends enqueue requests that bypass the `For(...)` predicates, so each `Reconcile` needs a label check. (The comment at `changetransferpolicy_controller.go` L519–520 claims such a short-circuit exists — it doesn't.)
- **Unfiltered Lists** — the only Lists that gained an instance-id selector are the webhook receiver's CTP lookups. The map-handler Lists, CTP gating/PR lookups, CommitStatus→CTP fan-out, and orphan-cleanup Lists are all unfiltered.

📍 Some comments are anchored to nearby in-diff lines because their actual targets aren't part of this PR's hunks; each one starts with a pointer to the real location.

Two follow-ups not tied to a line:

1. **Test gap**: `setup_predicate_chain_test.go` proves `SetupWithManager` wiring only. Please also assert (a) every child-creation site composes `CopyInstanceIDLabel*` (would have caught both propagation bugs) and (b) every map-function List passes an instance-id list option.
2. **Docs**: `docs/multi-install.md` should describe the enforcement layers explicitly (predicates → reconcile guards → filtered Lists/Gets) and document the gating-semantics decision (do cross-instance CommitStatuses gate promotions?) once the open thread with @crenshaw-dev is resolved.

internal/controller/gitcommitstatus_controller.go
zachaller: 📍 *Applies to `Reconcile` (~L90–99) and `enqueueGitCommitStatusForPromotionStrategy` (L458–462) — anchored here on the Watches wiring.*

**[blocking] Reconcile guard + unfiltered List in map handler**

Same pattern as the other two gate controllers (the label propagation at L418 is already correct here 👍):

```diff
 	err = r.Get(ctx, req.NamespacedName, &gcs, &client.GetOptions{})
 	if err != nil {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&gcs, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", gcs.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		gcs = promoterv1alpha1.GitCommitStatus{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

```diff
 		// List all GitCommitStatus resources in the same namespace
 		var gcsList promoterv1alpha1.GitCommitStatusList
-		if err := r.List(ctx, &gcsList, client.InNamespace(ps.Namespace)); err != nil {
+		if err := r.List(ctx, &gcsList, client.InNamespace(ps.Namespace), utils.InstanceIDListOption(r.instanceID)); err != nil {
```

internal/controller/promotionstrategy_controller.go
zachaller, line 200: **[blocking] `Owns(ChangeTransferPolicy)` bypasses the InstanceID predicate — guard needed; cleanup List unfiltered**

`builder.WithPredicates` inside `For(...)` applies only to that watch. This `Owns(&ChangeTransferPolicy{})` carries no InstanceID predicate, so any CTP event (e.g. a pre-migration unlabeled child) maps to its owner PS and enqueues it — and `Reconcile` will then upsert CTPs/CommitStatuses for a foreign PromotionStrategy. Guard after the Get (~L95–109, outside this diff):

```diff
 	// If the resource is being deleted, stop reconciling immediately without requeuing
 	if !ps.DeletionTimestamp.IsZero() {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&ps, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", ps.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		ps = promoterv1alpha1.PromotionStrategy{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

(Alternative: add `builder.WithPredicates(instanceIDPredicate)` to the `Owns(...)` clause — but the guard is still needed for full coverage, e.g. the cross-controller enqueue paths.)

In `cleanupOrphanedChangeTransferPolicies` (L299–306, outside this diff), scope the List to this PS's instance. Deriving the filter from the parent (rather than `r.instanceID`) keeps it correct even in single-install mode:

```diff
-	var ctpList promoterv1alpha1.ChangeTransferPolicyList
-	err := r.List(ctx, &ctpList, client.InNamespace(ps.Namespace), client.MatchingLabels{
-		promoterv1alpha1.PromotionStrategyLabel: utils.KubeSafeLabel(ps.Name),
-	})
+	ctpSelector := map[string]string{
+		promoterv1alpha1.PromotionStrategyLabel: utils.KubeSafeLabel(ps.Name),
+	}
+	ctpSelector = utils.CopyInstanceIDLabelToMap(ps, ctpSelector)
+	var ctpList promoterv1alpha1.ChangeTransferPolicyList
+	err := r.List(ctx, &ctpList, client.InNamespace(ps.Namespace), client.MatchingLabels(ctpSelector))
```

⚠️ Migration caveat worth documenting: children created *before* this feature carry no label, so a label-filtered cleanup List will skip them. The docs' migration section should tell operators that children inherit the label "on their next reconcile" — which only works if the parent reconciles them at least once before the filter is relied on.

internal/controller/changetransferpolicy_controller.go
zachaller: **[blocking] This comment claims a short-circuit that doesn't exist**

> `// Transient cache error — log and proceed without filtering rather than silently dropping. Reconcile will short-circuit if the label is wrong.`

There is no label check anywhere in `Reconcile`. On a transient cache error this forwards an enqueue into a reconcile that runs **full git clone/fetch/merge plus PR creation** on a potentially foreign CTP. The `Owns(&PullRequest{})` watch similarly enqueues this CTP with no instance filtering. Add the guard that makes the comment true:

```diff
 	if deleted, err := r.handleFinalizer(ctx, &ctp); err != nil || deleted {
 		return ctrl.Result{}, err
 	}
+
+	if !utils.MatchesInstanceID(&ctp, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", ctp.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		ctp = promoterv1alpha1.ChangeTransferPolicy{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

(requires an `instanceID string` struct field set in `SetupWithManager`; note the guard goes *after* `handleFinalizer` deliberately — or before it, but then a mislabeled deleting CTP keeps its finalizer forever; worth an explicit decision + comment either way)

internal/controller/changetransferpolicy_controller.go
zachaller: 📍 *Applies to `setCommitStatusState` (L727–737), which isn't part of this PR's diff hunks.*

**[blocking — design decision] CommitStatus gating List is unfiltered; this is the unresolved thread with @crenshaw-dev**

`setCommitStatusState` lists CommitStatuses by key + SHA to decide whether a promotion gate passes. As written, a CommitStatus created by *another* install (same key, same SHA) gates — or worse, trips the `TooManyMatchingShaError` path when two installs both emit a status for the same SHA, wedging promotion for both.

Per the "fully independent installs" position, derive the filter from the parent CTP:

```diff
 		var csList promoterv1alpha1.CommitStatusList
-		// Find all the replicasets that match the commit status configured name and the sha of the hydrated commit
+		// Find all the CommitStatuses that match the configured key, the sha of the hydrated
+		// commit, and this CTP's instance id (no-op selector in single-install mode).
+		csSelector := map[string]string{
+			promoterv1alpha1.CommitStatusLabel: utils.KubeSafeLabel(status.Key),
+		}
+		csSelector = utils.CopyInstanceIDLabelToMap(ctp, csSelector)
 		err := r.List(ctx, &csList, &client.ListOptions{
-			LabelSelector: labels.SelectorFromSet(map[string]string{
-				promoterv1alpha1.CommitStatusLabel: utils.KubeSafeLabel(status.Key),
-			}),
+			LabelSelector: labels.SelectorFromSet(csSelector),
 			FieldSelector: fields.SelectorFromSet(map[string]string{
 				".spec.sha": targetCommitBranchState.Hydrated.Sha,
 			}),
 		})
```

Whichever way this lands ("filter everything" vs. "observe all for gating"), it should be an explicit, documented decision in `docs/multi-install.md` — right now the PR does neither consistently (webhook CTP lookups filter, gating lookups don't).

internal/controller/changetransferpolicy_controller.go
zachaller: **[important] PullRequest child Lists don't filter by instance id**

`ctpPullRequestListOptions` (L933–943; used by `setPullRequestState`, `handlePRFinalizerRemoval`, `handleCTPCleanupOnDelete`) and the inline duplicate in `mergePullRequests` (L1246–1253) list child PRs by PS/CTP/env labels only. Since this line makes CTP→PullRequest propagation work 👍, the child's instance label is guaranteed for new PRs — derive the filter from the parent:

```diff
 func ctpPullRequestListOptions(ctp *promoterv1alpha1.ChangeTransferPolicy) *client.ListOptions {
+	prSelector := map[string]string{
+		promoterv1alpha1.PromotionStrategyLabel:    utils.KubeSafeLabel(ctp.Labels[promoterv1alpha1.PromotionStrategyLabel]),
+		promoterv1alpha1.ChangeTransferPolicyLabel: utils.KubeSafeLabel(ctp.Name),
+		promoterv1alpha1.EnvironmentLabel:          utils.KubeSafeLabel(ctp.Spec.ActiveBranch),
+	}
+	prSelector = utils.CopyInstanceIDLabelToMap(ctp, prSelector)
 	return &client.ListOptions{
 		Namespace: ctp.Namespace,
-		LabelSelector: labels.SelectorFromSet(map[string]string{
-			promoterv1alpha1.PromotionStrategyLabel:    utils.KubeSafeLabel(ctp.Labels[promoterv1alpha1.PromotionStrategyLabel]),
-			promoterv1alpha1.ChangeTransferPolicyLabel: utils.KubeSafeLabel(ctp.Name),
-			promoterv1alpha1.EnvironmentLabel:          utils.KubeSafeLabel(ctp.Spec.ActiveBranch),
-		}),
+		LabelSelector: labels.SelectorFromSet(prSelector),
 	}
 }
```

`mergePullRequests` should call `ctpPullRequestListOptions(ctp)` instead of rebuilding the same selector inline — that removes the duplication and picks up the filter for free.

Same migration caveat as the PS cleanup List: PRs created before the upgrade lack the label until their parent CTP re-applies them once.

internal/controller/commitstatus_controller.go
zachaller: 📍 *Applies to `Reconcile` (~L85–94) and `triggerReconcileChangeTransferPolicy` (L209–252) — anchored here on the predicate wiring.*

**[important] Reconcile guard + four unfiltered CTP Lists**

1. Guard (this Reconcile calls the SCM provider's `Set()` — a cross-instance reconcile writes statuses to the SCM twice from two installs):

```diff
 	err = r.Get(ctx, req.NamespacedName, &cs, &client.GetOptions{})
 	if err != nil {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&cs, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", cs.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		cs = promoterv1alpha1.CommitStatus{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

2. `triggerReconcileChangeTransferPolicy` does four cluster-wide CTP Lists by SHA with no label selector, then relies entirely on the enqueueFunc's sender-side gate to drop foreign CTPs. Filter at the source (apply to all four `ListOptions`):

```diff
 	err := r.List(ctx, ctpListActiveOldSha, &client.ListOptions{
 		FieldSelector: fields.SelectorFromSet(map[string]string{
 			".status.active.hydrated.sha": oldSha,
 		}),
+		LabelSelector: utils.InstanceIDSelector(r.instanceID),
 	})
```

This also cuts cache-listing cost in big multi-install clusters instead of fetching every install's CTPs on every CommitStatus reconcile.

internal/controller/argocdcommitstatus_controller.go
zachaller, line 559: 📍 *Partially anchored — also applies to `lookupArgoCDCommitStatusFromArgoCDApplication` (L477–482) and `Reconcile` (L120–132), outside this diff's hunks.*

**[blocking] The unfiltered `Watches(Application)` funnels into an unfiltered cluster-wide List — instance A reconciles instance B's ArgoCDCommitStatus on every app health flap**

Design decision #3 (don't filter remote Application watches) is fine, but it makes the downstream filtering mandatory, and it's missing:

1. `lookupArgoCDCommitStatusFromArgoCDApplication` lists **all** ArgoCDCommitStatuses cluster-wide and enqueues the first selector match regardless of instance. Map-function requests bypass `For(...)` predicates entirely. Pass the instanceID in and filter:

```diff
-func lookupArgoCDCommitStatusFromArgoCDApplication(mgr mcmanager.Manager) mchandler.TypedEventHandlerFunc[client.Object, mcreconcile.Request] {
+func lookupArgoCDCommitStatusFromArgoCDApplication(mgr mcmanager.Manager, instanceID string) mchandler.TypedEventHandlerFunc[client.Object, mcreconcile.Request] {
 ...
-			if err := mgr.GetLocalManager().GetClient().List(ctx, &argoCDCommitStatusList, &client.ListOptions{}); err != nil {
+			if err := mgr.GetLocalManager().GetClient().List(ctx, &argoCDCommitStatusList, &client.ListOptions{
+				LabelSelector: utils.InstanceIDSelector(instanceID),
+			}); err != nil {
```

2. Same for `enqueueArgoCDCommitStatusForPromotionStrategy(mcMgr)` — add the `instanceID` param and `utils.InstanceIDListOption(instanceID)` to its List. Update both call sites in `SetupWithManager`.

3. Reconcile guard (add an `instanceID string` field next to `watchLocalApplications` and set it where the value is fetched):

```diff
 	err = r.localClient.Get(ctx, req.NamespacedName, &argoCDCommitStatus, &client.GetOptions{})
 	if err != nil {
 		...
 	}
+
+	if !utils.MatchesInstanceID(&argoCDCommitStatus, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", argoCDCommitStatus.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		argoCDCommitStatus = promoterv1alpha1.ArgoCDCommitStatus{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

Without #3, every Application event from the unfiltered watch runs this reconcile's git `ls-remote` + CommitStatus SSA for foreign resources — the exact cross-instance race this PR exists to prevent. `TestArgoCDCommitStatusOptionB` should gain an assertion that the *handler* List is filtered, since the watch itself intentionally isn't.

internal/controller/pullrequest_controller.go
zachaller: 📍 *Applies to `Reconcile` (~L85–94), outside this diff's hunks — anchored here on the predicate wiring.*

**[important] Reconcile-time guard for the simple reconcilers (defense in depth)**

Predicates are an optimization, not a security boundary — anything that lands in the queue gets reconciled. For consistency with the guarded controllers, add the same check here (and identically in `revertcommit_controller.go`, `scmprovider_controller.go`, `clusterscmprovider_controller.go`, `gitrepository_controller.go`), storing `r.instanceID` in `SetupWithManager` where it's already fetched:

```diff
 	if err := r.Get(ctx, req.NamespacedName, &pr); err != nil {
 		if errors.IsNotFound(err) {
 			...
 		}
 		return ctrl.Result{}, fmt.Errorf("failed to get PullRequest: %w", err)
 	}
+
+	if !utils.MatchesInstanceID(&pr, r.instanceID) {
+		logger.V(4).Info("Skipping reconcile: resource belongs to another instance",
+			"want", r.instanceID, "got", pr.GetLabels()[promoterv1alpha1.InstanceIDLabel])
+		pr = promoterv1alpha1.PullRequest{} // prevent deferred status write on a foreign object
+		return ctrl.Result{}, nil
+	}
```

PullRequest is the highest-stakes of the five: a cross-instance reconcile here merges/closes PRs on the SCM. The others are cheap reconciles but the guard keeps the invariant uniform and lets the AST test assert "every reconciler except ControllerConfiguration guards" as a single rule.

crenshaw-dev:
## Cache-first multi-install rewrite (ready for re-review)

Rebuilt `ARGO-3085-instance-id` on current `main` with a cache-first partitioning design that addresses @zachaller's Jun 10 review without per-controller predicate/guard/List-filter sprawl.

### Enforcement model

1. **Read path:** `ControllerConfiguration.spec.instanceID` is bootstrap-read before `mcmanager.New`; `cache.ByObject` label selectors scope all reconciled Promoter CRDs (empty `instanceID` = single-install, unchanged behavior).
2. **Write path:** `utils.CopyInstanceIDLabel*` propagates `promoter.argoproj.io/instance-id` at all 7 child-creation sites, including **TimedCommitStatus** and **WebRequestCommitStatus** `upsertCommitStatus`.
3. **Foreign reconciles:** stale enqueues / unfiltered Argo CD `Application` watches → `Get` returns `NotFound` → existing no-op path (no `SkipStatusApply` needed).

### What was removed vs the prior PR approach

- No `promoterpredicate.InstanceID` on every `For()`/`Watches()`
- No CTP channel-sender filter, webhook `List` filters, reconcile guards, or per-reconciler `instanceID` fields
- `setup_predicate_chain_test` InstanceID AST tests replaced with cache-options + label-propagation AST tests

### Gating semantics

Fully independent installs: cached `List` of `CommitStatus` only sees this install's objects when `instanceID` is set.

### Docs

- New `docs/multi-install.md` (cache-first architecture, migration, gating)
- Updated `developing-a-commitstatus.md`, `debugging/labels.md`, `multi-tenancy.md`, `contributing/index.md`, `gating-promotions/index.md`

### Commits (4)

1. `feat(cache): bootstrap instanceID read and ByObject cache filtering`
2. `feat(controller): propagate instance-id label to child resources`
3. `test: cache options and label propagation AST assertions`
4. `docs: document cache-first multi-install partitioning`

**Note:** Branch push to `pbhatnagar-oss/ARGO-3085-instance-id` is pending (local commits on `crenshaw-dev` machine). CI will run after push.

@zachaller — please re-review when the branch is updated.

config/config/controllerconfiguration.yaml
```diff
@@ -7,6 +7,11 @@ metadata:
     app.kubernetes.io/managed-by: kustomize
   name: controller-configuration
 spec:
+  # InstanceID scopes this install to resources carrying the matching
+  # promoter.argoproj.io/instance-id label. Empty means reconcile everything
```
zachaller: I think emtpy should reconcile resources without an instance id only?

crenshaw-dev:
Closing in favor of PR that was built on to of this one. Thanks @pbhatnagar-oss!

PR #1497, opened 2026-05-29
fix(ctp): update PR merge sha after conflict (#1496)

### LINKED ISSUE #1496: Conflict-resolved promotions briefly fail with a stale PR mergeSha
### What happens

When the active branch and the proposed branch end up with conflicting changes, GitOps Promoter resolves the conflict by adding a merge commit on the proposed branch and then opens / updates the PullRequest as usual. With auto-merge on, the PR is immediately flipped to `merged` and the SCM is asked to complete the merge.

That first SCM merge attempt is rejected, with an error along the lines of `source branch HEAD SHA "..." does not match expected merge SHA "..."`. Users see a `Warning` event on the PullRequest (and on the ChangeTransferPolicy via `PullRequestNotReady`), and at least one extra SCM merge call is consumed per conflict-resolved promotion. The merge eventually succeeds on a follow-up reconcile, so the system recovers on its own, but the noise and wasted API calls are visible.

### When it shows up most

Any time the controller has to run conflict resolution before a PR merges.

### Symptom users notice

- A `Warning` / `ReconciliationError` event on the PR with `does not match expected merge SHA` right after a conflict was resolved.
- A brief `PullRequestNotReady` condition / event on the ChangeTransferPolicy referencing the same error.
- An extra SCM `Merge` API call per affected promotion (which can matter against rate-limited SCM APIs at scale).

The promotion still completes on the next reconcile, so this is correctness-OK but noisy and wasteful.

crenshaw-dev:
Fixes #1496

### DIFF
--- a/internal/controller/changetransferpolicy_controller_test.go
+++ b/internal/controller/changetransferpolicy_controller_test.go
@@ -21,9 +21,11 @@ import (
 	_ "embed"
 	"fmt"
 	"os"
+	"path"
 	"strings"
 
 	promoterv1alpha1 "github.com/argoproj-labs/gitops-promoter/api/v1alpha1"
+	"github.com/argoproj-labs/gitops-promoter/internal/scms/fake"
 	promoterConditions "github.com/argoproj-labs/gitops-promoter/internal/types/conditions"
 	"github.com/argoproj-labs/gitops-promoter/internal/types/constants"
 	"github.com/argoproj-labs/gitops-promoter/internal/utils"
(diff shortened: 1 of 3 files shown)


internal/controller/changetransferpolicy_controller_test.go
```diff
@@ -913,6 +915,129 @@ var _ = Describe("ChangeTransferPolicy Controller", func() {
...
+		//   5. Eventually a follow-up CTP reconcile (triggered by the PR Owns watch) re-derives
+		//      Status.Proposed from the now-resolved tip and updates PR.Spec.MergeSha to S_new,
+		//      after which the next merge attempt succeeds and the active branch advances.
+		//
+		// The merge does eventually land, but the bug burns at least one extra
+		// SCM merge call per conflict-resolved promotion. With activePath promotions
+		// conflicting routinely on shared root files, that extra call multiplies. The fix is
+		// to short-circuit the rest of this reconcile when MergeWithOursStrategy rewrites the
+		// proposed branch and requeue immediately so the next reconcile creates/updates the PR
+		// with the correct mergeSha on the very next attempt.
```
crenshaw-dev, line 946:
```suggestion
		// The merge does eventually land, but the bug burns at least one extra
		// SCM merge call per conflict-resolved promotion. The fix is
		// to short-circuit the rest of this reconcile when MergeWithOursStrategy rewrites the
		// proposed branch and requeue immediately so the next reconcile creates/updates the PR
		// with the correct mergeSha on the very next attempt.
```

zachaller (approved):

PR #1511, opened 2026-05-30
fix: keep embedded dashboard assets present in fresh checkouts

immanuwell:
Fresh checkout is kinda busted for Go builds.

`ui/web/embed.go` embeds `ui/web/static`, but that dir is ignored and missing until `make build-dashboard` runs. So a plain fresh-clone `go build ./cmd/main.go` blows up.

This keeps a tiny bootstrap `index.html` and `assets/placeholder.txt` in git, and adds a small test. I also checked `npm run build:embed`, still fine.

Repro:
```bash
git clone git@github.com:argoproj-labs/gitops-promoter.git
cd gitops-promoter
go build ./cmd/main.go
```

Before:
```text
ui/web/embed.go:7:12: pattern static: no matching files found
```

After this branch, the same command works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,6 +89,12 @@ yarn-error.log*
 
 # Built UI static files
 ui/web/static/
+!ui/web/static/
+ui/web/static/*
+!ui/web/static/index.html
+!ui/web/static/assets/
+ui/web/static/assets/*
+!ui/web/static/assets/placeholder.txt
 
 # MkDocs build output
 mkdocs-lint.log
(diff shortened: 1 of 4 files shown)


crenshaw-dev:
@immanuwell thanks for the PR!

A downside to checking this into git is that every time someone builds the dashboard locally, the built dashboard is gonna have staged changes the user will have to ignore.

Should we instead just improve the make targets to make sure prerequisites run before go build? Or maybe just check in the actual built dashboard?

crenshaw-dev:
Closing for now, feel free to reopen or respond, and we'll go from there!

PR #1512, opened 2026-05-31
feat: APIService for dashboard

zachaller:


api/dashboard/dashboard/types.go
```diff
@@ -0,0 +1,55 @@
...
+	ChangeTransferPolicies   []promoterv1alpha1.ChangeTransferPolicy
+	PullRequests             []promoterv1alpha1.PullRequest
+	CommitStatuses           []promoterv1alpha1.CommitStatus
+	ArgoCDCommitStatuses     []promoterv1alpha1.ArgoCDCommitStatus
+	GitCommitStatuses        []promoterv1alpha1.GitCommitStatus
+	TimedCommitStatuses      []promoterv1alpha1.TimedCommitStatus
+	WebRequestCommitStatuses []promoterv1alpha1.WebRequestCommitStatus
+	GitRepository            *promoterv1alpha1.GitRepository
+	ScmProvider              *promoterv1alpha1.ScmProvider
+	ClusterScmProvider       *promoterv1alpha1.ClusterScmProvider
```
crenshaw-dev: I wonder if it would still make sense to have a spec/status split, just in case we want to later provide a way for the caller to parameterize the request. For example, maybe they want to exclude certain fields to keep the payload small. That could be specified via spec.

zachaller: I would need to look into what kind of parameterization kubectl would allow.

docs/dashboard-apiserver.md
```diff
@@ -0,0 +1,135 @@
...
+
+Each bundle contains the `PromotionStrategy`, its `ChangeTransferPolicy`,
+`PullRequest`, and `CommitStatus` children, the four commit-status manager kinds
+(`ArgoCDCommitStatus`, `GitCommitStatus`, `TimedCommitStatus`,
+`WebRequestCommitStatus`), and the git config (`GitRepository` plus its
+`ScmProvider` or `ClusterScmProvider`).
+
+!!! warning "Secrets are never included"
+    The bundle resolves the SCM provider but **never** reads or includes the
+    credentials `Secret` it references.
```
crenshaw-dev:
```suggestion
> [!WARNING]
> *Secrets are never included*
>
> The bundle resolves the SCM provider but **never** reads or includes the
> credentials `Secret` it references.
```

github-style admonitions get translated into mkdocs-style by a plugin - using gh-style makes it look pretty good in both

api/dashboard/dashboard/types.go
zachaller: ahh your thinking an apply side api, that apply API does not exist with this resource not sure if we could add it or if it would have to be another type like PromotionStrategyDetailsFiltered or something where you end up calling apply/create and your filter settings are in the spec however that would not work with Watch endpoint

docs/getting-started.md
```diff
@@ -494,10 +494,71 @@ spec:
...
+
+### Install the dashboard aggregation API
+
+The aggregation apiserver is **not** part of the plain `install.yaml` (it needs a TLS serving
+cert, and the `caBundle` must match it). Instead, each release publishes a **combined bundle**
+that contains the controller *and* the dashboard apiserver in a single file — so installing
+with the dashboard is one `kubectl apply`. Two variants are published; pick the one that
+matches how you want to manage the serving cert. Use one of these **instead of** the plain
+`install.yaml` from [Installation](#installation) above (each combined bundle already includes
+the controller).
```
crenshaw-dev:
```suggestion
The UI is backed by the `PromotionStrategyDetails` resource. This resource is backed by a Kubernetes APIService and requires a certificate to be set up. This step is safe to skip if you are not going to use the dashboard.

### Install the dashboard API

The `PromotionStrategyDetails` APIService does not come with the default install.yaml. Use one of the following bundles to set it up for your environment.
```

Seemed like way tmi for a getting started page.

docs/getting-started.md
```diff
@@ -494,10 +494,71 @@ spec:
...
+
+See the [Dashboard Aggregation API](dashboard-apiserver.md#serving-certs) page for the
+scripted/manual cert steps and cert rotation.
+
+///
+
+> [!NOTE]
+> The combined bundles are published starting with the first release that ships the dashboard
+> aggregation API; on older releases only the plain `install.yaml` exists.
+
```
crenshaw-dev:
```suggestion
```

Don't need it, look forward

docs/getting-started.md
```diff
@@ -494,10 +494,71 @@ spec:
...
+issue and rotate the serving cert (and keep the `caBundle` injected) automatically — a single
+apply with nothing else to do:
+
+```bash
+kubectl apply -f https://github.com/argoproj-labs/gitops-promoter/releases/download/v0.30.1/install-with-dashboard-cert-manager.yaml
+```
+
+///
+
+/// tab | Bring your own cert
```
crenshaw-dev, line 557: We don't want the insecure option here as well?

internal/apiserver/builder.go
```diff
@@ -0,0 +1,220 @@
...
+// sanitize strips managedFields and the last-applied-configuration annotation so
+// bundles stay small and free of noisy server-side-apply metadata.
+func sanitize(obj client.Object) {
+	obj.SetManagedFields(nil)
+	annotations := obj.GetAnnotations()
+	if annotations != nil {
+		delete(annotations, lastAppliedAnnotation)
+		obj.SetAnnotations(annotations)
+	}
+}
```
crenshaw-dev: I think you could use a transforming informer to drop these on cache population. Would also help keep the at-rest memory use low.

docs/getting-started.md
zachaller: No because we don't bundle it at release time so there is  no artifact etc and you have to clone repo apply etc etc  it's mostly dev not for the getting started, and I have debated removing it and just using local because that is how I run dev but I could see some wanting an in cluster dev setup too so meh.

ui/shared/src/types/bundle.ts
```diff
@@ -0,0 +1,62 @@
...
+  changeTransferPolicies?: ChangeTransferPolicy[];
+  pullRequests?: unknown[];
+  commitStatuses?: unknown[];
+  argoCDCommitStatuses?: unknown[];
+  gitCommitStatuses?: unknown[];
+  timedCommitStatuses?: unknown[];
+  webRequestCommitStatuses?: unknown[];
+  gitRepository?: unknown;
+  scmProvider?: unknown;
+  clusterScmProvider?: unknown;
```
crenshaw-dev, line 61: Are we at the point where we should codegen these types?

mkdocs.yml
```diff
@@ -53,6 +53,7 @@ nav:
       - Finalizers: debugging/finalizers.md
       - Labels: debugging/labels.md
   - Architecture: architecture.md
+  - Dashboard Aggregation API: dashboard-apiserver.md
```
crenshaw-dev, line 58: Seems like maintainer docs, maybe should be nested under the contributing section

ui/shared/src/types/bundle.ts
zachaller: Not in this PR but it was on my todo.

mkdocs.yml
zachaller: these are not really maintainer docs

cmd/demo/config/config.yaml
```diff
@@ -2,4 +2,4 @@ argocd:
   upstream: https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install-with-hydrator.yaml
 
 gitops-promoter:
-  upstream: https://github.com/argoproj-labs/gitops-promoter/releases/download/v0.31.0/install.yaml
+  upstream: https://github.com/argoproj-labs/gitops-promoter/releases/download/v0.31.0/install-without-ui.yaml
```
crenshaw-dev: TODO: open issue to fix UI for the demo command

hack/bump-docs-manifests.sh
```diff
@@ -29,6 +29,6 @@ echo "Bumped manifest versions to v$NEW_VERSION in docs."
 sed_replace "s/.*/v$NEW_VERSION/" cmd/demo/config/promoter_version
 sed_replace "s/download\/v[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}/download\/v$NEW_VERSION/g" cmd/demo/config/argocd-extension.yaml
 sed_replace "s/gitops-promoter_[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}_checksums\.txt/gitops-promoter_${NEW_VERSION}_checksums.txt/g" cmd/demo/config/argocd-extension.yaml
-sed_replace "s|gitops-promoter/releases/download/v[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}/install.yaml|gitops-promoter/releases/download/v$NEW_VERSION/install.yaml|g" cmd/demo/config/config.yaml
+sed_replace "s|gitops-promoter/releases/download/v[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}/install-without-ui.yaml|gitops-promoter/releases/download/v$NEW_VERSION/install-without-ui.yaml|g" cmd/demo/config/config.yaml
```
crenshaw-dev: Add bump logic for all bundles?

internal/apiserver/index.go
```diff
@@ -0,0 +1,444 @@
...
+		if wasKnown {
+			tombstone := &viewv1alpha1.PromotionStrategyDetails{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            key.Name,
+					Namespace:       key.Namespace,
+					ResourceVersion: p.currentResourceVersion(),
+				},
+			}
+			p.broadcast(watch.Event{Type: watch.Deleted, Object: tombstone}, key)
+		}
```
crenshaw-dev, line 188: I could be misreading, but this makes it look like the deletion of _any_ aggregated resource will produce a DELETED event on the PromotionStrategyDetails resource. But unless the PromotionStrategy was deleted, the PromotionStrategyDetails resource has only been updated, not deleted.

internal/apiserver/index.go
```diff
@@ -0,0 +1,444 @@
...
+
+// BundleProvider builds PromotionStrategyDetails bundles from a read-only
+// controller-runtime cache, and fans out watch events to registered watchers when
+// any related resource changes.
+type BundleProvider struct {
+	cache    cache.Cache
+	reader   client.Reader
+	queue    workqueue.TypedRateLimitingInterface[types.NamespacedName]
+	watchers map[int]*bundleWatcher
+	known    map[types.NamespacedName]bool
```
crenshaw-dev:
```suggestion
	// known keeps track of PromotionStrategies we've encountered, so we can detect adds/deletes
	known    map[types.NamespacedName]bool
```

internal/apiserver/index.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+	if err != nil {
+		log.Error(err, "failed to build bundle", "namespace", key.Namespace, "name", key.Name)
+		return
+	}
+
+	p.mu.Lock()
+	eventType := watch.Modified
+	if !p.known[key] {
+		eventType = watch.Added
```
crenshaw-dev, line 199: So we'll produce an added event for every PSD on APIService restart... that's probably fine. But why not just align with the PromotionStrategy add/delete events?

internal/apiserver/index.go
```diff
@@ -0,0 +1,444 @@
...
+		list, err := p.List(ctx, namespace)
+		if err != nil {
+			w.Stop()
+			return nil, err
+		}
+		for i := range list.Items {
+			item := list.Items[i]
+			if name != "" && item.Name != name {
+				continue
+			}
```
crenshaw-dev: If name is not empty, wouldn't it be much more efficient to just Get instead of List?

internal/apiserver/index.go
```diff
@@ -0,0 +1,444 @@
...
+const debounceInterval = 250 * time.Millisecond
+
+// BundleProvider builds PromotionStrategyDetails bundles from a read-only
+// controller-runtime cache, and fans out watch events to registered watchers when
+// any related resource changes.
+type BundleProvider struct {
+	cache    cache.Cache
+	reader   client.Reader
+	queue    workqueue.TypedRateLimitingInterface[types.NamespacedName]
+	watchers map[int]*bundleWatcher
```
crenshaw-dev, line 56:
```suggestion
	// watchers tracks all active watches occurring at the moment - this facilitates broadcasting events to all relevant watches
	watchers map[int]*bundleWatcher
```

internal/apiserver/index.go
```diff
@@ -0,0 +1,444 @@
...
+
+// broadcast sends an event to all watchers matching the affected key.
+func (p *BundleProvider) broadcast(ev watch.Event, key types.NamespacedName) {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	for _, w := range p.watchers {
+		if !w.matches(key.Namespace, key.Name) {
+			continue
+		}
+		evCopy := watch.Event{Type: ev.Type, Object: ev.Object.DeepCopyObject()}
```
crenshaw-dev, line 444:
```suggestion
		// TODO: is the DeepCopyObject necessary?
		evCopy := watch.Event{Type: ev.Type, Object: ev.Object.DeepCopyObject()}
```

hack/bump-docs-manifests.sh
zachaller: not really needed for demo cli but went ahead and did it anyways for future ref if we update those

internal/apiserver/index.go
zachaller: Because we debounce the many updates that happen to resources such as the PS there is not a 1:1 mapping so we can't tie to that stream

internal/apiserver/index.go
zachaller: this branch only fires when buildBundle returns NotFound, and the only thing that can produce that is the Get of the PromotionStrategy itself. Child resources are gathered via List (returns empty, never NotFound), and the two Gets for GitRepository/ScmProvider swallow NotFound. So deleting an aggregated resource rebuilds the bundle with the PS still present and broadcasts MODIFIED; DELETED is reserved for the PromotionStrategy itself being gone.

crenshaw-dev (approved):

PR #1520, opened 2026-06-01
chore: group controller-tools bump with k8s platform

crenshaw-dev:
controller-tools should upgrade in lockstep with the related k8s dependencies


zachaller (approved):

crenshaw-dev:
Actually looks like this is wrong, controller-tools is reasonably independent from k8s version.

PR #1539, opened 2026-06-03
fix: remove PR finalizer when GitRepository is deleted during termination

ZeBidule:
## Problem

When a `PullRequest` CRD has a `deletionTimestamp` but its referenced `GitRepository` no longer exists, `getPullRequestProvider` fails with a not-found error **before** `handleFinalizer` is ever reached. This causes the finalizer to remain permanently, leaving the `PullRequest` stuck in `Terminating` state indefinitely.

This has a cascading effect:
1. The stuck `PullRequest` causes `ChangeTransferPolicy` reconciliation to report `PullRequestNotReady`
2. The `PromotionStrategy` oscillates between `Unknown` and `True`
3. The ArgoCD application for the addon flaps between `Healthy` and `Degraded`

The error logged repeatedly is:
```
ERROR Reconciler error {"controller": "pullrequest", ..., "error": "failed to get PullRequest provider: failed to get ScmProvider and secret: failed to get GitRepository: failed to get GitRepository: GitRepository.promoter.argoproj.io \"<name>\" not found"}
```

This was observed in production when a `GitRepository` CRD was deleted (e.g. an addon was removed from the promotion list) while a `PullRequest` CRD referencing it still had a pending `deletionTimestamp`.

## Fix

In `Reconcile()`, after `getPullRequestProvider` fails with a not-found error, check whether the `PullRequest` is already being deleted. If so, skip provider resolution and remove the finalizer directly. There is nothing to close on the SCM since we cannot resolve the provider, and unblocking garbage collection is the correct behavior.

The existing `handleEmptyIDDeletion` already handles the case where `status.ID` is empty. This fix handles the complementary case where `status.ID` is set (a PR was created on the SCM) but the `GitRepository` is gone.

## Test

Adds a regression test: creates a `PullRequest` backed by a `GitRepository`, waits for it to be open, deletes the `GitRepository`, then deletes the `PullRequest` and asserts it is eventually garbage-collected without manual finalizer intervention.

### DIFF
--- a/internal/controller/pullrequest_controller.go
+++ b/internal/controller/pullrequest_controller.go
@@ -40,7 +40,7 @@ import (
 	promoterConditions "github.com/argoproj-labs/gitops-promoter/internal/types/conditions"
 	"github.com/argoproj-labs/gitops-promoter/internal/types/constants"
 	"github.com/argoproj-labs/gitops-promoter/internal/utils"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -53,6 +53,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
+
+// isNotFoundError returns true if the error chain contains a Kubernetes not-found error.
+func isNotFoundError(err error) bool {
+	return k8serrors.IsNotFound(err)
+}
 
 // PullRequestReconciler reconciles a PullRequest object
 type PullRequestReconciler struct {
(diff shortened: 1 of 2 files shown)


internal/controller/pullrequest_controller.go
```diff
@@ -99,6 +103,21 @@ func (r *PullRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		// on the SCM but we must still remove the finalizer to unblock garbage collection.
+		// Leaving the finalizer in place would cause the PullRequest to remain stuck in
+		// Terminating state indefinitely, blocking the PromotionStrategy reconciliation.
+		if !pr.DeletionTimestamp.IsZero() && isNotFoundError(err) {
+			logger.Info("PullRequest is being deleted but provider cannot be resolved due to missing dependency; removing finalizer without closing SCM PR", "error", err)
+			if controllerutil.ContainsFinalizer(&pr, promoterv1alpha1.PullRequestFinalizer) {
+				controllerutil.RemoveFinalizer(&pr, promoterv1alpha1.PullRequestFinalizer)
+				if updateErr := r.Update(ctx, &pr); updateErr != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to remove finalizer after missing dependency: %w", updateErr)
+				}
```
crenshaw-dev, line 117: The `gitrepository.promoter.argoproj.io/finalizer` on the GitRepository CR is meant to prevent this from happening. I think the bug is that the GitRepository finalizer is released when the PullRequest deletion timestamp is set, not when the PullRequest is actually deleted (i.e. cleanup is finished). I think we should make the GitRepository finalizer logic more strict rather than making the PullRequest logic less strict.

I do like this block of code though, because it gives us a place to return a more tailored message. e.g., we can say "This PullRequest's GitRepository has been deleted, so it cannot ensure that corresponding SCM pull request is closed. Either restore the GitRepository or remove the `pullrequest.promoter.argoproj.io/finalizer` finalizer and manually close the pull request if it is not already closed."

crenshaw-dev: Can you take a look at an alternative fix here? https://github.com/argoproj-labs/gitops-promoter/pull/1542

zachaller:
closed by: https://github.com/argoproj-labs/gitops-promoter/pull/1542

PR #1540, opened 2026-06-03
fix: unblock CTP finalizer removal when PR is externally merged or closed

ZeBidule:
## Problem

When a promoter PullRequest is **manually merged or closed** on the SCM (outside the controller), the PR CRD gets a `deletionTimestamp` but can remain stuck in `Terminating` state indefinitely. This causes the `PromotionStrategy` to oscillate between `Unknown` and `True`, and the ArgoCD application for the addon to flap between `Healthy` and `Degraded`.

This was observed in production on a GitLab-backed installation: a promoter MR was manually closed, and the corresponding `PullRequest` CRD stayed stuck in `Terminating` for several days, continuously degrading the ArgoCD app health check.

## Root Cause

In `changetransferpolicy_controller.go`, `handlePRFinalizerRemoval` decides whether it is safe to remove the `ChangeTransferPolicyPullRequestFinalizer` from a `PullRequest` that is being deleted. It does so by comparing the CTP persisted status against the PR current status:

```go
statusMatches := ctp.Status.PullRequest.ID == prItem.Status.ID &&
    ctp.Status.PullRequest.State == prItem.Status.State &&
    boolPtrEqual(ctp.Status.PullRequest.ExternallyMergedOrClosed, prItem.Status.ExternallyMergedOrClosed)
```

When a PR is externally merged or closed, the PR controller sets `ExternallyMergedOrClosed=true` and **clears `State` to `""`** (because we do not know whether it was merged or closed). If the CTP has not yet reconciled to copy this new status — for example because the PR controller restarted before persisting the status update, or because the CTP reconcile cycle simply has not run yet — the CTP status still shows `State="open"` while the PR status shows `State=""`. This mismatch causes `handlePRFinalizerRemoval` to return without removing the finalizer, and the `PullRequest` is permanently stuck.

## Fix

When `ExternallyMergedOrClosed=true` on the PR, skip the `State` comparison and only require that `ExternallyMergedOrClosed` has been synced to the CTP status. The `State` field is intentionally `""` in this case (see the comment in `syncStateFromProvider`) and may legitimately be stale on the CTP — the `ExternallyMergedOrClosed` flag is the authoritative signal that the PR is no longer active and the finalizer can be safely removed.

For PRs in a normal terminal state (controller-initiated merge or close), the existing full status match check is preserved.

## Test

Adds a regression test that simulates the race condition: `ExternallyMergedOrClosed=true` is set on the PR and the PR is deleted before the CTP has had a chance to sync the new status (CTP still shows `State="open"`, `ExternallyMergedOrClosed=nil`). The test asserts that the PR is eventually garbage-collected without manual finalizer intervention.


crenshaw-dev:
@ZeBidule is it possible that the proposed branch was deleted when the PR was merged? I've been unable to reproduce the described failure.

ZeBidule:
I will try to reproduce it and I will come back to you...

ZeBidule:
I am not able to reproduce the issue neither so I close the issue, sorry for the inconvenience

crenshaw-dev:
No worries, thanks for the report!

PR #1542, opened 2026-06-03
fix: keep GitRepository until referencing PullRequests are removed

crenshaw-dev:
## Summary

Builds on the investigation in #1539 (thanks @ZeBidule). The root cause is that the `GitRepository` finalizer treated a `PullRequest` with only a `deletionTimestamp` as non-blocking, so the repository CR could disappear while a terminating `PullRequest` still needed it for SCM cleanup.

- **GitRepository:** block deletion until referencing `PullRequest` objects are gone from the API (not merely terminating); watch `PullRequest` deletes to requeue the repo.
- **PullRequest:** when a terminating PR loses its `GitRepository`, log a tailored operator message and return an error instead of silently removing `pullrequest.promoter.argoproj.io/finalizer`.
- **Test:** regression for the race using a bogus finalizer so the PR stays in etcd with `deletionTimestamp` set.

Preserves @ZeBidule's commit for credit; follow-up commit implements the stricter finalizer approach from [review on #1539](https://github.com/argoproj-labs/gitops-promoter/pull/1539#discussion_r3348874885).

## Test plan

- [x] `make lint`
- [x] `make test-parallel`
- [x] Focused regression: `ginkgo.focus="should not delete the GitRepository while the PullRequest object still exists"`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/controller/gitrepository_controller.go
+++ b/internal/controller/gitrepository_controller.go
@@ -90,7 +90,12 @@ func (r *GitRepositoryReconciler) SetupWithManager(ctx context.Context, mgr ctrl
 			&promoterv1alpha1.PullRequest{},
 			handler.EnqueueRequestsFromMapFunc(r.enqueueGitRepositoryForPullRequest),
 			// Delete fires once the PullRequest leaves the informer cache (fully removed from the API).
-			builder.WithPredicates(predicate.Funcs{DeleteFunc: func(event.DeleteEvent) bool { return true }}),
+			builder.WithPredicates(predicate.Funcs{
+				CreateFunc:  func(event.CreateEvent) bool { return false },
+				UpdateFunc:  func(event.UpdateEvent) bool { return false },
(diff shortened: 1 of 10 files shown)


internal/controller/gitrepository_controller.go
```diff
@@ -99,10 +114,6 @@ func (r *GitRepositoryReconciler) handleFinalizer(ctx context.Context, gitRepo *
 
 		var dependentPRs []string
 		for _, pr := range pullRequests.Items {
-			// Skip PullRequests that are also being deleted (allows cascade deletion)
-			if !pr.DeletionTimestamp.IsZero() {
-				continue
-			}
```
crenshaw-dev, line 105: This is the actual fix.

zachaller (approved):

ZeBidule:
Thanks a lot for your reactivity :)

PR #1548, opened 2026-06-03
chore(ui): generate typescript types

crenshaw-dev:
## Summary by CodeRabbit

* **New Features**
  * Added automated generation to keep UI TypeScript types in sync with the view API, including new `make` targets and scripts.
* **Improvements**
  * Strengthened CI code-generation verification for both API and UI artifacts.
  * Streamlined UI dependency installation for dashboard/extension builds and linting.
  * Updated UI linting/type-checking to better handle generated and optional data.
* **Bug Fixes**
  * Promotion strategy details list endpoint now returns an empty list reliably (not a null payload).
* **Documentation**
  * Expanded guidance for regenerating UI types.
* **Schema Changes**
  * `PromotionStrategy.spec` is now always included and required by the CRD.


zachaller (commented):
Review of the generated-types change. The codegen mechanics (view2ts extractor, openapi-typescript runner, CI/Renovate wiring, type-alias layer) look solid. A few items to address inline — one likely-accidental file plus an empty-list runtime regression.

install.yaml
```diff
@@ -0,0 +1,8403 @@
+# This is an auto-generated file. DO NOT EDIT
```
zachaller: This file looks accidental — recommend removing it.

`install.yaml` isn't tracked on `main`; it was intentionally removed in favor of the three `dist/install-*.yaml` bundles. Nothing in this PR regenerates a root-level `install.yaml`, and it's only worked around in `_typos.toml` rather than registered in `.gitattributes` like the other generated bundles — which suggests it was produced by a stray `make build-installer` / `kustomize build` and committed by mistake.

Please `git rm install.yaml` and drop the matching `_typos.toml` exclusion.

_typos.toml
```diff
@@ -3,6 +3,8 @@ extend-exclude = [
     "docs/assets/**",
     "config/crd/bases/**",
     "dist/**",
+    "install.yaml",
```
zachaller: Remove this exclusion along with the accidental `install.yaml` file (see other comment). The `ui/shared/src/types/generated/**` exclusion below is correct and should stay.

```suggestion
```

ui/dashboard/src/stores/CRDStore.ts
```diff
@@ -68,37 +70,35 @@ export function createCRDStore<T extends CRDItem>(kind: string, eventName: strin
...
       set({ loading: true, error: null });
       try {
         const res = await fetch(`/list?kind=${kind}&namespace=${namespace}`);
 
         if (!res.ok) throw new Error(`Error: ${res.status}`);
-        const data: PromotionStrategyBundle[] = await res.json();
+        const data: PromotionStrategyDetails[] = await res.json();
 
-        set({ items: (data || []).map((b) => bundleToItem<T>(b)), loading: false });
+        set({ items: data.map((b) => bundleToItem<T>(b)), loading: false });
```
zachaller: This will throw on an empty namespace. The `/list` handler returns `c.JSON(http.StatusOK, bundleList.Items)`, and `bundleList.Items` is a Go nil slice when no `PromotionStrategy` exists — which marshals to JSON `null`. With the `data || []` guard gone, `data.map(...)` throws, the `catch` fires, and the dashboard shows an error instead of an empty state.

The `no-unnecessary-condition` rule flagged the old guard only because the annotation claims `data` is always an array, which isn't true on the wire. Make the type honest so the guard is necessary and the lint passes:

```suggestion
        const data = (await res.json()) as PromotionStrategyDetails[] | null;

        set({ items: (data ?? []).map((b) => bundleToItem<T>(b)), loading: false });
```

Alternatively (or additionally), return `[]` from `httpList` instead of a nil slice so every client is safe.

ui/dashboard/src/stores/CRDStore.ts
```diff
@@ -12,38 +12,40 @@ interface CRDItem extends PromotionStrategy {
...
   const ps = bundle.promotionStrategy;
-  const environments = environmentsFromCTPs(ps, bundle.changeTransferPolicies ?? []);
-  const psWithEnvironments: PromotionStrategy = {
+  const environments = environmentsFromCTPs(ps.spec, bundle.changeTransferPolicies ?? []);
+  const psWithEnvironments = {
     ...ps,
+    metadata: {
+      name: bundle.metadata.name,
+      namespace: bundle.metadata.namespace,
+    },
```
zachaller, line 47: This drops PromotionStrategy metadata that the Live Manifest view renders. The previous `{ ...ps, status }` preserved the embedded PS's full `metadata` (the server sends full ObjectMeta, dropping only status). Overwriting `metadata` with just `name`/`namespace` discards `labels`, `annotations`, `creationTimestamp`, `uid`, etc., and `LiveManifestView` does `yaml.dump(strategy)`, so the manifest loses those fields.

You only need to override to satisfy the required `name`/`namespace`; keep the rest:

```suggestion
    metadata: {
      ...ps.metadata,
      name: bundle.metadata.name,
      namespace: bundle.metadata.namespace,
    },
```

zachaller (approved):

PR #1559, opened 2026-06-04
Fix/ctp finalizer stuck on externally merged pr alternative

crenshaw-dev:


crenshaw-dev:
Closing since we've been unable to reproduce the bug

Issue #1513, opened 2026-06-01
PromotionStrategy improvements


crenshaw-dev:
# PromotionStrategy: the Status Quo

### 1. Provide a nice interface for configuring multiple ChangeTransferPolicies

Take this example:

```yaml
kind: PromotionStrategy
spec:
  gitRepositoryRef:
    name: my-repo
  proposedCommitStatuses:
  - key: lint
  activeCommitStatuses:
  - key: argocd-health
  environments:
  - branch: dev
  - branch: test
    activeCommitStatuses:
    - key: perf-test
  - branch: prod
```

This PromotionStrategy is equivalent to these ChangeTransferPolicies:

```yaml
kind: ChangeTransferPolicy
spec:
  gitRepositoryRef:
    name: my-repo
  proposedCommitStatuses:
  - key: lint
  - key: promoter-previous-environment
  activeCommitStatuses:
  - key: argocd-health
---
kind: ChangeTransferPolicy
spec:
  gitRepositoryRef:
    name: my-repo
  proposedCommitStatuses:
  - key: lint
  - key: promoter-previous-environment
  activeCommitStatuses:
  - key: argocd-health
  - key: perf-test
---
kind: ChangeTransferPolicy
spec:
  gitRepositoryRef:
    name: my-repo
  proposedCommitStatuses:
  - key: lint
  - key: promoter-previous-environment
  activeCommitStatuses:
  - key: argocd-health
```

Compared to creating the ChangeTransferPolicies manually, the PromotionStrategy has several advantages:

1. "global" active/proposed commit statuses are automatically applied to _all_ CTPs - less duplication
2. the promoter-previous-environment proposed commit status is automatically applied to all CTPs

### 2. Aggregate the statuses of the CTPs

The UI needs the status of the CTPs (currently running git commits, states of commit statuses, etc.) to display the state of the system. Today, the PromotionStrategy controller retrieves the state of all its managed CTPs and duplicates them in its own status field. This is what the UI consumes.

```yaml
kind: PromotionStrategy
status:
  environments:
  - branch: dev
     # stuff from the CTP status goes here
```

### 3. Maintain the promoter-previous-environment commit statuses

[...]

Issue #1364, opened 2026-04-24
Support deploying to two branches in parallel


crenshaw-dev:
Today, to deploy to two environments in parallel, you have to put both the environments' manifests into the same environment branch.

This has downsides:

1) Awkward branch naming: `environments/e2e-and-prf`?
2) Difficult to change order: need to create new branches, maybe delete old ones, etc.

The user should be able to define a DAG (or at least parallel branches if not a full DAG), and the promoter-previous-environment check should apply smartly to the correct CTPs to allow parallel deployment.

The UI should be updated to present the parallel branches instead of showing them in sequence.

zachaller:
I think even if we remove the previous environment auto inject and create a new controller for it like we want to it might make sense to support say this

```
spec:
  activeCommitStatuses:
    - key: argocd-health
  environments:
    - autoMerge: false
      branch: environments/development
    - autoMerge: true
      branch: environments/staging-1
      proposedCommitStatuses:
        - key: promoter-previous-environment
    - autoMerge: true
      branch: environments/staging-2
    - autoMerge: true
      branch: environments/staging-3
      proposedCommitStatuses:
        - key: promoter-previous-environment
    - autoMerge: true
      branch: environments/production
      proposedCommitStatuses:
        - key: change-management-approval
        - key: change-management-open
        - key: promoter-previous-environment
 ```

1 and 2 would go out together

Actually this is a bit of an awkward spec too just to reason about, although we would either  support it or error if prev is not on all env. We just need a DAG 😜

Even in dag mode two environments might not go together if there are other proposedCommitStatuses, need to think about that. 


Maybe 
```
  environments:
    - autoMerge: false
      branch: environments/development
    - autoMerge: true
      branch: environments/staging-1
      proposedCommitStatuses:
        - key: promoter-previous-environment
    - autoMerge: true
      branches: 
        - name: environments/staging-2
        - name: environments/staging-3
      proposedCommitStatuses:
        - key: promoter-previous-environment
    - autoMerge: true
      branch: environments/production
      proposedCommitStatuses:
        - key: change-management-approval
        - key: change-management-open
        - key: promoter-previous-environment
```
Although all commit status managers I think would have to support this?

There is a lot of complexity here for naming should probably weight that as well, really need to think about in all these cases supporting this at the CTP level across all commit statuses.

crenshaw-dev:
[...]

PR #1571, opened 2026-06-09
feat: add PreviousEnvironmentCommitStatus controller and CRD

Kaiying-Yu:
## What

Introduces a new `PreviousEnvironmentCommitStatus` (PECS) CRD and controller as
an intermediate stepping stone toward DAG-based promotion support.

Previously the previous-environment commit status was computed inside the
`PromotionStrategy` controller and auto-injected. This PR extracts that logic
into a dedicated, independent controller driven by a user-created
`PreviousEnvironmentCommitStatus` resource.

## Changes

- **New CRD** `PreviousEnvironmentCommitStatus` with `promotionStrategyRef` and
  `key` fields, OpenAPI schemas, sample manifest, and applyconfiguration codegen.
- **New controller** that watches the referenced `PromotionStrategy` and, for
  each environment, reports whether the preceding environment is synced and
  healthy via a `CommitStatus` on the current environment's proposed commit.
- **Decoupled** the previous-environment logic from the `PromotionStrategy`
  controller (removed auto-injection; the PECS controller now owns this end to end).
- Added `PreviousEnvironmentCommitStatusConfiguration` (work queue / requeue
  settings), RBAC roles, a controller field-owner constant, and startup wiring.
- Registered the new CRD and RBAC roles in the `Tiltfile` for local development.

## Notes

- The PECS resource is now created explicitly by the user (no auto-injection).
  The UX for global vs. per-environment configuration is an open question to be
  settled alongside DAG support.

## Known limitations

- `spec.key` is plumbed into the CommitStatus the controller writes, but the
  `PromotionStrategy` controller still hardcodes the proposed-status selector to
  the default key (`promoter-previous-environment`). Setting a custom `spec.key`
  will silently stall promotion until both sides agree. The field is reserved for
  DAG support; whether/how a custom key is used will be settled with the DAG spec.

Refs: #1513, #1364


crenshaw-dev:
Closing in favor of https://github.com/argoproj-labs/gitops-promoter/pull/1689

PR #1579, opened 2026-06-11
feat: Preserve promotion history via git notes when SCM rewrites merge commits

zachaller:
When a pull request is merged directly on the SCM (or the SCM squashes/ rewrites the merge commit message), the special trailers embedded in the commit message are lost and ChangeTransferPolicy history can no longer be reconstructed.

Write the trailer data as a git note (refs/notes/promoter.history) on the merge commit during PR finalization, when the CTP still holds the deleting PullRequest via its finalizer and pr.spec.commit.message still carries the full trailer block. History building now reads the note first and falls back to commit message trailers for commits that predate notes.

- Add GetHistoryNote/SetHistoryNote git primitives; SetHistoryNote is idempotent (notes add -f) and retries non-fast-forward pushes after re-fetching the shared notes ref
- FetchNotes now fetches both hydrator and promoter history notes refs, each tolerating a missing remote ref
- Locate the merge commit by walking first-parent history of the active branch: exact match (fast-forward), second parent or descendant of the recorded merge SHA (merge commit), or tree equality (squash merge); closed-not-merged PRs match nothing and skip the note
- Derive Pull-request-merge-time from the merge commit's committer time since the in-flight merge trailer is never persisted to the spec
- Move handlePRFinalizerRemoval after the clone/notes fetch; note write failures keep the PR finalizer so the write is retried, with a PromotionHistoryNoteFailed warning event for visibility

## Summary by CodeRabbit

* **New Features**
  * Promotion history for merged pull requests is now saved as Git notes on merge commits.
  * Pull request status records the merged target commit.
  * External merges and squash merges are tracked more reliably.

* **Bug Fixes**
  * Finalizers remain in place when history cannot be saved, enabling retries.
  * Git notes take precedence, with commit trailers as a fallback.
  * Added detection and warnings for merge-commit snapshot mismatches.

* **Documentation**
  * Clarified finalizer behavior, history-loss risks, merge tracking, and warning events.

* **Tests**
  * Added coverage for note persistence, retries, external merges, squash merges, fallback behavior, and merge detection.


FourFifthsCode:
Tested against a local cluster with gitops-promoter, Argo CD with the GitOps Promoter extension, and a multi-environment PromotionStrategy (dev → e2e → prod) backed by GitHub.

---

### What was tested

| Scenario | Result |
|----------|--------|
| **Happy path (auto-merge)** | Promotion cascades through environments; git notes written on merge commits; CTP retains merged PR metadata; History populated with merge times |
| **External SCM merge** | Manual merge on GitHub detected; history note written even when merge commit message lacks trailers |
| **Squash merge** | SCM-stripped commit message; full trailer metadata recovered from git note |
| **Close without merge** | Closed PR on GitHub without merging; active branch unchanged; no history note written; no false LIVE entry in History |
| **Finalizer ordering** | Note written before CTP finalizer removed; no `PromotionHistoryNoteFailed` events observed |
| **Pre-notes fallback (UI)** | Older History rows (pre–git-notes) still render correctly alongside new note-backed promotions |

---
### Conclusion
Verified the primary use case (trailers preserved via git notes when SCM rewrites merge commits) and the main edge cases in the PR description: external merge, squash merge, close-without-merge, finalizer-gated note writes, and History UI behavior for both note-backed and legacy commits all worked correctly.

PR should be ready to merge once any conflicts are resolved.

zachaller:
Opened follow-up issue for the authoritative SCM merge-commit SHA path documented in the `findMergeCommitOnActiveBranch` TODO: #1895

That issue proposes persisting `status.mergeCommitSha` from the provider (e.g. GitHub `merge_commit_sha`) so `writePromotionHistoryNote` can skip `findMergeCommitOnActiveBranch` / `findDrySha` in the common case, while keeping this PR's git inference as fallback.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1074,248 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+	logger := log.FromContext(ctx)
+
+	if livePR.Status.State == promoterv1alpha1.PullRequestClosed {
+		logger.V(4).Info("PR was closed, not merged, skipping promotion history note")
+		return nil
+	}
+	if livePR.Spec.MergeSha == "" {
+		logger.V(4).Info("PR has no merge sha, skipping promotion history note")
+		return nil
+	}
```
crenshaw-dev: Why skip if there's no merge sha?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1074,248 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+	if livePR.Status.State != promoterv1alpha1.PullRequestMerged && !externallyMergedOrClosed {
+		logger.V(4).Info("PR is not merged or externally merged/closed, skipping promotion history note",
+			"prState", livePR.Status.State)
+		return nil
+	}
+
+	trailers, err := git.ParseTrailersFromMessage(ctx, livePR.Spec.Commit.Message)
+	if err != nil {
+		return fmt.Errorf("failed to parse trailers from PullRequest commit message: %w", err)
+	}
```
crenshaw-dev, line 1128: Is there a risk of drift here? i.e. if the PR was externally merged, there's no guarantee that the commit message is up-to-date, is there? Most importantly, could the DRY sha have drifted?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1074,248 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+	if len(trailers) == 0 {
+		// A brand-new PR can be merged before the CTP's trailer-bearing update ever ran; there is nothing to record.
+		logger.V(4).Info("PR commit message has no trailers, skipping promotion history note")
+		return nil
+	}
+
+	mergeCommitSha, err := findMergeCommitOnActiveBranch(ctx, gitOperations, ctp.Spec.ActiveBranch, ctp.Spec.ProposedBranch, ctp.Spec.ActivePath,
+		livePR.Spec.MergeSha, getFirstTrailerValue(trailers, constants.TrailerShaDryProposed))
```
crenshaw-dev: What if it was a squash commit?

crenshaw-dev: I guess "merge commit" here is overloaded. It's not necessarily a "merge commit" in the git sense, but it is the commit that merged the `-next` branch into the active branch.

Could we call it the "promote commit" or something instead to help clarify?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1074,248 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+//     a final pass so it can never shadow a parent-link match).
+//
+// Returns "" when no matching commit is found, which usually means the PR was closed without merging.
+//
+// TODO: replace this inference with the authoritative answer from the SCM. Providers expose the merge commit
+// sha on the pull request even after it is merged (e.g. GitHub's merge_commit_sha, GitLab's
+// merge_commit_sha/squash_commit_sha); a provider method that looks up a no-longer-open PR could let the
+// PullRequest controller persist status.mergeCommitSha and the true merge time before the resource is
+// deleted, so finalization annotates exactly that commit — and it would also disambiguate
+// ExternallyMergedOrClosed into "merged" vs "closed".
```
crenshaw-dev: Why not do this as the initial implementation?

internal/git/git.go
```diff
@@ -872,6 +934,170 @@ func (g *EnvironmentOperations) GetHydratorNote(ctx context.Context, sha string)
...
+			logger.V(4).Info("Pushed history note", "sha", sha, "attempt", attempt)
+			return nil
+		}
+
+		lastErr = fmt.Errorf("failed to push history note for sha %q: %w", sha, err)
+		if !strings.Contains(stderr, "non-fast-forward") && !strings.Contains(stderr, "fetch first") && !strings.Contains(stderr, "[rejected]") {
+			logger.Error(err, "Failed to push history note", "sha", sha, "stderr", stderr)
+			return lastErr
+		}
+		logger.V(4).Info("History note push rejected, retrying", "sha", sha, "attempt", attempt, "stderr", stderr)
```
crenshaw-dev, line 1070: No backoff?

crenshaw-dev (changes requested):

internal/controller/testdata/ChangeTransferPolicy.yaml
```diff
@@ -84,3 +84,21 @@ status:
...
+    # dry/hydrated SHAs were reconstructed from the merge commit, but the recorded commit statuses may be stale.
+    - active:
+        dry:
+          sha: "1111111111111111111111111111111111111111"
+      proposed:
+        hydrated:
+          sha: "2222222222222222222222222222222222222222"
+        commitStatuses:
+          - key: example-key
+            phase: success
```
crenshaw-dev, line 102: My understanding is that we can _always_ recover the merged dry sha. And, I assume, we can always match up the commit status state with a hydrated sha which in turn corresponds to a single dry sha.

So if we know that the merged dry sha doesn't match up with the stored commit status state, shouldn't we just discard the contents of the commitStatuses field?

api/v1alpha1/pullrequest_types.go
```diff
@@ -121,6 +121,13 @@ type PullRequestStatus struct {
 	// +kubebuilder:validation:XValidation:rule="self == '' || isURL(self)",message="must be a valid URL"
 	// +kubebuilder:validation:Pattern="^(https?://.*)?$"
 	Url string `json:"url,omitempty"`
+	// MergeCommitSha is the commit SHA on the target branch reported by the SCM after merge.
+	// Set once by the PullRequest controller via a Get-by-ID lookup when FindOpen no longer finds the PR.
```
crenshaw-dev: Instead of _always_ recovering this via `Get` shouldn't we proactively set it when the `Merge` call succeeds with a known merge sha?

zachaller: done, not all providers return it so meh but the ones that do we save some work

api/v1alpha1/pullrequest_types.go
```diff
@@ -121,6 +121,15 @@ type PullRequestStatus struct {
...
 	Url string `json:"url,omitempty"`
+	// MergeCommitSha is the commit SHA on the target branch reported by the SCM after merge.
+	// Set once by the PullRequest controller, either from the merge response for providers that
+	// report the SHA there, or from a Get-by-ID lookup when FindOpen no longer finds the PR
+	// (external merges, and providers whose merge response omits the SHA).
+	// +optional
+	// +kubebuilder:validation:MinLength=40
+	// +kubebuilder:validation:MaxLength=64
+	// +kubebuilder:validation:Pattern=`^([a-f0-9]{40}|[a-f0-9]{64})$`
+	MergeCommitSha string `json:"mergeCommitSha,omitempty"`
```
crenshaw-dev: Maybe just `mergeSha` to match spec?

api/v1alpha1/pullrequest_types.go
```diff
@@ -121,6 +121,15 @@ type PullRequestStatus struct {
...
 	// +kubebuilder:validation:Pattern="^(https?://.*)?$"
 	Url string `json:"url,omitempty"`
+	// MergeCommitSha is the commit SHA on the target branch reported by the SCM after merge.
+	// Set once by the PullRequest controller, either from the merge response for providers that
+	// report the SHA there, or from a Get-by-ID lookup when FindOpen no longer finds the PR
+	// (external merges, and providers whose merge response omits the SHA).
+	// +optional
+	// +kubebuilder:validation:MinLength=40
+	// +kubebuilder:validation:MaxLength=64
+	// +kubebuilder:validation:Pattern=`^([a-f0-9]{40}|[a-f0-9]{64})$`
```
crenshaw-dev, line 135: Would it be reasonable to have a CEL rule that Merge[Commit]Sha is only set if status.state=merged?

internal/controller/testdata/ChangeTransferPolicy.yaml
zachaller: going to not lose data but hide and show in UI that it's stale etc

docs/debugging/finalizers.md
```diff
@@ -22,6 +22,66 @@ All promoter-defined finalizer strings live in the API package as constants (see
 
 No separate finalizer constant is defined for `PromotionStrategy`; RBAC may still mention `promotionstrategies/finalizers` for generic metadata updates. Behavior you care about for promotions is mostly on `ChangeTransferPolicy` and `PullRequest` as in the table above.
 
+## Promotion history git notes
+
+The `changetransferpolicy.promoter.argoproj.io/pullrequest-finalizer` exists so the ChangeTransferPolicy controller can write a **promotion-history git note** on the merge commit before the `PullRequest` CR is deleted. That note is the durable record of what was promoted (pull request metadata, gate phases, dry/hydrated SHAs) when the SCM rewrites or strips the merge commit message — for example after a squash merge or a merge performed in the SCM UI.
+
+Notes are stored at `refs/notes/promoter.history` on the Git repository. During reconciliation the controller reads them (falling back to commit-message trailers on older merges) and rebuilds `ChangeTransferPolicy.status.history`.
```
crenshaw-dev: > on older merges

We still produce trailers for all commits, right? So we would also fall back in cases where the note for some reason failed to be created?

docs/debugging/finalizers.md
```diff
@@ -22,6 +22,66 @@ All promoter-defined finalizer strings live in the API package as constants (see
...
+
+> [!TIP]
+> Prefer letting the promoter merge pull requests it manages. If you use `autoMerge: false` with Prow/Tide or similar, see [Dynamic Pull Request Labels](../advanced-usage/pull-request-labels.md#prow--tide-example) and the mismatch caveats below.
+
+### How the merge commit is identified
+
+The merge commit comes from **`PullRequest.status.mergedTargetSha`**, populated by the PullRequest controller from the SCM. When the promoter performs the merge itself and the provider returns the SHA in the merge response (GitHub, GitLab, Bitbucket Cloud), it is recorded immediately. Otherwise — external merges, and providers whose merge response omits the SHA (Gitea, Forgejo, Azure DevOps) — it is recovered by a `Get` by `status.id` once the PR is no longer open. CTP finalization attaches the promotion-history note to that commit — it does not walk git history to locate it.
```
crenshaw-dev:
```suggestion
The merge commit comes from **`PullRequest.status.mergedTargetSha`**, populated by the PullRequest controller from the SCM. When the promoter performs the merge itself and the provider returns the SHA in the merge response (GitHub, GitLab, Bitbucket Cloud), it is recorded immediately. Otherwise — external merges, and providers whose merge response omits the SHA (Gitea, Forgejo, Azure DevOps) — it is recovered by a `Get` by `status.id` once the PR is no longer open. CTP finalization attaches the promotion-history note to that commit.
```

Unnecessary detail since we removed that code.

docs/debugging/finalizers.md
zachaller: Yes on both counts — "older merges" was too narrow. `applyPullRequest` writes the trailers into `spec.commit.message` on every PR update, and `buildHistoryEntry` falls back whenever `GetHistoryNote` returns nothing, so a note that failed to be created hits the same path.

The one caveat worth stating is that the fallback only helps when the merge style preserved the commit message: a regular merge commit still carries the trailers, but a squash or SCM-UI merge generally doesn't, which is the case the note exists for.

Reworded the doc paragraph and the `buildHistoryEntry` comment to cover both "predates the notes" and "note was never written".

docs/debugging/finalizers.md
```diff
@@ -22,6 +22,66 @@ All promoter-defined finalizer strings live in the API package as constants (see
...
+
+Both merge styles use the same **`PullRequest.status.mergedTargetSha`** reported by the SCM (GitHub `merge_commit_sha`, GitLab `squash_commit_sha` or `merge_commit_sha`, and so on), whether it came from the merge response or a later `Get`. CTP finalization writes the promotion-history git note on that commit for either style — there is no git history walk to locate it.
+
+What still differs is **what can be reconstructed from git at that commit**:
+
+| | Regular merge (`--no-ff`) | Squash merge |
+| --- | --- | --- |
+| SCM-reported `mergedTargetSha` | Merge commit on active | Squash commit on active |
+| Promotion-history note written | Yes, when SCM reports merged + SHA | Yes, when SCM reports merged + SHA |
+| Correct proposed hydrated SHA from git | Yes — second parent of merge commit | **No** — single-parent squash commit; snapshot hydrated SHA kept |
```
crenshaw-dev: Specify the exact CTP status field that could be stale? Probably worth documenting the potential-staleness on the field's docstring.

docs/debugging/finalizers.md
```diff
@@ -22,6 +22,66 @@ All promoter-defined finalizer strings live in the API package as constants (see
...
+| Correct proposed dry SHA from git | Yes — `hydrator.metadata` on merge commit | Yes — `hydrator.metadata` on squash commit (when present) |
```
crenshaw-dev: Ditto on specifying the field and augmenting the docstring.

internal/controller/testdata/ChangeTransferPolicy.yaml
```diff
@@ -84,3 +85,24 @@ status:
         prCreationTime: 2023-09-01T00:00:00Z
         prMergeTime: 2023-09-01T01:00:00Z
         url: https://github.com/org/repo/pull/847
+        # In a history entry this is the active-branch commit the entry describes (same as active.hydrated.sha).
+        mergedTargetSha: "3333333333333333333333333333333333333333"
+    # A promotion whose PR was merged externally after the proposed branch advanced: the proposed
+    # dry/hydrated SHAs were reconstructed from the merge commit, but the recorded commit statuses may be stale.
```
crenshaw-dev: Is it also possible that the proposed SHAs (hydrated and dry) are stale?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -177,6 +171,15 @@ func (r *ChangeTransferPolicyReconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, fmt.Errorf("failed to fetch git notes: %w", err)
 	}
 
+	// Handle PR finalizer removal if PR is being deleted and CTP status is already synced. This runs after the
+	// clone/notes fetch because it writes the promotion-history git note before releasing the PR: removal now
+	// requires a working clone, which is acceptable since promotion as a whole is blocked anyway when git is
+	// unreachable, and handleCTPCleanupOnDelete remains the note-free escape hatch when the CTP itself is deleted.
```
crenshaw-dev: Probably worth being explicit why we need the note fetch in addition to the clone.

internal/controller/testdata/ChangeTransferPolicy.yaml
zachaller: Yes. The dry sha is always re-read from the merge commit, but the hydrated sha only comes from the merge commit's second parent — a squash or fast-forward has none, so it stays at the (possibly stale) snapshot. However, if there is a second parent it is updated.

internal/controller/changetransferpolicy_controller.go
zachaller: Its not needed it used to be until the retries where added, updated comment

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1080,153 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
 
+// writePromotionHistoryNote records the pull request's commit message trailers as a git note on the merge
+// commit of the active branch. The note is the durable copy of the trailer data: when the SCM rewrites the
+// merge commit message (squash merges, or merges performed directly on the SCM), the trailers are lost from
+// the commit itself, and history building falls back to this note.
+//
+// The merge commit SHA comes from status.mergedTargetSha (populated by the PullRequest controller from SCM
+// Get after merge). The old findMergeCommitOnActiveBranch git walk was removed in favor of that authoritative
+// source; reintroduce inference only if a provider cannot return the merged target SHA reliably, and wire any
+// revival through status.mergedTargetSha rather than silent first-parent walks here.
```
```suggestion
// Get after merge). If we discover that a provider cannot return the merged target SHA reliably, 
// we can introduce inference via a git walk and wire it through status.mergedTargetSha.
```

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1080,153 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+// or it never accumulated trailers.
+func (r *ChangeTransferPolicyReconciler) writePromotionHistoryNote(ctx context.Context, ctp *promoterv1alpha1.ChangeTransferPolicy, gitOperations *git.EnvironmentOperations, livePR *promoterv1alpha1.PullRequest) error {
+	logger := log.FromContext(ctx)
+
+	if livePR.Status.State == promoterv1alpha1.PullRequestClosed {
+		logger.V(4).Info("PR was closed, not merged, skipping promotion history note")
+		return nil
+	}
+	externallyMergedOrClosed := livePR.Status.ExternallyMergedOrClosed != nil && *livePR.Status.ExternallyMergedOrClosed
+	if livePR.Status.State != promoterv1alpha1.PullRequestMerged && !externallyMergedOrClosed {
```
crenshaw-dev, line 1119: Would status.state not be recorded as "merged" in the case that externallyMergedOrClosed is true?

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1080,153 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+		// A brand-new PR can be merged before the CTP's trailer-bearing update ever ran; there is nothing to record.
+		logger.V(4).Info("PR commit message has no trailers, skipping promotion history note")
+		return nil
+	}
+
+	mergedTargetSha := livePR.Status.MergedTargetSha
+	if mergedTargetSha == "" {
+		if livePR.Status.State == promoterv1alpha1.PullRequestMerged {
+			return fmt.Errorf("merged pull request %q has no status.mergedTargetSha", livePR.Name)
+		}
```
crenshaw-dev, line 1140:
```suggestion
		if livePR.Status.State == promoterv1alpha1.PullRequestMerged {
			// This will eventually be resolved by the PullRequest's own finalizer logic. Error to retry.
			return fmt.Errorf("merged pull request %q has no status.mergedTargetSha", livePR.Name)
		}
```

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1080,153 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+		logger.V(4).Info("PR externally merged or closed without mergedTargetSha, skipping promotion history note",
+			"prID", livePR.Status.ID)
```
```suggestion
		logger.V(4).Info("PR externally closed without mergedTargetSha, skipping promotion history note",
			"prID", livePR.Status.ID)
```

internal/controller/changetransferpolicy_controller.go
zachaller: no, status.state is explicitly set to "" when externallyMergedOrClosed is true, because a PR that has disappeared from the SCM can't be classified as merged vs. closed

I don't love the name for this particular case because it essentially a fallback for not able to find the PR which I think the only case is a delete or say some hide feature I think github might have released a hide something feature recently not sure if it was PRs though

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1080,153 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+	}
+	// Empty means the merge commit has no readable metadata for this activePath (missing or malformed); keep
+	// the snapshot rather than clobbering it with nothing.
+	if mergedDrySha == "" {
+		return nil
+	}
+
+	snapshotDrySha := getFirstTrailerValue(trailers, constants.TrailerShaDryProposed)
+	if snapshotDrySha == mergedDrySha {
+		return nil
```
crenshaw-dev, line 1201:
```suggestion
		// The actually-merged dry sha didn't change since the trailer snapshot. Keep the snapshot.
		return nil
```

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1080,153 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+
+	logger.Info("promotion history merge commit snapshot mismatch; using the merge commit as the source of truth",
+		"mergeCommit", mergedTargetSha, "snapshotDrySha", snapshotDrySha, "mergedDrySha", mergedDrySha)
+	recorder.Eventf(ctp, nil, "Warning", constants.PromotionHistoryNoteMergeCommitSnapshotMismatchReason, "WritingPromotionHistoryNote",
+		constants.PromotionHistoryNoteMergeCommitSnapshotMismatchMessage, livePR.Name, snapshotDrySha, mergedDrySha)
+
+	trailers[constants.TrailerShaDryProposed] = []string{mergedDrySha}
+	if mergedHydratedSha := mergeCommitSecondParent(ctx, gitOperations, mergedTargetSha); mergedHydratedSha != "" {
+		trailers[constants.TrailerShaHydratedProposed] = []string{mergedHydratedSha}
+	}
```
crenshaw-dev, line 1212: We should probably have a docs page for the git trailers and explicitly document which fields can behave differently between the note and the commit message.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1051,6 +1080,153 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
+// mergedProposedDrySha returns the dry sha the merge commit brought onto the active branch, read from
+// <activePath>/hydrator.metadata on the merge commit itself — the authoritative "what actually merged" value.
+// A missing or malformed metadata file returns "" so the caller keeps the trailer snapshot; genuine read
+// failures are returned as errors so finalization retries.
+func mergedProposedDrySha(ctx context.Context, gitOperations *git.EnvironmentOperations, mergedTargetSha, activePath string) (string, error) {
+	meta, err := gitOperations.GetShaMetadataFromFile(ctx, mergedTargetSha, activePath)
+	if err != nil {
+		if git.IsHydratorMetadataMalformed(err) {
+			log.FromContext(ctx).V(4).Info("malformed hydrator metadata on merge commit; keeping trailer snapshot",
+				"sha", mergedTargetSha, "err", err)
```
crenshaw-dev: Probably worth an error-level log? Or at least warn?

internal/controller/pullrequest_controller.go
```diff
@@ -305,40 +303,135 @@ func (r *PullRequestReconciler) syncStateFromProvider(ctx context.Context, pr *p
...
-			// The ExternallyMergedOrClosed flag means this PR is no longer open on the provider while we still
-			// desired open; that includes true external action and indistinguishable cases such as our delete finalizer
-			// having closed the SCM PR. An empty State with ExternallyMergedOrClosed=true means we cannot tell merge vs. close.
-			pr.Status.State = ""
-			return true, nil
+	return r.syncWhenNotFoundOpen(ctx, pr, provider)
+}
+
+// syncWhenNotFoundOpen reconciles status when FindOpen did not return an open PR but status.id is set.
+func (r *PullRequestReconciler) syncWhenNotFoundOpen(ctx context.Context, pr *promoterv1alpha1.PullRequest, provider scms.PullRequestProvider) (bool, error) {
```
crenshaw-dev: I'd just inline this function. Separating it makes it more difficult to follow how syncStateFromProvider works (e.g. I have to go back up to that function to understand what the bool return value means).

internal/controller/pullrequest_controller.go
```diff
@@ -305,40 +303,135 @@ func (r *PullRequestReconciler) syncStateFromProvider(ctx context.Context, pr *p
...
+	if pr.Status.ID == "" {
+		return false, nil
```
crenshaw-dev, line 339:
```suggestion
		// The PR wasn't found in the SCM, and we don't have an ID to explicitly Get it. 
		// We've done all we can do.
		return false, nil
```

internal/controller/pullrequest_controller.go
```diff
@@ -305,40 +303,135 @@ func (r *PullRequestReconciler) syncStateFromProvider(ctx context.Context, pr *p
...
+func (r *PullRequestReconciler) applyGetPullRequestDetails(ctx context.Context, pr *promoterv1alpha1.PullRequest, details scms.GetPullRequestResult) (requeue bool) {
+	logger := log.FromContext(ctx)
+
+	switch details.State {
+	case promoterv1alpha1.PullRequestMerged:
+		changed := pr.Status.State != promoterv1alpha1.PullRequestMerged
+		if details.MergedTargetSHA != "" && pr.Status.MergedTargetSha != details.MergedTargetSHA {
+			pr.Status.MergedTargetSha = details.MergedTargetSHA
+			changed = true
 		}
```
crenshaw-dev: Should we just make status.mergedTargetSha immutable after it's set to any non-empty value? I can't imagine a scenario where we'd want to set it to two different values during the lifecycle of a single PullRequest resource.

internal/controller/pullrequest_controller.go
```diff
@@ -305,40 +303,135 @@ func (r *PullRequestReconciler) syncStateFromProvider(ctx context.Context, pr *p
...
+// reconcileOpenAfterFindOpenMiss corrects status for a PR that FindOpen missed but Get reports as
+// still open. No immediate requeue is needed: the PR is open, so the rest of the reconcile
+// (terminal cleanup, state transitions, label sync) is safe to run against the corrected status in
+// this same pass, and the deferred status apply persists it at the end.
+func (r *PullRequestReconciler) reconcileOpenAfterFindOpenMiss(pr *promoterv1alpha1.PullRequest) {
+	if pr.Status.ExternallyMergedOrClosed != nil && *pr.Status.ExternallyMergedOrClosed {
+		pr.Status.ExternallyMergedOrClosed = new(false)
+	}
+	pr.Status.State = promoterv1alpha1.PullRequestOpen
+}
```
crenshaw-dev: Seems like it would be fine to inline this

internal/controller/pullrequest_controller.go
```diff
@@ -629,10 +722,16 @@ func (r *PullRequestReconciler) mergePullRequest(ctx context.Context, pr *promot
...
 	pr.Spec.Commit.Message = updatedMessage
 
-	if err := provider.Merge(ctx, *pr); err != nil {
+	result, err := provider.Merge(ctx, *pr)
+	if err != nil {
 		return err //nolint:wrapcheck // Error wrapping handled at top level
 	}
 	pr.Status.State = promoterv1alpha1.PullRequestMerged
+	// Providers that report the resulting target-branch commit in the merge response let us record it
+	// now; the rest leave it empty and syncWhenNotFoundOpen recovers it with a Get-by-ID lookup.
```
crenshaw-dev:
```suggestion
	// Providers that report the resulting target-branch commit in the merge response let us record it
	// now; the rest leave it empty and the next reconcile recovers it with a Get-by-ID lookup.
```

internal/git/git.go
```diff
@@ -202,6 +209,23 @@ func buildHydratorMetadataPath(activePath string) string {
...
+		strings.Contains(stderr, "exists on disk, but not in")
+}
+
+// IsHydratorMetadataMalformed reports whether GetShaMetadataFromFile failed because the blob was
+// present but not valid hydrator.metadata JSON. Callers that intentionally degrade on missing or
+// malformed metadata (for example findDrySha during PR finalization) should treat this as a
+// non-match rather than a retryable infrastructure error.
+func IsHydratorMetadataMalformed(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "could not unmarshal metadata file")
+}
```
crenshaw-dev: We should have `GetShaMetadataFromFile` return a typed error instead of inferring from the message

docs/debugging/git-trailers.md
```diff
@@ -0,0 +1,97 @@
...
+| `Commit-status-active-*`, `Commit-status-proposed-*` | **No — but they may describe the wrong revision.** | Gate phases cannot be reconstructed from Git, so the note keeps the snapshot values verbatim. After an external merge they may describe the gates that applied to an *earlier* proposed revision than the one that merged. |
+| `Promoter-merge-commit-snapshot-mismatch` | **Note only.** | Never written to a commit message. Its presence is the signal that the corrections above happened. |
+| `Pull-request-id`, `-url`, `-creation-time`, `-source-branch`, `-target-branch`, `Sha-dry-active`, `Sha-hydrated-active` | **No.** | Copied into the note verbatim from the snapshot. |
+
+So when `status.history[].mergeCommitSnapshotMismatch` is `true`, treat the proposed SHAs in that entry as trustworthy — they were re-read from the merge commit — and the commit statuses as possibly describing a superseded revision. The controller also emits [PromotionHistoryNoteMergeCommitSnapshotMismatch](../monitoring/events.md#changetransferpolicy) when it applies the correction.
```
crenshaw-dev: > treat the proposed SHAs in that entry as trustworthy — they were re-read from the merge commit

This seems misleading. Isn't it only trustworthy for merge commits, not squashes or fast-forwards?

zachaller: Yup it was corrected

internal/controller/changetransferpolicy_controller.go
```diff
@@ -1033,6 +1056,15 @@ func (r *ChangeTransferPolicyReconciler) handlePRFinalizerRemoval(ctx context.Co
...
 	}
 
+	// Write the promotion-history note before releasing the finalizer: the PR's commit message is the last
+	// place the trailers survive when the SCM rewrote the merge commit (squash or external merge), and the
+	// finalizer is what guarantees that data is still around. A failure here keeps the finalizer so the next
+	// reconcile retries; SetHistoryNote's overwrite semantics make the retry idempotent.
+	if err := r.writePromotionHistoryNote(ctx, ctp, gitOperations, &livePR); err != nil {
+		r.Recorder.Eventf(ctp, nil, "Warning", constants.PromotionHistoryNoteFailedReason, "WritingPromotionHistoryNote", constants.PromotionHistoryNoteFailedMessage, livePR.Name, err)
+		return fmt.Errorf("failed to write promotion history note for PullRequest %q: %w", livePR.Name, err)
+	}
```
crenshaw-dev, line 1080: There's a very narrow scenario that we currently mishandle:

1. Someone merges the PR in the SCM (no PullRequest reconcile yet)
2. Some external system deletes the PullRequest k8s resource (still no reconcile)
3. The CTP reconciles, spec.state == status.state == open, so we skip writing the note
4. CTP removes its finalizer
5. PullRequest _finally_ reconciles and sets its status to externally merged
6. PullRequest controller removes its finalizer, PR gets deleted without the CTP ever observing the final state

zachaller: fixed

internal/controller/pullrequest_controller.go
```diff
@@ -543,46 +652,111 @@ func (r *PullRequestReconciler) getPullRequestProvider(ctx context.Context, pr p
...
+	logger := log.FromContext(ctx)
+	holdsFinalizer := controllerutil.ContainsFinalizer(pr, promoterv1alpha1.PullRequestFinalizer)
+
+	// Still listed as open, so it cannot have merged and there is no terminal outcome left to learn.
+	// Closing it discharges the finalizer's obligation.
+	if openResult.Found && holdsFinalizer {
+		if err := r.closePullRequest(ctx, pr, provider); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to close pull request: %w", err) // Top-level wrap for close errors
+		}
+		return ctrl.Result{}, r.releaseFinalizer(ctx, pr)
```
crenshaw-dev: `closePullRequest` doesn't persist the updated state to status, so `releaseFinalizer` is premature. We should either immediately persist the status update or quit and requeue.

internal/controller/pullrequest_controller.go
```diff
@@ -543,46 +652,111 @@ func (r *PullRequestReconciler) getPullRequestProvider(ctx context.Context, pr p
...
+// then take the resource away before the merge commit was ever recorded, and with it the promotion
+// history note the ChangeTransferPolicy finalizer exists to write. So the finalizer is held until
+// status records a terminal outcome, which is either observed from the SCM or produced by closing the
+// pull request ourselves.
+//
+// State transitions, terminal cleanup, and label sync are deliberately unreachable from here: a
+// terminating PullRequest must never be created, merged, or relabeled on the SCM.
+func (r *PullRequestReconciler) reconcileDeletion(ctx context.Context, pr *promoterv1alpha1.PullRequest, provider scms.PullRequestProvider, openResult scms.FindOpenResult) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	holdsFinalizer := controllerutil.ContainsFinalizer(pr, promoterv1alpha1.PullRequestFinalizer)
```
crenshaw-dev: If we don't hold our own finalizer, shouldn't that indicate "we're done with this PR, quit early"?

crenshaw-dev (approved):

PR #1590, opened 2026-06-12
feat(pr): enforce terminal state

crenshaw-dev:
## Summary by CodeRabbit

* **New Features**
  * Pull requests that have reached terminal state (merged or closed) now have immutable state fields, preventing any further state transitions.

### DIFF
--- a/api/v1alpha1/pullrequest_types.go
+++ b/api/v1alpha1/pullrequest_types.go
@@ -151,6 +151,7 @@ func (ps *PullRequest) SetObservedGeneration(generation int64) {
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
 // +kubebuilder:printcolumn:name="URL",type=string,JSONPath=`.status.url`,priority=1
 // +kubebuilder:validation:XValidation:rule=`self.spec.state == 'open' || has(self.status.id) && self.status.id != ""`,message="Cannot transition to 'closed' or 'merged' state when status.id is empty"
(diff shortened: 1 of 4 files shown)


api/v1alpha1/pullrequest_types.go
```diff
@@ -151,6 +151,7 @@ func (ps *PullRequest) SetObservedGeneration(generation int64) {
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
 // +kubebuilder:printcolumn:name="URL",type=string,JSONPath=`.status.url`,priority=1
 // +kubebuilder:validation:XValidation:rule=`self.spec.state == 'open' || has(self.status.id) && self.status.id != ""`,message="Cannot transition to 'closed' or 'merged' state when status.id is empty"
+// +kubebuilder:validation:XValidation:rule=`!(has(oldSelf.status.state) && (oldSelf.status.state == 'merged' || oldSelf.status.state == 'closed')) || self.spec.state == oldSelf.spec.state`,message="spec.state is immutable once status.state has reached a terminal state (merged or closed)"
```
zachaller, line 154: Just a note not sure if it affects this close is not a terminal state from an SCMs point of view you can re-open and keep the same ID etc

crenshaw-dev: True... although I think a closed state pushes us into the delete/cleanup flow, so maybe we just clamp it down to that behavior? I'm trying to think about whether it makes sense to validate down to our "blessed" flows or just to SCM behavior.

crenshaw-dev:
Not sure I believe in this, think it's probably fine to just Code the Controller Right.

PR #1697, opened 2026-06-29
fix: reduce PR/CTP reconciles

zachaller:
### Summary
Stops a CTP → PullRequest → CTP loop that caused steady SCM PullRequest/list + update traffic while promotion is blocked on open PRs, even with a 60m requeue.

### Changes
 * Skip no-op PR applies: CTP compares the computed PR spec to the live object before SSA; unchanged specs are not patched.
 * Narrow Owns(PullRequest): CTP re-enqueues only on spec generation changes, finalizer changes, first PR ID, or terminal/externally-closed state — not routine status updates (URL, Ready, etc.).

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced ChangeTransferPolicy reconciliations by narrowing Pull Request event handling to relevant create/delete events and specific meaningful updates.
  * Prevents reconciliations from triggering on status-only changes (e.g., URL-only churn).
* **Bug Fixes**
  * Improves synchronization by enqueueing only when key finalizer/status fields change meaningfully, including initial `status.id` population and CTP finalizer removal.
* **Tests**
  * Added steady-state coverage to confirm no unnecessary SCM polling/update calls during status-only churn.
  * Enhanced fake SCM metrics and timing behavior to support stronger call-count and state assertions.

### DIFF
--- a/internal/controller/changetransferpolicy_controller.go
+++ b/internal/controller/changetransferpolicy_controller.go
@@ -1241,11 +1241,6 @@ func (r *ChangeTransferPolicyReconciler) createOrUpdatePullRequest(ctx context.C
 		prState = existingPR.Spec.State
 	}
 
-	if prExists && pullRequestManagedByCTPUnchanged(existingPR, title, description, commitMessage, ctp, prState) {
-		logger.V(4).Info("Pull request spec unchanged, skipping apply", "pullRequest", prName)
-		return existingPR, nil
-	}
-
 	// Build the apply configuration
 	prApply := acv1alpha1.PullRequest(prName, ctp.Namespace).
 		WithLabels(map[string]string{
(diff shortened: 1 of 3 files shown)


internal/controller/changetransferpolicy_controller.go
```diff
@@ -1479,6 +1484,90 @@ func TemplatePullRequest(prt promoterv1alpha1.PullRequestTemplate, data map[stri
...
+			newPR, ok := e.ObjectNew.(*promoterv1alpha1.PullRequest)
+			if !ok || newPR == nil {
+				return false
+			}
+			if oldPR.Generation != newPR.Generation {
+				return true
+			}
+			if len(oldPR.Finalizers) != len(newPR.Finalizers) {
+				return true
+			}
```
crenshaw-dev: This would miss a change that both adds and removes a finalizer... probably would never happen unless somehow two change events got coalesced into one.

crenshaw-dev (approved):

PR #1705, opened 2026-06-30
feat: controller isolation by instance-id

crenshaw-dev:
## Summary by CodeRabbit

* **New Features**
  * Added multi-install partitioning via `promoter.argoproj.io/instance-id`, scoping controller reconciliation and cache behavior.
  * Introduced `status.instanceID` across Promoter CRDs, and expanded `ControllerConfiguration` status with `observedGeneration` and `conditions`.
* **Bug Fixes**
  * Improved propagation of `instance-id` to generated resources and ensured `status.instanceID` is stamped on every reconcile attempt (including error cases).
* **Documentation**
  * Updated multi-install, labeling/debugging, and status update guidance, including a migration runbook.
* **Tests**
  * Added/extended unit coverage and an end-to-end instance-id migration test.


crenshaw-dev:
@coderabbitai review

internal/settings/instanceid.go
```diff
@@ -0,0 +1,46 @@
...
+// InstanceIDsEqual reports whether two optional instance ID values are equivalent.
+// Both nil means default install; one nil and one non-nil are different.
+func InstanceIDsEqual(a, b *string) bool {
+	if a == nil && b == nil {
+		return true
+	}
+	if a == nil || b == nil {
+		return false
+	}
+	return *a == *b
```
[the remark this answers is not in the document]
crenshaw-dev: Only unset should preserve existing behavior. An empty label is not valid, and is documented as basically orphaning the resource.

api/v1alpha1/revertcommit_types.go
```diff
@@ -34,10 +34,7 @@ type RevertCommitSpec struct {
 }
 
 // RevertCommitStatus defines the observed state of RevertCommit
-type RevertCommitStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-}
+type RevertCommitStatus struct{}
```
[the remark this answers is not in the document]
crenshaw-dev: RevertCommit hasn't been implemented, don't bother finishing it.

internal/controller/controllerconfiguration_controller.go
```diff
@@ -28,32 +29,52 @@ import (
...
 	Scheme *runtime.Scheme
+
+	Shutdown            context.CancelFunc
+	StartupInstanceID   *string
+	ControllerNamespace string
 }
 
 // +kubebuilder:rbac:groups=promoter.argoproj.io,resources=controllerconfigurations,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=promoter.argoproj.io,resources=controllerconfigurations/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=promoter.argoproj.io,resources=controllerconfigurations/finalizers,verbs=update
```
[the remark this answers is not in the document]
crenshaw-dev: The controller does need the ability to set the status field, but not the spec.

internal/controller/controllerconfiguration_controller.go
```diff
@@ -19,41 +19,85 @@ package controller
...
+func (r *ControllerConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	logger := log.FromContext(ctx)
+	logger.Info("Reconciling ControllerConfiguration")
+	startTime := time.Now()
+
+	// Only the shipped configuration in the install namespace drives partition management; ignore
+	// any other ControllerConfiguration objects entirely (no status is written for them).
+	if req.Namespace != r.ControllerNamespace || req.Name != settings.ControllerConfigurationName {
+		return ctrl.Result{}, nil
+	}
```
crenshaw-dev: Could this just be a predicate?

docs/multi-install.md
```diff
@@ -0,0 +1,276 @@
...
+  # … other controller settings unchanged
+```
+
+**Default install** — omit `instanceID` (do not set it to `""`):
+
+```yaml
+spec:
+  promotionStrategy:
+    # …
+```
```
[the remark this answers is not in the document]
crenshaw-dev: done

config/rbac/role.yaml
```diff
@@ -100,3 +97,23 @@ rules:
...
+- apiGroups:
+  - promoter.argoproj.io
+  resourceNames:
+  - promoter-controller-configuration
+  resources:
+  - controllerconfigurations
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
crenshaw-dev: done

docs/multi-install.md
```diff
@@ -0,0 +1,274 @@
...
+- **Coordinate across resources** — label all roots (`PromotionStrategy`, gate CRs, and any other Promoter CRs this install owns) with the **same** value before restart. Do not relabel a `PromotionStrategy` while shrinking a gate's `spec` in parallel.
+- **Install config is separate** — set `ControllerConfiguration.spec.instanceID` in its own step. Single-replica installs restart automatically; HA installs need a rolling restart of all controller pods afterward.
+- **Cross-CR coordination is runbook-only** — the API does not enforce ordering across multiple resources; follow the steps below deliberately.
+
+### Default install → multi-install
+
+1. **Label roots only** — add `promoter.argoproj.io/instance-id: <your-id>` to every Promoter CR this install should manage (`PromotionStrategy`, `TimedCommitStatus`, `GitCommitStatus`, `WebRequestCommitStatus`, `ArgoCDCommitStatus`, and others as needed). Label referenced `Secret` objects (SCM, HTTP auth, kubeconfig) with the same value. Use metadata-only patches.
```
zachaller: GitRepository and ScmProvider are in the partitioned cache list
(`internal/cache/instanceid.go`), but they're user-created roots — nothing
propagates the label onto them. Step 1 doesn't name them, and the "Confirm
roots are labeled" example further down only checks the strategy + gate kinds.
If someone follows this list as written, their PromotionStrategies get
relabeled but the GitRepository stays unlabeled, and after the restart every
CTP reconcile fails NotFound on a GitRepository that exists — promotions stop
with a pretty confusing error. Can we name them explicitly?

```suggestion
1. **Label roots only** — add `promoter.argoproj.io/instance-id: <your-id>` to every Promoter CR this install should manage (`PromotionStrategy`, `GitRepository`, `ScmProvider`/`ClusterScmProvider`, `TimedCommitStatus`, `GitCommitStatus`, `WebRequestCommitStatus`, `ArgoCDCommitStatus`, and others as needed). Label referenced `Secret` objects (SCM, HTTP auth, kubeconfig) with the same value. Use metadata-only patches.
```

The verification example around line 193 should get the same treatment:

```bash
kubectl get promotionstrategy,gitrepository,scmprovider,timedcommitstatus,gitcommitstatus,webrequestcommitstatus,argocdcommitstatus \
  -n team-a -l promoter.argoproj.io/instance-id=wave-0
```

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/cache/instanceid.go
```diff
@@ -0,0 +1,93 @@
...
+	&promoterv1alpha1.GitCommitStatus{},
+	&promoterv1alpha1.TimedCommitStatus{},
+	&promoterv1alpha1.WebRequestCommitStatus{},
+	&promoterv1alpha1.ArgoCDCommitStatus{},
+	&promoterv1alpha1.RevertCommit{},
+}
+
+// partitionedSecretObject is the cache.ByObject key for corev1.Secret. A package-level pointer
+// is required because controller-runtime matches ByObject map keys by pointer identity.
+var partitionedSecretObject client.Object = &corev1.Secret{}
```
zachaller, line 34: Partitioning Secrets by label interacts badly with the ScmProvider secret
finalizers. `removeSecretFinalizerForProvider` (finalizer_utils.go) treats a
cache `NotFound` as "secret is gone, skip removal". With the migration steps
in this PR's docs, that's reachable:

1. Default install adds `scmprovider…/secret-finalizer` to Secret S.
2. Operator labels S with `instance-id: wave-0` (runbook step 1) while the
   ScmProvider is still in the default partition.
3. Operator deletes the ScmProvider and S. The default install's cached `Get`
   returns NotFound → it skips finalizer removal → S is stuck in
   `Terminating` forever, because wave-0 only removes the finalizer when one
   of *its own* ScmProviders referencing S is deleted.

The mirror image also holds: `ensureSecretFinalizer` no-ops on NotFound, so a
wave-0 ScmProvider referencing a not-yet-labeled Secret silently gets no
deletion protection at all.

Since the finalizer helpers are always name-targeted single-object ops, could
they go through the APIReader (uncached) instead of the partitioned cache?
That keeps the partition for list/watch semantics but makes finalizer
bookkeeping see the truth. Alternatively the docs need a hard warning that a
Secret must not be relabeled while another partition's provider still holds a
finalizer on it.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/cache/instanceid.go
```diff
@@ -0,0 +1,93 @@
...
+			controllerNamespace: {},
+		},
+		Field: fields.OneTermEqualSelector("metadata.name", settings.ControllerConfigurationName),
+	}
+	return cache.Options{ByObject: byObject}
+}
+
+func instanceIDSelector(instanceID *string) labels.Selector {
+	if instanceID == nil {
+		req, err := labels.NewRequirement(promoterv1alpha1.InstanceIDLabel, selection.DoesNotExist, nil)
```
zachaller, line 62: A resource whose label value matches no running install falls into a black
hole. `instance-id: ""` is a valid label *value* to the API server (easy to
produce from a Helm template that renders empty), but the default install's
`DoesNotExist` selector excludes it (the key exists) and no multi-install can
ever match it (`spec.instanceID` has minLength=1). Same for a typo'd value or
a decommissioned install. Result: never reconciled, no status, no condition,
no event — and a `PullRequest` carrying its finalizer becomes undeletable
(`kubectl delete` hangs forever with no controller able to process it).

We can't validate labels via CRD schema, so at minimum this deserves a
troubleshooting entry in multi-install.md ("resource has an instance-id value
no install serves — including the empty string"), and ideally something
operational, e.g. a periodic check or kubectl one-liner in the docs:

```bash
# CRs whose instance-id matches no known install (empty value included)
kubectl get promotionstrategy,changetransferpolicy,commitstatus,pullrequest -A \
  -l 'promoter.argoproj.io/instance-id,promoter.argoproj.io/instance-id notin (wave-0,wave-1)'
```

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/controller/gitcommitstatus_controller.go
```diff
@@ -405,8 +405,12 @@ func (r *GitCommitStatusReconciler) upsertCommitStatus(ctx context.Context, gcs
 	gvk := promoterv1alpha1.GroupVersion.WithKind(kind)
 
 	// Build the apply configuration
+	commitStatusLabels := utils.CopyInstanceIDLabelToMap(
```
zachaller: This `CopyInstanceIDLabelToMap(parent, CommitStatusStandardLabels(parent, …))`
two-step is repeated at four gate-controller call sites, and
developing-a-commitstatus.md teaches the same dance. Since
`CommitStatusStandardLabels` already takes the parent object, can we fold the
copy in and make propagation impossible to forget?

```go
// CommitStatusStandardLabels returns the labels gate controllers set on each CommitStatus:
// parent gate, environment branch, commit-status key, and (when present) instance-id.
func CommitStatusStandardLabels(parent client.Object, branch, commitStatusKey string) map[string]string {
	return CopyInstanceIDLabelToMap(parent, map[string]string{
		CommitStatusGateLabelKeyForParent(parent): KubeSafeLabel(parent.GetName()),
		promoterv1alpha1.EnvironmentLabel:         KubeSafeLabel(branch),
		promoterv1alpha1.CommitStatusLabel:        commitStatusKey,
	})
}
```

Then the four wrappers here, in timedcommitstatus, webrequestcommitstatus, and
argocdcommitstatus collapse to a plain `CommitStatusStandardLabels` call, and
the contributing doc loses a step. The failure mode this removes is nasty: a
future gate controller that calls `CommitStatusStandardLabels` alone creates
children the partitioned cache can't see — gates hang with no error anywhere.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/cache/instanceid.go
```diff
@@ -0,0 +1,93 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// promotorCRDObjects lists every promoter.argoproj.io type reconciled by the controller that
+// participates in instance-id partitioning. ControllerConfiguration is scoped separately by
+// install namespace (see partitionedControllerConfigurationObject).
+var promotorCRDObjects = []client.Object{
```
zachaller: The partitioning contract rests on this hand-maintained list, and
`instanceid_test.go` builds its expectations by iterating
`PartitionedObjects()` — the same slice `OptionsForInstanceID` consumes — so a
type missing from the list can never fail CI. We add gate CRDs fairly
regularly; the next one that's forgotten here gets an *unpartitioned*
informer, and two installs reconcile it concurrently (duplicate SCM writes —
exactly what this feature exists to prevent).

Suggest making the test independent of the production slice, e.g.:

```go
// Fails when a new promoter.argoproj.io kind is registered but not partitioned.
func TestPartitionCoversAllPromoterKinds(t *testing.T) {
	partitioned := map[string]bool{}
	for _, obj := range cache.PartitionedObjects() {
		gvk, _ := apiutil.GVKForObject(obj, utils.GetScheme())
		partitioned[gvk.Kind] = true
	}
	for gvk := range utils.GetScheme().AllKnownTypes() {
		if gvk.Group != promoterv1alpha1.GroupVersion.Group || !isReconciledKind(gvk.Kind) {
			continue // skip List types, ControllerConfiguration, etc.
		}
		if !partitioned[gvk.Kind] {
			t.Errorf("kind %s is registered in the scheme but not instance-id partitioned", gvk.Kind)
		}
	}
}
```

(Or simply assert against an independent literal list of kind names.) That
also lets the three `Exported for tests` accessors go away — with the current
setup they only enable the tautology. Also `promotorCRDObjects` → typo,
should be `promoterCRDObjects`.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/controller/controllerconfiguration_controller.go
```diff
@@ -19,49 +19,90 @@ package controller
...
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, fmt.Errorf("get ControllerConfiguration: %w", err)
+	}
+
+	// Remove any existing Ready condition. We want to start fresh.
+	previousReady = utils.RemoveReadyCondition(&cc)
+
+	if settings.InstanceIDsEqual(r.StartupInstanceID, cc.Spec.InstanceID) {
```
zachaller: The drift check only runs when this reconciler gets an event, and only on the
active leader. In HA, followers keep the boot-time cache partition; if
`spec.instanceID` changes and a stale follower later wins the lease, *all* its
controllers start with the wrong partition and reconcile resources now owned
by another install until this reconcile happens to fire — duplicate PRs /
hydration commits at the SCM, which SSA idempotency doesn't protect. The docs
say "roll the deployment", but nothing enforces it.

One cheap mitigation: a leader-gated runnable that re-verifies the instance ID
at election time, before/alongside controller startup:

```go
// cmd/main.go, after settingsMgr is created
if err := localManager.Add(&instanceIDGuard{
	cfg: restConfig, namespace: controllerNamespace,
	startup: instanceID, shutdown: shutdown,
}); err != nil { … }

type instanceIDGuard struct {
	cfg       *rest.Config
	namespace string
	startup   *string
	shutdown  context.CancelFunc
}

func (g *instanceIDGuard) NeedLeaderElection() bool { return true }

func (g *instanceIDGuard) Start(ctx context.Context) error {
	current, err := settings.ReadInstanceID(ctx, g.cfg, g.namespace)
	if err != nil {
		return fmt.Errorf("verifying instanceID on leadership acquisition: %w", err)
	}
	if !ptr.Equal(g.startup, current) {
		g.shutdown()
	}
	return nil
}
```

It doesn't fully close the race (the GET runs concurrently with the first
reconciles), but it shrinks the exposure from "until the CC controller's first
reconcile" to a single direct read at election time, and it covers the case
where the CC event was consumed by the previous leader.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/cache/instanceid.go
```diff
@@ -0,0 +1,93 @@
...
+	&promoterv1alpha1.PromotionStrategy{},
+	&promoterv1alpha1.ChangeTransferPolicy{},
+	&promoterv1alpha1.CommitStatus{},
+	&promoterv1alpha1.PullRequest{},
+	&promoterv1alpha1.ScmProvider{},
+	&promoterv1alpha1.ClusterScmProvider{},
```
zachaller, line 23: Cluster-scoped and shared resources can only carry **one** instance-id value,
so putting `ClusterScmProvider` and `Secret` in the partition means they can
never be shared across installs: labeling a shared ClusterScmProvider for
wave-0 makes it NotFound to the default install, and relabeling a shared
kubeconfig Secret shows up as a synthetic DELETE in the other install's
kubeconfig-provider watch — its remote-cluster Application events silently
stop and ArgoCDCommitStatus gates hang. That may be an acceptable constraint,
but multi-install.md doesn't say it, and never mentions ClusterScmProvider at
all. Can we document "shared credentials/providers across installs are not
supported — duplicate them per install", or exclude the kubeconfig secrets
from partitioning if sharing them is meant to work?

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/utils/utils.go
```diff
@@ -477,6 +474,9 @@ func HandleReconciliationResult(
 		recorder.Eventf(obj, nil, eventType, persistedReady.Reason, "Reconciling", persistedReady.Message)
 	}()
 
+	// Stamp status.instanceID on every reconcile attempt so consumers can see which install
+	// is actively reconciling this resource, including when Ready=False.
+	obj.SetStatusInstanceID(InstanceIDStatusValue(obj))
```
zachaller: Right now `status.instanceID` is stamped from the object's **own** label, and
the cache guarantees a controller only ever sees objects matching its
partition — so status == label by construction and the field can't tell us
anything the label doesn't. If we stamp the **controller's configured**
instance ID instead, the field becomes a real cross-install-write detector: in
the two partition-leak cases that can actually happen (a CRD missing from the
partition list, a stale HA follower after failover), the leaking install
writes *its* ID into status and the label/status mismatch flags the bug.

```go
// internal/utils (value set once at startup, before the manager runs)
var controllerInstanceID *string

func SetControllerInstanceID(id *string) { controllerInstanceID = id }

// HandleReconciliationResult
obj.SetStatusInstanceID(controllerInstanceID)
```

(Or thread it through settings.Manager if we'd rather avoid the package
global.) The multi-install.md verification table would then read
"`status.instanceID` — which install last wrote status; must equal the label,
mismatch indicates a partitioning leak", which is a much stronger check than
the current mirror.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/utils/labels.go
```diff
@@ -0,0 +1,45 @@
+package utils
+
+import (
+	promoterv1alpha1 "github.com/argoproj-labs/gitops-promoter/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CopyInstanceIDLabel copies promoter.argoproj.io/instance-id from parent to child when the
+// parent carries a non-empty value. No-op when the parent label is missing or empty.
+func CopyInstanceIDLabel(parent, child client.Object) {
```
zachaller: nit: this variant has no production callers — all eight propagation sites use
`CopyInstanceIDLabelToMap`. Suggest dropping it (a one-line wrapper can come
back if an object-based call site ever appears). Related test dedupe: the
Describe blocks for these helpers in `internal/utils/commitstatus_test.go`
duplicate the ones in `internal/utils/labels_test.go`, and the
`OptionsForInstanceID` Describe in
`internal/controller/controllerconfiguration_controller_test.go` duplicates
`internal/cache/instanceid_test.go` inside the slow envtest suite — one copy
of each is enough.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/settings/manager.go
```diff
@@ -17,6 +19,41 @@ const (
...
+	}, cc); err != nil {
+		return nil, fmt.Errorf("failed to get ControllerConfiguration %q: %w", ControllerConfigurationName, err)
+	}
+
+	return cc.Spec.InstanceID, nil
+}
+
+// InstanceIDsEqual reports whether two optional instance ID values are equivalent.
+// Both nil means default install; one nil and one non-nil are different.
+func InstanceIDsEqual(a, b *string) bool {
```
zachaller: nit: this is exactly `ptr.Equal` from `k8s.io/utils/ptr`, which we already
import elsewhere (changetransferpolicy_controller.go). The single call site in
controllerconfiguration_controller.go can use `ptr.Equal(r.StartupInstanceID,
cc.Spec.InstanceID)` and this helper plus its table test can go.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/controller/promotionstrategy_controller_test.go
```diff
@@ -5852,3 +5857,72 @@ var _ = Describe("PromotionStrategy Bug Tests", func() {
...
+				return true
+			})
+			if !found {
+				t.Fatalf("%s must call utils.CopyInstanceIDLabelToMap for instance-id label propagation", file)
+			}
+		})
+	}
+}
+
+func TestNoInstanceIDPredicateOnControllers(t *testing.T) {
```
zachaller: nit: this greps controller sources for `promoterpredicate.InstanceID`, but
that symbol doesn't exist anywhere in the codebase, so the test can never
fail — and anyone building the same anti-pattern under a different name walks
right past it. The design decision ("cache ByObject is the partition
boundary, no per-controller predicates") reads better as the doc comment on
`cache.OptionsForInstanceID`, which already says it. Suggest deleting the
test.

---
_Generated by [Claude Code](https://claude.ai/code)_

zachaller (commented):
Review comments on the instance-id partitioning feature, focused on migration-doc gaps and a few edge cases in the partitioning logic.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/cache/instanceid.go
crenshaw-dev: I don't think this is valid. If someone updates the Secret instance-id but not the referencing ScmProvider's instance-id, then they're already in a broken state.

Right now we only allow instances to "reach outside" their instance scope for write operations in service of migrating child resources to the new instance. The suggested fix would expand that to "cases where the user messed up," by failing to update all relevant resources, and that quickly gets into a mess. I think the current behavior is fine. Worst case scenario, someone realizes their mistake and either updates the ScmProvider's instance-id or manually removes the finalizer to let the Secret be deleted.

internal/cache/instanceid.go
crenshaw-dev: Docs make sense. In the future we'll probably have a validating webhook. This would be a good candidate for validation.

It's one point in favor of using `spec.instanceID`, but I think the scales still favor a label for now.

internal/cache/instanceid.go
crenshaw-dev: They're not shared, user must duplicate. We can add more docs.

internal/utils/utils.go
crenshaw-dev: It's kind of weak protection, since we detect drift at the _end_ of the reconcile. We could consider adding an additional safeguard at the top. But that's a lot of duplicated code for a pretty far edge case.

internal/cache/instanceid.go
zachaller: Yea I think not to fix maybe document that finalizers might need to be manually removed in cases of instance id re-labeling?

zachaller: could be un-common enough to not need docs or the concepts familiar enough to not?

internal/controller/promotionstrategy_controller_test.go
```diff
@@ -5852,3 +5857,51 @@ var _ = Describe("PromotionStrategy Bug Tests", func() {
...
+// which use CommitStatusStandardLabels).
+var childLabelPropagationSites = []struct {
+	file  string
+	calls int
+}{
+	{file: "promotionstrategy_controller.go", calls: 2},
+	{file: "changetransferpolicy_controller.go", calls: 1},
+}
+
+func TestChildCreationPropagatesInstanceIDLabel(t *testing.T) {
```
zachaller: Ginkgo

internal/cache/instanceid.go
crenshaw-dev: I think uncommon enough to not need docs. If you have a label mismatch between a resource and its finalizer owner, other stuff is gonna break before you run into cleanup problems.

zachaller (approved):

PR #1713, opened 2026-07-01
chore: minimize rbac

crenshaw-dev:
RBAC was broader than we needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined finalizer update retry handling so finalizer changes are applied consistently for Secrets and other resources.
* **Security / RBAC**
  * Tightened RBAC across controllers by narrowing main-resource verbs to read-only where possible and constraining status/finalizer write access to only what’s required.
  * Reworked granular promotion-related permissions to reduce broad access.
* **Documentation**
  * Expanded contributing guidance on controller-scoped RBAC for finalizers and status subresources.


zachaller (approved):

config/rbac/role.yaml
```diff
@@ -42,39 +41,29 @@ rules:
...
   - scmproviders
   - timedcommitstatuses
   - webrequestcommitstatuses
   verbs:
-  - create
-  - delete
   - get
   - list
-  - patch
-  - update
```
samuelmasuy, line 63: Thanks for the RBAC minimization. I think there is a finalizer permission mismatch.
`handleResourceFinalizerWithDependencies` still calls `c.Update(ctx, obj)` to add/remove CRD finalizers, but this PR removes update from primary-resource RBAC for GitRepository and ChangeTransferPolicy.
We see cannot update resource "gitrepositories" during reconciliation. Since CRDs have no /finalizers subresource, should those resources retain primary-resource update?

crenshaw-dev: Good catch, working on fixes here: https://github.com/argoproj-labs/gitops-promoter/pull/1769

PR #1714, opened 2026-07-01
fix: reduce reconciles

zachaller:
## Summary by CodeRabbit

* **New Features**
  * Improved efficiency for background pull request and branch status checks by probing the remote for branch SHA matches and skipping unnecessary fetches when nothing has changed.
* **Bug Fixes**
  * Avoids redundant title/description synchronization when a pull request generation is already marked ready.
  * Prevents extra branch retrieval work when the remote SHA can be confirmed, with safe fallback to full retrieval if it can’t.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -210,27 +210,31 @@ func (g *EnvironmentOperations) GetBranchShas(ctx context.Context, branch, activ
 	return g.getBranchShas(ctx, branch, activePath, false)
 }
 
-// LsRemoteBranches returns the current remote SHAs for the given branches without touching this
-// environment's local clone. It's a thin wrapper around the package-level LsRemote using this
-// environment's own GitOperationsProvider and GitRepository, for callers that want to cheaply
-// check whether a branch changed before paying for a full fetch (see
-// GetBranchShasSkipFetchIfUnchanged).
-func (g *EnvironmentOperations) LsRemoteBranches(ctx context.Context, branches ...string) (map[string]string, error) {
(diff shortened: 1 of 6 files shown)


internal/git/git.go
```diff
@@ -207,6 +207,38 @@ func buildHydratorMetadataPath(activePath string) string {
...
+// successful call). If it matches, the commit is guaranteed to already be present in this clone (it
+// was fetched the last time this identity observed that SHA), so the network fetch is skipped and
+// rev-parse/ls-tree resolve it from the existing local objects instead.
+//
+// Unlike a caller-supplied skip flag, there is no way to skip the fetch without this method having
+// just confirmed the SHA is unchanged, so callers cannot accidentally skip based on a stale probe or
+// a SHA observed by a different clone/identity. Pass an empty lastKnownHydratedSha (e.g. before any
+// SHA has been observed for this branch) to always fetch; the probe is skipped in that case since
+// there's nothing to compare against.
+func (g *EnvironmentOperations) GetBranchShasIfUnchanged(ctx context.Context, branch, activePath, lastKnownHydratedSha string) (BranchShas, error) {
```
crenshaw-dev: The only non-test callers of `GetBranchShas` supply the `lastKnownHydratedSha` param. Instead of extracting the `getBranchShas` method and creating  `GetBranchShasIfUnchanged`, I'd just fold all the logic into the existing function. Pass an empty string from tests if the param is irrelevant.

crenshaw-dev (approved):

PR #1736, opened 2026-07-07
chore(ci): assert there are no go tests

crenshaw-dev:
## Summary by CodeRabbit

* **Tests**
  * Standardized testing on the project’s BDD-style framework and added a repository-wide check to prevent new legacy Go unit tests from being introduced.
  * Updated an existing URL-generation test to match the new test style while keeping the same expected behavior.

* **Chores**
  * Wired the new test validation step into local and automated test workflows so it runs consistently before test execution.


crenshaw-dev:
Looks fraught

PR #1744, opened 2026-07-08
feat: add ScheduledCommitStatus CRD for cron-based promotion gating

### LINKED ISSUE #1734: feat: PromotionWindowCommitStatus — cron-based time-window gates for promotions
## Summary

Add a new `PromotionWindowCommitStatus` CRD that gates promotions on cron-defined time windows. The commit status reports **success** when the current time falls within an allowed window (and outside any exclusion window), and **pending** otherwise. This enables maintenance-window-style controls — e.g., "only promote to production Mon–Fri 09:00–17:00 UTC" or "block promotions during change freezes."

This is the "Deployment window gates" item listed on the [roadmap](https://gitops-promoter.readthedocs.io/en/stable/roadmap/) under *Commit status and gating*, which currently has no tracking issue.

## Motivation

Duration-based soak time (`TimedCommitStatus`) answers "has the commit been running long enough?" but not "is now a safe time to promote?" Organizations need calendar-aware gates for:

- **Change management** — restrict production promotions to business hours or approved maintenance windows.
- **Release freezes** — block all promotions during code-freeze periods.
- **Off-hours safety** — prevent promotions when the on-call team has limited capacity.
- **Regional compliance** — gate on local-timezone business hours for region-specific environments.

Today this can be approximated with `WebRequestCommitStatus` polling an external scheduler, but a built-in CRD would be simpler, declarative, and self-contained.

## Proposed API

Following the established CommitStatus CRD pattern (`PromotionStrategyRef` + `Key` + per-environment config):

```yaml
kind: PromotionWindowCommitStatus
metadata:
  name: production-business-hours
spec:
  key: promotion-window
  promotionStrategyRef:
    name: webservice-tier-1
  environments:
    - branch: environment/production
      timezone: "UTC"                  # Optional, defaults to UTC
      windows:
        - cron: "0 9 * * 1-5"         # Monday–Friday at 09:00 UTC
          duration: 8h                 # Window open for 8 hours
        - cron: "0 10 * * 6"          # Saturday at 10:00 UTC
          duration: 4h                 # Shorter weekend window
      exclusions:
        - cron: "0 0 24 12 *"         # Christmas Eve
          duration: 48h
```

### Default behavior

**Environments without an entry default to success (24/7 open).** Only environments explicitly listed are gated. This means a `PromotionWindowCommitStatus` only blocks environments you explicitly configure — all others pass through.

### Allow windows, exclusion windows, or both

The API supports three configuration styles:

1. **Allow-only** — define `windows`, no `exclusions`. Promotion allowed only during listed windows.
2. **Exclusion-only** — define `exclusions`, no `windows`. Promotion allowed anytime *except* during exclusion windows (change freezes, holidays, etc.).
3. **Both** — define `windows` and `exclusions`. Promotion allowed when inside an allow window AND outside all exclusion windows. Exclusions take precedence over allow windows.

```yaml
# Example: exclusion-only config — block during change freezes, allow anytime else
spec:
  key: change-freeze
  promotionStrategyRef:
    name: webservice-tier-1
  environments:
    - branch: environment/production
      exclusions:
        - cron: "0 0 20 12 *"         # Annual change freeze: Dec 20
          duration: 336h               # Through Jan 2
        - cron: "0 18 * * 5"          # Friday 18:00 → Monday 06:00 weekend freeze
          duration: 60h
```

### Spec fields

| Field | Type | Description |
|-------|------|-------------|
| `promotionStrategyRef` | `ObjectReference` | Reference to the PromotionStrategy (standard pattern) |
| `key` | `string` | Gate key referenced in PromotionStrategy (default: `promotion-window`) |
| `environments[].branch` | `string` | Branch/environment to gate |
| `environments[].timezone` | `string` | IANA timezone for cron evaluation (default: `UTC`) |
| `environments[].windows[]` | `[]Window` | Allowed promotion windows (OR semantics — any open window suffices) |
| `environments[].exclusions[]` | `[]Window` | Blocked periods that override allow windows (OR semantics) |
| `Window.cron` | `string` | Cron expression defining the start of the window |
| `Window.duration` | `Duration` | How long the window remains open/blocked after the cron trigger |

### Phase logic

```
if environment not listed → success (24/7)
if inside any exclusion    → pending (blocked)
if no windows defined      → success (exclusion-only mode)
if inside any window       → success (allowed)
else                       → pending (outside all windows)
```

### No override mechanism

To handle emergency deployments or unplanned changes, modify the `PromotionWindowCommitStatus` spec directly (e.g., add a temporary allow window or remove an exclusion). The controller reconciles immediately on spec changes. This keeps the system declarative and auditable — no out-of-band override state.

### Status

```yaml
status:
  observedGeneration: 3
  environments:
    - branch: environment/production
      phase: success            # or "pending"
      activeWindow:             # set when inside an allow window
        start: "2026-07-07T09:00:00Z"
        end: "2026-07-07T17:00:00Z"
      activeExclusion: null     # set when inside an exclusion window
      nextTransition: "2026-07-07T17:00:00Z"
  conditions:
    - type: Ready
      status: "True"
```

## Usage with PromotionStrategy

```yaml
kind: PromotionStrategy
metadata:
  name: webservice-tier-1
spec:
  activeCommitStatuses:
    - key: argocd-health
    - key: timer
    - key: promotion-window   # ← blocks promotion outside allowed windows
  environments:
    - branch: environment/development
    - branch: environment/staging
    - branch: environment/production
```

## Open questions

- **Naming**: `PromotionWindowCommitStatus` vs `WindowCommitStatus` vs `ScheduleCommitStatus`?
- **Cron library**: Use `robfig/cron` (already common in the Go ecosystem) or a simpler parser?
- **Validation**: Should we enforce that at least one of `windows` or `exclusions` is non-empty per environment entry?

patjlm:
## Summary

Implements the `ScheduledCommitStatus` CRD — a new CommitStatus gate that controls promotions based on cron-defined time windows with allow/exclude semantics.

Closes #1734

### Key design decisions

- **Environments not listed default to success** (24/7 open) — only explicitly configured environments are gated
- **Global + per-environment windows** — `spec.allow`/`spec.exclude` apply to all listed environments, merged with per-environment windows (OR semantics for allow, union for exclude)
- **Exclusion windows override allow windows** — enables change-freeze definitions that take priority
- **Exclusion-only mode** — configs with only exclusions allow promotions anytime except during exclusion windows
- **Per-environment timezone** — each environment can use a different IANA timezone for cron evaluation
- **CEL validation** — each environment must have at least one allow or exclude window (either per-env or global)
- **No override mechanism** — modify the spec directly to change behavior (keeps the API surface small)
- **Precise requeue** at next transition time + 1s buffer instead of fixed-interval polling
- **`robfig/cron/v3`** for 5-field cron parsing (same library Kubernetes uses)

### Phase logic

```
if environment not listed → success (24/7)
if inside any exclusion    → pending (blocked)
if no allow windows        → success (exclusion-only mode)
if inside any allow window → success (allowed)
else                       → pending (outside all windows)
```

### Example

```yaml
kind: ScheduledCommitStatus
metadata:
  name: deploy-window
spec:
  promotionStrategyRef:
    name: my-strategy
  key: promotion-window
  exclude:
    - cron: "0 0 25 12 *"       # Christmas freeze (global)
      duration: 48h
  environments:
    - branch: environment/staging
      timezone: America/New_York
      allow:
        - cron: "0 6 * * 1-5"   # Weekdays 6 AM–6 PM
          duration: 12h
```

## Test plan

- [x] Integration tests: inside allow window, exclusion-only mode, exclusion overrides allow, environment cleanup, custom key, missing PromotionStrategy
- [x] Unit tests: `evaluateWindows` table-driven tests covering boundaries, multiple windows, exclusion override, frequent-cron churn prevention, error cases
- [x] YAML testdata deserialization
- [x] Full test suite passes (`make test-parallel`)
- [x] All pre-commit checks pass (`build-installer`, `fmt`, `vet`, `lint`, `deadcode`, `build`)


api/v1alpha1/scheduledcommitstatus_types.go
```diff
@@ -0,0 +1,209 @@
...
+	// +required
+	// +kubebuilder:validation:MinLength=9
+	// +kubebuilder:validation:MaxLength=256
+	Cron string `json:"cron"`
+
+	// Duration is how long the window remains open after the cron trigger.
+	// The duration should be in a format accepted by Go's time.ParseDuration function, e.g., "8h", "30m", "2h30m".
+	// +required
+	Duration metav1.Duration `json:"duration"`
+}
```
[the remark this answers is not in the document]
patjlm: Addressed in fc6d17f. CEL validation isn't possible here because `metav1.Duration` is serialized as `type: string` in the CRD schema without `format: duration` (controller-gen doesn't emit it), so CEL can't do numeric comparison on it.

Added runtime validation instead — `evaluateWindows` now rejects non-positive durations with a clear error before any cron evaluation:

```go
if excl.Duration.Duration <= 0 {
    return windowEvalResult{}, fmt.Errorf("exclusion cron %q has non-positive duration %s", excl.Cron, excl.Duration.Duration)
}
```

Same guard on allow windows.

internal/controller/scheduledcommitstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+	found := false
+	for !candidate.After(now) {
+		if candidate.Add(duration).After(now) {
+			latest = candidate
+			found = true
+		}
+		candidate = sched.Next(candidate)
+	}
+	return found, latest
+}
```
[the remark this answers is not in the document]
patjlm: Addressed in fc6d17f. Added two guards:

1. **Loop termination**: `isInsideCronWindow` now breaks when `sched.Next()` returns zero or doesn't advance:
```go
for !candidate.IsZero() && !candidate.After(now) {
    // ...
    next := sched.Next(candidate)
    if !next.After(candidate) {
        break
    }
    candidate = next
}
```

2. **Zero `nextStart` filtering**: all `sched.Next(now)` results are checked before appending to `nextTransitions`:
```go
if nextStart := sched.Next(now); !nextStart.IsZero() {
    nextTransitions = append(nextTransitions, nextStart)
}
```

This prevents both infinite loops on unsatisfiable crons and spurious 1s requeues from zero transition times.

api/v1alpha1/promotionwindowcommitstatus_types.go
```diff
@@ -0,0 +1,209 @@
...
+type PromotionWindowCommitStatusSpec struct {
+	// PromotionStrategyRef is a reference to the promotion strategy that this promotion window commit status applies to.
+	// +required
+	PromotionStrategyRef ObjectReference `json:"promotionStrategyRef"`
+
+	// Key is the gate name referenced in the PromotionStrategy's activeCommitStatuses or
+	// proposedCommitStatuses. When omitted, the CRD default is promotion-window. Set Key explicitly, even if you use the CRD default.
+	// Must be lowercase alphanumeric with hyphens, 1–63 characters (pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$).
+	// +optional
+	// +kubebuilder:default=promotion-window
```
crenshaw-dev: I'd suggest we go ahead and make this required. A couple *CommitStatuses made it optional to avoid a breaking change, but since this is new we can follow the new convention.

api/v1alpha1/scheduledcommitstatus_types.go
```diff
@@ -0,0 +1,209 @@
...
+	return PromotionWindowCommitStatusDefaultKey
+}
+
+// PromotionWindowEnvironment defines the window configuration for a single environment.
+// At least one of windows or exclusions must be specified.
+// +kubebuilder:validation:XValidation:rule="size(self.windows) > 0 || size(self.exclusions) > 0",message="at least one of windows or exclusions must be specified"
+type PromotionWindowEnvironment struct {
+	// Branch is the name of the branch/environment you want to gate with time windows.
+	// +required
+	// +kubebuilder:validation:MinLength=1
```
crenshaw-dev, line 77: Copy additional validations from Branch field in PromotionStrategy.

api/v1alpha1/scheduledcommitstatus_types.go
```diff
@@ -0,0 +1,209 @@
...
+	Branch string `json:"branch"`
+
+	// Timezone is the IANA timezone name used for evaluating cron expressions (e.g. "America/New_York", "Europe/London").
+	// Defaults to UTC.
+	// +optional
+	// +kubebuilder:default="UTC"
+	// +kubebuilder:validation:MaxLength=64
+	Timezone string `json:"timezone,omitempty"`
```
crenshaw-dev: Max currently is 32, but 64 is probably fine to give us breathing room. `curl https://www.geoapify.com/data-share/timezones/timezone-info.json | jq 'map(.tzIdentifier| length) | max'`

If we wanted to get silly, we could have an enum with all 598 possible values.

api/v1alpha1/promotionwindowcommitstatus_types.go
patjlm: Done — made `key` required in 1df46cf. Removed `+optional`, `+kubebuilder:default`, and `omitempty`. Also removed the now-unnecessary `CommitStatusKey()` method and `PromotionWindowCommitStatusDefaultKey` constant since there is no backward compat concern.

api/v1alpha1/scheduledcommitstatus_types.go
patjlm: Done — added the same Branch validations as PromotionStrategy in 1df46cf: `MaxLength=100`, plus CEL rules for leading `-`, `:`, and `..`.

Also added `MaxItems=1000` on environments (matching PromotionStrategy) and `MaxItems=20` on windows/exclusions. The extra `MaxItems` bounds are needed because our Branch field is nested deeper than PromotionStrategy's (environments → items with sibling windows/exclusions arrays), which multiplies the CEL estimated cost and exceeds the CRD cost budget without them.

api/v1alpha1/scheduledcommitstatus_types.go
patjlm: Keeping `MaxLength=64` — gives breathing room over the current max of 32 without going overboard. Skipping the 598-value enum for now 😄

api/v1alpha1/promotionwindowcommitstatus_types.go
```diff
@@ -0,0 +1,203 @@
...
+	// Environments defines the per-environment window configuration. Environments not listed here
+	// default to success (24/7 open) — only listed environments are gated.
+	// +required
+	// +kubebuilder:validation:MaxItems=1000
+	Environments []PromotionWindowEnvironment `json:"environments"`
+}
+
+// PromotionWindowEnvironment defines the window configuration for a single environment.
+// At least one of windows or exclusions must be specified.
+// +kubebuilder:validation:XValidation:rule="size(self.windows) > 0 || size(self.exclusions) > 0",message="at least one of windows or exclusions must be specified"
```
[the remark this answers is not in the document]
patjlm: Not needed — windows and exclusions are array-typed fields. In CRD CEL validation, absent optional arrays evaluate as empty arrays ([]), not errors. So size(self.windows) returns 0 when the field is omitted, no error absorption involved. The has() guard pattern applies to scalar/object optional fields, not arrays with omitempty.

internal/controller/testdata/PromotionWindowCommitStatus.yaml
```diff
@@ -0,0 +1,50 @@
...
+    name: webservice-tier-1
+  environments:
+    - branch: environment/development
+      timezone: America/New_York
+      windows:
+        - cron: "0 6 * * 1-5"
+          duration: 12h
+      exclusions:
+        - cron: "0 0 25 12 *"
+          duration: 48h
```
crenshaw-dev: How about adding a global `spec.windows` and `spec.exclusions`?

internal/controller/testdata/PromotionWindowCommitStatus.yaml
```diff
@@ -0,0 +1,50 @@
...
+  key: promotion-window
+  promotionStrategyRef:
+    name: webservice-tier-1
+  environments:
+    - branch: environment/development
+      timezone: America/New_York
+      windows:
+        - cron: "0 6 * * 1-5"
+          duration: 12h
+      exclusions:
```
crenshaw-dev: Maybe rename `allow`, and `exclude`? `exclusions` are also windows, so `windows` feels a bit ambiguous.

internal/controller/testdata/ScheduledCommitStatus.yaml
```diff
@@ -0,0 +1,50 @@
...
+  namespace: default
+spec:
+  key: promotion-window
+  promotionStrategyRef:
+    name: webservice-tier-1
+  environments:
+    - branch: environment/development
+      timezone: America/New_York
+      windows:
+        - cron: "0 6 * * 1-5"
```
crenshaw-dev: We'll wanna be careful with our cron flavor selection. I know some have pretty nice features that we might not want to lock ourselves out of. Might also wanna check what Argo CD uses for sync windows.

crenshaw-dev (changes requested):

internal/controller/testdata/PromotionWindowCommitStatus.yaml
patjlm: Done in ce3edbd. Added global `spec.allow` and `spec.exclude` fields — merged with per-environment windows at evaluation time (OR semantics for allow, precedence for exclude). Testdata and sample updated to show a global Christmas exclusion applied to all environments.

internal/controller/testdata/PromotionWindowCommitStatus.yaml
patjlm: Done in ce3edbd. Renamed throughout:
- `windows` → `allow`
- `exclusions` → `exclude`
- `activeWindow` → `activeAllow`
- `activeExclusion` → `activeExclude`

internal/controller/testdata/ScheduledCommitStatus.yaml
patjlm: We use `robfig/cron/v3` — same library Argo CD uses for sync windows, with the same 5-field parser config (`Minute | Hour | Dom | Month | Dow`). Argo CD also uses the same design pattern: separate `TimeZone` field + `Duration`.

One note: `robfig/cron/v3` natively supports `CRON_TZ=` prefix in expressions. Since we have a dedicated `timezone` field (matching Argo CD), we should decide whether to strip/reject `CRON_TZ=` in expressions or document that `timezone` takes precedence. Open to discussion on this.

patjlm: Argo CD sync window references for comparison:

- **SyncWindow struct**: [types.go#L2925-L2950](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L2925-L2950)
- **Cron parser** — `robfig/cron/v3` with `Minute | Hour | Dom | Month | Dow`: [types.go#L29](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L29)
- **Single window active check**: [types.go#L3369-L3393](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L3369-L3393)
- **Collection active check**: [types.go#L2958-L2992](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L2958-L2992)

Same library, same 5-field parser config, same pattern (separate TimeZone + Duration). We are feature-equivalent.

api/v1alpha1/promotionwindowcommitstatus_types.go
```diff
@@ -0,0 +1,217 @@
...
+	// Environments defines the per-environment window configuration. Environments not listed here
+	// default to success (24/7 open) — only listed environments are gated.
+	// +required
+	// +kubebuilder:validation:MaxItems=1000
+	Environments []PromotionWindowEnvironment `json:"environments"`
+}
+
+// PromotionWindowEnvironment defines the window configuration for a single environment.
+// At least one of allow or exclude must be specified.
+// +kubebuilder:validation:XValidation:rule="size(self.allow) > 0 || size(self.exclude) > 0",message="at least one of allow or exclude must be specified"
```
zachaller: Does this mean global windows can not be the only configured option, and that we have to have an environment window?

zachaller:
We need to add a docs page like other commit status managers have.

internal/controller/scheduledcommitstatus_controller_test.go
```diff
@@ -0,0 +1,670 @@
+/*
```
zachaller, line 1: mergeWindows is untested. Add cases for:

global exclude + local allow
global allow + empty local (once CEL allows it)
OR across global + local allow

internal/controller/promotionwindowcommitstatus_controller_test.go
```diff
@@ -0,0 +1,670 @@
+/*
+Copyright 2024.
```
zachaller: 2026?

api/v1alpha1/promotionwindowcommitstatus_types.go
```diff
@@ -0,0 +1,217 @@
...
+	// per-environment exclusions. Exclusions take precedence over allow windows.
+	// +optional
+	// +kubebuilder:validation:MaxItems=20
+	Exclude []CronWindow `json:"exclude,omitempty"`
+
+	// Environments defines the per-environment window configuration. Environments not listed here
+	// default to success (24/7 open) — only listed environments are gated.
+	// +required
+	// +kubebuilder:validation:MaxItems=1000
+	Environments []PromotionWindowEnvironment `json:"environments"`
```
zachaller: MinItems=1

internal/controller/promotionwindowcommitstatus_controller.go
```diff
@@ -0,0 +1,468 @@
...
+		return fmt.Errorf("failed to get PromotionWindowCommitStatus rate limiter: %w", err)
+	}
+
+	maxConcurrentReconciles, err := settings.GetMaxConcurrentReconcilesDirect[promoterv1alpha1.PromotionWindowCommitStatusConfiguration](ctx, r.SettingsMgr)
+	if err != nil {
+		return fmt.Errorf("failed to get PromotionWindowCommitStatus max concurrent reconciles: %w", err)
+	}
+
+	err = ctrl.NewControllerManagedBy(mgr).
+		For(&promoterv1alpha1.PromotionWindowCommitStatus{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
zachaller: Let's add a .Named().

zachaller:
Not a thing to change per say but maybe think about the name of the kind I had used ScheduledCommitStatus in a previous PR, PromotionWindowCommitStatus feels a bit off maybe but lets just chat about it before any changes are made.

internal/controller/testdata/ScheduledCommitStatus.yaml
crenshaw-dev: Is there any reason to allow the user to specify different time zones per pattern? I guess maybe you have a pattern for the Europe team's release freeze and a different one for the NA team's freeze. We could just punt TZ and let user's set it in the cron field.

crenshaw-dev:
Yeah I like `Scheduled` or `Schedule`. The word `Promote` finds its way into too many contexts.  :-)

internal/controller/testdata/PromotionWindowCommitStatus.yaml
patjlm: actually, what should the behavior be for global windows? Should they apply to all the environments from promotionstrategy  (my preference) or only to those listed in this commitstatus?

crenshaw-dev: Good question... the most flexible option would be to only apply to listed environments and allow environment list items to not specify any crons if they only want the global crons.

internal/controller/testdata/ScheduledCommitStatus.yaml
patjlm: you mean prefixing with `CRON_TZ=` ?

crenshaw-dev: Yeah... although there's no reason we couldn't add a field for per-window timezones if people wanted that. 

idk i guess I marginally favor a dedicated field. Makes the feature self documenting and gives us the chance to validate early.

patjlm: ok, keeping the field then

internal/controller/testdata/PromotionWindowCommitStatus.yaml
crenshaw-dev: Whatever pattern we pick, we should probably document the standard and try to apply it to all *CommitStatus kinds.

patjlm:
`Schedule` or `Scheduled` ? :smile: 

Is there a risk to be unclear with `TimedCommitStatus` (might be better named `DelayedCommitStatus` actually, in which case `Scheduled` looks nice IMHO) ?

internal/controller/promotionwindowcommitstatus_controller_test.go
patjlm: Fixed in d4f547b — copyright updated to 2026 in all three new files.

internal/controller/promotionwindowcommitstatus_controller.go
patjlm: Done in d4f547b — added `.Named("promotionwindowcommitstatus")`.

api/v1alpha1/promotionwindowcommitstatus_types.go
patjlm: Done in d4f547b — added `MinItems=1` on Environments.

internal/controller/scheduledcommitstatus_controller_test.go
patjlm: Done in d4f547b — added 8 mergeWindows tests covering all requested cases:

- `nil`/`nil` → nil
- global-only → returns global
- local-only → returns local
- global exclude + local allow → concatenated
- global exclude + local exclude → concatenated
- OR across global + local allow → evaluates correctly
- global-only allow (inside/outside) → correct phase
- global-only exclude active → blocks
- global exclude overrides local allow → blocks
- local exclude blocks despite global allow → blocks

api/v1alpha1/promotionwindowcommitstatus_types.go
patjlm: Addressed in d4f547b — per crenshaw-dev's suggestion, global windows apply only to listed environments. The CEL validation moved to spec level so environments can have empty per-env windows when global allow/exclude are defined:

```
self.environments.all(e, size(e.allow) > 0 || size(e.exclude) > 0 || size(self.allow) > 0 || size(self.exclude) > 0)
```

To opt an environment into global-only windows, list it with just a branch (no per-env crons needed).

patjlm:
An other alternative maybe (not saying it's better): `PlannedCommitStatus` ?

internal/controller/testdata/PromotionWindowCommitStatus.yaml
patjlm: Done in 296b240. Two additions:

1. **Dedicated docs page** for PromotionWindowCommitStatus at `docs/gating-promotions/built-in-gates/promotion-window-commit-status.md` — covers cron windows, timezones, global/per-env merge semantics, exclusion-only mode, precedence rules, and troubleshooting.

2. **"Spec Configuration Patterns" section** in `docs/contributing/developing-a-commitstatus.md` — documents the three patterns used across CommitStatus CRDs:
   - Pattern 1: Global-only (GitCommitStatus, ArgoCDCommitStatus, WebRequestCommitStatus)
   - Pattern 2: Per-env only (TimedCommitStatus)
   - Pattern 3: Global + per-env merge (PromotionWindowCommitStatus)

   Includes when to use each, what Pattern 3 requires (documented merge semantics, CEL validation, clear precedence), and a decision table.

Note: once the CRD rename is decided, these docs will need to be updated with the new name as well (filenames, YAML examples, nav entries).

patjlm:
Done in 296b240. Added:

- **Docs page**: `docs/gating-promotions/built-in-gates/promotion-window-commit-status.md` — follows the same structure as the TimedCommitStatus and GitCommitStatus pages (overview, examples, cron/duration/timezone references, precedence rules, status fields, troubleshooting).
- **mkdocs nav entry** and **gating-promotions index entry** added.
- **Contributing guide update**: documented the three spec configuration patterns across CommitStatus CRDs (global-only, per-env-only, global+per-env merge) in `docs/contributing/developing-a-commitstatus.md`.

Note: all doc references will need updating once the CRD rename is decided (`ScheduleCommitStatus` vs `ScheduledCommitStatus` — still pending).

crenshaw-dev:
I think Scheduled, for consistency with Timed

patjlm:
Done in 5097dd6 — renamed `PromotionWindowCommitStatus` → `ScheduledCommitStatus` across the entire codebase: types, controller, tests, applyconfiguration, CRD, RBAC, samples, config, docs, and dist bundles. Docs page is now at `scheduled-commit-status.md`.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
```diff
@@ -0,0 +1,353 @@
...
+
+Global and per-environment windows are **merged** with OR semantics for allow windows (any match allows) and OR semantics for exclude windows (any match blocks). Exclusions always take precedence over allow windows.
+
+An environment with no per-environment windows is valid as long as global windows are defined. The CEL validation rule enforces that each environment has at least one allow or exclude window, either per-environment or global.
+
+## Cron Expressions
+
+The `cron` field accepts standard 5-field cron expressions:
+
+```
```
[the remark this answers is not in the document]
patjlm: Fixed in 618db94 — added `text` language specifier to the cron diagram fenced code block.

internal/controller/scheduledcommitstatus_controller.go
```diff
@@ -0,0 +1,471 @@
...
+			found = true
+		}
+		next := sched.Next(candidate)
+		if !next.After(candidate) {
+			break
+		}
+		candidate = next
+	}
+	return found, latest
+}
```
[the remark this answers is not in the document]
patjlm: Fixed in 618db94 — added `maxIterations = 10000` cap on the backward walk loop. This covers `* * * * *` with ~7-day duration (well beyond practical configs) while preventing pathological inputs from spinning.

internal/controller/testdata/PromotionWindowCommitStatus.yaml
patjlm: About
> Whatever pattern we pick, we should probably document the standard and try to apply it to all *CommitStatus kinds.

Updated the [developing-a-commitstatus](https://github.com/argoproj-labs/gitops-promoter/blob/maintenance-window/docs/contributing/developing-a-commitstatus.md#spec-configuration-patterns) doc with a suggestion: standardize on two patterns:

1. **Global-only** — for CRDs where the same logic applies identically to every environment (ArgoCD, Git, WebRequest)
2. **Global + per-env merge** (standard) — for any CRD where environments need different settings. Global fields provide defaults/shared config, per-env fields override or merge with documented semantics.

This would mean adding a global default `duration` to TimedCommitStatus in a follow-up, so all per-env CRDs follow the same model.

Happy to adjust if you have a different direction in mind.

api/v1alpha1/scheduledcommitstatus_types.go
```diff
@@ -0,0 +1,219 @@
...
+	// +kubebuilder:validation:XValidation:rule="!self.contains(':')",message="branch must not contain ':'"
+	// +kubebuilder:validation:XValidation:rule="!self.contains('..')",message="branch must not contain '..'"
+	Branch string `json:"branch"`
+
+	// Timezone is the IANA timezone name used for evaluating cron expressions (e.g. "America/New_York", "Europe/London").
+	// Defaults to UTC.
+	// +optional
+	// +kubebuilder:default="UTC"
+	// +kubebuilder:validation:MaxLength=64
+	Timezone string `json:"timezone,omitempty"`
```
crenshaw-dev: In lieu of enum validation, what about a regex?

```regex
^[a-zA-Z0-9_/+-]*$
```

docs/contributing/developing-a-commitstatus.md
```diff
@@ -202,6 +202,71 @@ commitStatus.Spec.Phase = promoterv1alpha1.CommitPhaseFailure
...
+```yaml
+spec:
+  key: my-gate
+  promotionStrategyRef:
+    name: my-app
+  # All config here — no environments list in spec
+  expression: '...'
+```
+
+**When to use:** The same logic applies identically to every environment. No per-env overrides are needed.
```
crenshaw-dev, line 241:
```suggestion
**When to use:** The same logic applies identically to every environment. No per-env overrides are needed.

Note that environment-specific logic may be embedded in the global settings. For example, an expression might contain logic to differentiate by branch. But the expression itself applies globally and is not configured per environment.
```

docs/contributing/developing-a-commitstatus.md
```diff
@@ -202,6 +202,71 @@ commitStatus.Spec.Phase = promoterv1alpha1.CommitPhaseFailure
...
+**Planned for:** TimedCommitStatus (currently per-env only; adding a global default `duration` is tracked as a follow-up)
+
+```yaml
+spec:
+  key: promotion-window
+  promotionStrategyRef:
+    name: my-app
+  exclude:                          # global — applies to all listed envs
+    - cron: "0 0 25 12 *"
+      duration: 48h
```
crenshaw-dev: Should we have a `description` field like in Argo CD? Might be helpful when expressing to users what is blocking their promotion.

crenshaw-dev: Or perhaps a `name` field that doubles as a map key? That would make it a required field, but it would also allow more structured construction of these resources with e.g. Kustomize.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
```diff
@@ -0,0 +1,353 @@
...
+2. It merges global windows (from `spec.allow`/`spec.exclude`) with per-environment windows
+3. It evaluates which windows are currently active:
+   - If inside any **exclusion** window → `pending` (blocked), regardless of allow windows
+   - If no allow windows are defined → `success` (exclusion-only mode: open unless excluded)
+   - If inside any **allow** window → `success` (allowed)
+   - If outside all allow windows → `pending` (blocked)
+4. It creates/updates a CommitStatus for the environment's active SHA
+5. It requeues at the next window transition time for precise state changes
+
+Environments **not listed** in the ScheduledCommitStatus are not gated — they default to success (24/7 open).
```
crenshaw-dev: Do we _produce_ CommitStatuses for all environments, or do we read the configured keys on the PromotionStrategy to smartly only produce for the necessary environments? (I'll probably answer my own question as I read, but don't wanna forget.)

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
```diff
@@ -0,0 +1,353 @@
...
+    - key: promotion-window
+  environments:
+    - branch: environment/development
+    - branch: environment/staging
+    - branch: environment/production
+```
+
+In this configuration:
+
+- Promotions from `development` to `staging` are blocked unless `development` is inside an allow window
```
crenshaw-dev, line 146: I could be missing something, but I feel like ScheduledCommitStatus should be _purely_ a proposed commit status, not active.

TimedCommitStatus makes sense as an active check, because it's reactive to what happens on the active branch. A merge happens, and then from that moment we count X amount of time forward before considering the environment "healthy."

But ScheduledCommitStatus doesn't care about what's active. It only answers "am I in the window? if yes, merge." That seems like a proposed commit status.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
```diff
@@ -0,0 +1,353 @@
...
+The ScheduledCommitStatus resource maintains detailed status information:
+
+```yaml
+status:
+  environments:
+    - branch: environment/development
+      sha: abc123def456
+      phase: success
+      activeAllow: "0 9 * * 1-5"
+      nextTransition: "2026-07-08T17:00:00Z"
```
crenshaw-dev: I wonder if we should nest to leave room for expansion, especially since PromotionStrategyDetails opens the door for using more *CommitStatus info for a rich UI.

```yaml
active:
  name: # field I proposed above, maybe isn't a thing
  allow:
  exclude:
  transition:
next:
  name: # again, might not be a thing
  allow:
  exclude:
  transition:
```

It's possible that everything we'd want to show in the UI could instead be simply part of the `description` string. But maybe we end up wanting something like a countdown timer with the name of the next gate or something like that.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
```diff
@@ -0,0 +1,353 @@
...
+
+### Gate Stuck in Pending
+
+If a scheduled gate remains in pending status:
+
+1. Check if the current time is outside all allow windows for the environment
+2. Check if the current time is inside an exclusion window
+3. Verify the timezone is correct — cron expressions are evaluated in the configured timezone
+4. Inspect the status for `activeExclude` and `nextTransition` fields
+5. Check the controller logs for cron parsing errors
```
crenshaw-dev: This should bubble up to the ready condition right?

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
```diff
@@ -0,0 +1,353 @@
...
+
+### Gate Not Created
+
+If no CommitStatus is created:
+
+1. Verify the PromotionStrategy reference is correct (`spec.promotionStrategyRef.name`)
+2. Ensure the environment branch names match exactly between the ScheduledCommitStatus and the PromotionStrategy
```
crenshaw-dev: We should probably have logic that throws an error on a mismatch (although other mistakes could escape such validation, e.g. applying a window to the wrong branch).

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
```diff
@@ -0,0 +1,353 @@
...
+3. Check that the PromotionStrategy has been reconciled and has status populated
```
crenshaw-dev:
```suggestion
3. Check that the PromotionStrategy has been reconciled and has status populated (compare metadata.generation to status.observedGeneration)
```

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
```diff
@@ -0,0 +1,353 @@
...
+4. Verify the environment has an active hydrated commit
```
crenshaw-dev, line 401: Add an example `kubectl get`?

internal/controller/scheduledcommitstatus_controller.go
```diff
@@ -0,0 +1,472 @@
...
+// ScheduledCommitStatusReconciler reconciles a ScheduledCommitStatus object.
+type ScheduledCommitStatusReconciler struct {
+	client.Client
+	Scheme      *runtime.Scheme
+	Recorder    events.EventRecorder
+	SettingsMgr *settings.Manager
+	EnqueueCTP  CTPEnqueueFunc
+}
+
+// +kubebuilder:rbac:groups=promoter.argoproj.io,resources=scheduledcommitstatuses,verbs=get;list;watch;create;update;patch;delete
```
crenshaw-dev:
```suggestion
// +kubebuilder:rbac:groups=promoter.argoproj.io,resources=scheduledcommitstatuses,verbs=get;list;watch
```

I don't think the controller will be C/U/D'ing anything besides status and finalizers.

internal/controller/timedcommitstatus_controller.go
```diff
@@ -279,6 +281,7 @@ func (r *TimedCommitStatusReconciler) calculateCommitStatusPhase(requiredDuratio
 	return promoterv1alpha1.CommitPhasePending, fmt.Sprintf("Waiting for %s duration gate to complete on %s environment", requiredDuration.String(), envBranch)
 }
 
+//nolint:dupl // Gate controllers share the same CommitStatus upsert pattern by design.
```
crenshaw-dev: Time to pull out a shared util function?

crenshaw-dev (changes requested):
The main feedback is about switching from an active to a proposed commit status model.

I only skimmed the controller code, since I think that decision could change the code quite a bit.

api/v1alpha1/scheduledcommitstatus_types.go
patjlm: Done in 619c647 — added `Pattern=^[a-zA-Z0-9_/+-]*$` to the Timezone field.

docs/contributing/developing-a-commitstatus.md
patjlm: Done in 619c647 — applied your suggested paragraph.

docs/contributing/developing-a-commitstatus.md
patjlm: Done in 337dd61 — added required `name` field (list map key for Kustomize overlays) and optional `description`.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
patjlm: Done in 619c647 — clarified in docs: CommitStatuses are only produced for environments listed in the ScheduledCommitStatus, not for all PromotionStrategy environments.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
patjlm: Done in f84eb00 — switched to proposed commit status. Targets `Proposed.Hydrated.Sha`. Docs and tests updated.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
patjlm: Done in 5f234a1 — nested into `active:`/`next:` sub-objects (WindowStatus). Each tracks allow, exclude, transition time, and source cron expression for UI countdown timers.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
patjlm: Done in 619c647 — yes, already works via `HandleReconciliationResult`. Updated doc to reference the Ready condition instead of logs.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
patjlm: Done in 337dd61 — controller validates that each environment branch exists in the PromotionStrategy spec and checks for duplicates. Errors surface via the Ready condition.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
patjlm: Done in 619c647 — applied your suggestion.

docs/gating-promotions/built-in-gates/scheduled-commit-status.md
patjlm: Done in 619c647 — added sample `kubectl get` YAML output.

internal/controller/scheduledcommitstatus_controller.go
patjlm: Done in 619c647 — reduced to `get;list;watch` on the main resource.

internal/controller/timedcommitstatus_controller.go
patjlm: Done in 619c647 — extracted `utils.UpsertCommitStatus` shared across timed/scheduled/git controllers. Removed `nolint:dupl` directives.

internal/controller/testdata/ScheduledCommitStatus.yaml
patjlm: Checked — Argo CD sync windows use the same setup: `robfig/cron/v3` with `cron.Minute|cron.Hour|cron.Dom|cron.Month|cron.Dow` (standard 5-field, no seconds, no descriptors). Their parser lives in `pkg/apis/application/v1alpha1/types.go` as `specParser` — identical flags to our `cronParser`.

Neither Argo CD nor we enable `cron.Descriptor` (`@daily`, `@every`, etc.) or `cron.SecondOptional`. Adding those later is a non-breaking one-line change to the parser flags, so no lock-out risk.

api/v1alpha1/scheduledcommitstatus_types.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+// ScheduledEnvironment defines the window configuration for a single environment.
+type ScheduledEnvironment struct {
+	// Name is a unique identifier for this environment entry, used as the list map key
+	// for Strategic Merge Patch (e.g. Kustomize overlays).
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
```
crenshaw-dev: Actually... since we have a `Description` field, I think this ends up being kinda redundant, and `Branch` should be the list key. Does that make sense? My bad for flip-flopping on this one, didn't really click until I saw it.

api/v1alpha1/scheduledcommitstatus_types.go
```diff
@@ -0,0 +1,250 @@
...
+
+	// Description is an optional human-readable explanation of this environment's window
+	// configuration (e.g. "Business hours deployment window", "EU region maintenance").
+	// +optional
+	// +kubebuilder:validation:MaxLength=256
+	Description string `json:"description,omitempty"`
```
crenshaw-dev: And I think the Description should go into the CronWindow struct so that the user can describe each cron in plan language.

api/v1alpha1/scheduledcommitstatus_types.go
```diff
@@ -0,0 +1,250 @@
...
+	// +kubebuilder:validation:XValidation:rule="!self.contains('..')",message="branch must not contain '..'"
+	Branch string `json:"branch"`
+
+	// Timezone is the IANA timezone name used for evaluating cron expressions (e.g. "America/New_York", "Europe/London").
+	// Defaults to UTC.
+	// +optional
+	// +kubebuilder:default="UTC"
+	// +kubebuilder:validation:MaxLength=64
+	// +kubebuilder:validation:Pattern=^[a-zA-Z0-9_/+-]*$
+	Timezone string `json:"timezone,omitempty"`
```
crenshaw-dev: Should the Timezone instead by at the CronWindow level? Increases duplication but ensures someone could, for example, define a Europe and a US exclusion window easily.

api/v1alpha1/scheduledcommitstatus_types.go
```diff
@@ -0,0 +1,250 @@
...
+	Exclude string `json:"exclude,omitempty"`
+
+	// Transition is when this window state is expected to change
+	// (e.g. window closing for active, window opening for next).
+	// +optional
+	Transition *metav1.Time `json:"transition,omitempty"`
+}
+
+// +kubebuilder:ac:generate=true
+// +kubebuilder:externalDocs:url="https://gitops-promoter.readthedocs.io/en/stable/crd-specs/#scheduledcommitstatus",description="CRD reference (examples and behavior)"
```
crenshaw-dev, line 208: Don't forget to add this section to the crd-specs docs page.

internal/controller/scheduledcommitstatus_controller.go
```diff
@@ -0,0 +1,513 @@
...
+	seenBranches := make(map[string]bool, len(pwcs.Spec.Environments))
+	var duplicateBranches []string
+	psBranches := make(map[string]bool, len(ps.Spec.Environments))
+	for _, env := range ps.Spec.Environments {
+		psBranches[env.Branch] = true
+	}
+	var mismatchedBranches []string
+	for _, envConfig := range pwcs.Spec.Environments {
+		if seenBranches[envConfig.Branch] {
+			duplicateBranches = append(duplicateBranches, envConfig.Branch)
```
crenshaw-dev: Dupe branches should be impossible thanks to CRD validation, right?

internal/controller/scheduledcommitstatus_controller.go
```diff
@@ -0,0 +1,513 @@
...
+			}
+		}
+	}
+
+	if earliest != nil {
+		duration := time.Until(*earliest)
+		if duration <= 0 {
+			return time.Second
+		}
+		return duration + time.Second
```
crenshaw-dev, line 299: Might already be done, but want to make sure: we should probably document that "closing the window" is _not_ a precise operation. We expect the window closure to lag on the order of ~seconds. And if the controller goes down or if there's a reconciliation error, a window might get "stuck" open.

If this becomes a significant issue, we could look at adding a `validUntil` field to the CommitStatus resource. But that's a ways down the line I think.

api/v1alpha1/scheduledcommitstatus_types.go
patjlm: Done — removed `Name` field, `Branch` is now the `+listMapKey`. Sorry for the back-and-forth cost on this one!

api/v1alpha1/scheduledcommitstatus_types.go
patjlm: Done — `description` is now on `CronWindow` so each window can describe itself (e.g. `description: "US East business hours"`).

api/v1alpha1/scheduledcommitstatus_types.go
patjlm: Went with a global+per-window merge pattern: `spec.timezone` sets the default for all windows, `cronWindow.timezone` overrides per-window. UTC if neither is set. This avoids repeating the same timezone on every window while still supporting the Europe+US example you mentioned. Same merge pattern as allow/exclude windows.

api/v1alpha1/scheduledcommitstatus_types.go
patjlm: Done — added `### ScheduledCommitStatus` section to `docs/crd-specs.md` with testdata include.

internal/controller/scheduledcommitstatus_controller.go
patjlm: Correct — removed. `+listMapKey=branch` prevents duplicates at the CRD level, so the controller check was dead code.

internal/controller/scheduledcommitstatus_controller.go
patjlm: Good callout. Added a "Window Closure Precision" section to the docs explaining that closure lags on the order of seconds, and a controller outage can leave a window stuck open until the next successful reconcile. Noted `validUntil` as a potential future enhancement if this becomes a significant issue.

patjlm:
Squashed the 21-commit history down to 3 clean commits:

1. **`feat`**: core ScheduledCommitStatus implementation (types, controller, tests, utils)
2. **`chore`**: all generated files (codegen, CRDs, RBAC, dist bundles) + config scaffolding (kustomization patches, sample CR, controller config defaults)
3. **`docs`**: user guide, CRD specs reference, contributing guide updates

zachaller (approved):

crenshaw-dev (approved):
Brilliant work, thank you!

PR #1773, opened 2026-07-16
feat: support previous-environment commit status details link

jwinters01:
* add new PreviousEnvironmentCommitStatusConfig property to the promotionstrategy spec to allow for consuming a url template for the details link for that commit status
* add support for deeplinking to a particular environment card in the UI (via `env={environment}` url parameter)
* linking to an environment highlights that environment's card in the extension/dashboard UI.

## Summary by CodeRabbit

* **New Features**
  * Added configurable commit statuses summarizing active statuses from the previous environment.
  * Added templated URLs with validation, query escaping, and configurable fallback behavior.
  * Added shareable environment deep links that highlight and scroll to the selected environment.
  * Added keyboard- and mouse-accessible environment card focus states.
* **Documentation**
  * Expanded guidance and examples for configuring previous-environment status URLs and templates.
* **Bug Fixes**
  * Improved internal and external link navigation behavior across commit, pull request, and health-check links.


internal/controller/promotionstrategy_controller.go
```diff
@@ -581,7 +582,44 @@ func (r *PromotionStrategyReconciler) handleRateLimitedEnqueue(
...
+	renderedURL, err := utils.RenderStringTemplate(urlConfig.Template, data, urlConfig.Options...)
+	if err != nil {
+		return "", fmt.Errorf("failed to render previous environment URL template: %w", err)
+	}
+
+	parsedURL, err := neturl.Parse(renderedURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse rendered previous environment URL %q: %w", renderedURL, err)
+	}
+	if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
```
patjlm, line 615: When a conditional template renders to an empty string (e.g. `{{ if .Sha }}https://example.com/{{ .Sha }}{{ end }}` with empty Sha), `url.Parse("")` succeeds with `scheme=""`, the http/https check fails, and the error propagates all the way up to `Reconcile` — causing the controller to requeue indefinitely.

Suggestion: treat an empty rendered string as "no URL" and return `("", nil)` instead of running it through validation. This matches the no-template fallback behavior on line 603.

Related: `Active.Hydrated.Sha` can legitimately be empty when the previous environment has never promoted, so `.Sha` being empty in the template data is a realistic scenario (see the call site at line 775).

ui/components-lib/src/components/Card.tsx
```diff
@@ -57,6 +60,17 @@ const Card: React.FC<CardProps> = ({ environments }) => {
     });
   }, [environments, historyMode]);
 
+  useEffect(() => {
+    if (!highlightBranch) return;
+    const node = highlightRef.current;
+    if (!node) return;
+    node.scrollIntoView({ block: 'nearest', inline: 'nearest', behavior: 'smooth' });
```
patjlm, line 67: `enrichedEnvs` in the useEffect deps changes on every WebSocket/polling refresh (new array reference from the `environments` prop), causing `scrollIntoView` to fire repeatedly. This snaps the viewport back to the highlighted card every few seconds, even after the user has scrolled away.

Suggestion: either remove `enrichedEnvs` from the deps, or use a ref to track that the initial scroll already happened and skip subsequent calls.

internal/controller/promotionstrategy_controller.go
```diff
@@ -581,7 +582,44 @@ func (r *PromotionStrategyReconciler) handleRateLimitedEnqueue(
...
+func resolvePreviousEnvironmentURL(urlConfig promoterv1alpha1.PreviousEnvironmentURLConfig, data PreviousEnvironmentURLTemplateData, previousCRPCSPhases []promoterv1alpha1.ChangeRequestPolicyCommitStatusPhase) (string, error) {
+	if urlConfig.Template == "" {
+		// If there is only one commit status, use the URL from that commit status.
+		if len(previousCRPCSPhases) == 1 {
+			return previousCRPCSPhases[0].Url, nil
+		}
+		return "", nil
+	}
+
+	renderedURL, err := utils.RenderStringTemplate(urlConfig.Template, data, urlConfig.Options...)
```
patjlm, line 606: Nit: this 3-step sequence (RenderStringTemplate → url.Parse → scheme check) and the `PreviousEnvironmentURLConfig` type are structurally identical to `URLConfig` + the validation in `argocdcommitstatus_controller.go:706-719`. Could be worth extracting a shared `utils.RenderAndValidateURL()` to avoid the duplication.

PR #1778, opened 2026-07-17
feat: webhook receiver retry

zachaller:
## Summary by CodeRabbit

* **New Features**
  * Added asynchronous “miss retry” handling for webhook pushes when no matching transfer policy is found on the first hydrated-SHA lookup.
  * Retries use exponential backoff, have a cap on concurrent pending retries, and stop cleanly during shutdown.
  * Added Prometheus gauge `webhook_miss_retry_pending` to expose the number of in-flight pending miss-retries (no labels).
* **Bug Fixes**
  * Improved matching by retrying with a broader hydrated-SHA strategy; ambiguous cases (multiple matches) do not retry.
* **Tests**
  * Expanded webhook receiver tests to cover miss-retry behavior, retry timeouts, retry-limit capacity, and metric/enqueue assertions.
* **Documentation**
  * Documented the new metric and its semantics.


internal/webhookreceiver/server.go
```diff
@@ -34,28 +36,49 @@ const (
...
+		mgr:          mgr,
+		k8sClient:    mgr.GetClient(),
+		enqueueCTP:   enqueueCTP,
+		missRetrySem: make(chan struct{}, maxPendingMissRetries),
 	}
 }
 
 // Start starts the webhook receiver server on the given address.
 func (wr *WebhookReceiver) Start(ctx context.Context, addr string) error {
+	wr.shutdown = ctx.Done()
```
crenshaw-dev: Why store a channel instead of just the context? Then other contexts can just inherit it instead of explicitly handling the shutdown channel.

crenshaw-dev (approved):

PR #1715, opened 2026-07-01
feat(ui): redesign History view

jwinters01:
Redesigns the promotion History view and extracts it into a shared, prop-driven
`HistoryView` component (`ui/components-lib`) so both the dashboard and the ArgoCD
app-view extension can render it.

- **`HistoryView`** — commit-first matrix of promotions across environments. Takes a
  `PromotionStrategy` plus optional `name`/`namespace` and an `onBack` callback (Back
  button renders only when `onBack` is given, so the extension can omit it).
- **Dashboard** — `HistoryPage` becomes a thin wrapper that resolves the strategy and
  wires router-based back navigation.
- **Extension** — gains an Overview / History toggle rendering `<HistoryView>`, sized
  to the viewport via a measured `--hp-offset-top`. This entrypoint is a starting
  point; I plan to iterate on it in a future PR.

## Results
image

## Summary by CodeRabbit

* **New Features**
  * Added a new History view with filters, sorting, environment selection, and a details drawer for selected commits.
  * Added a view toggle in the extension to switch between Overview and History.
  * Improved tooltips and commit/history interactions for clearer context and navigation.

* **Bug Fixes**
  * History navigation and routing now work more reliably, including hosted sub-paths.
  * Updated several labels and loading/empty states for clearer user feedback.

PR #1786, opened 2026-07-20
feat(ui): redesign promotion History detail drawer

jwinters01:
## What this does

Builds on the History view (#1715) to rework its per-commit **detail
drawer**, and makes the checks list ordering deterministic across the UI.

### Detail drawer
- "This commit across environments" section with clickable rows that jump
  the drawer to that environment's cell.
- Rename the "WAS HERE" cell state to "REPLACED" across the cell pill,
  drawer badge, and environments list.
- Show "Replaced on {timestamp}" for replaced cells.
- "Deployed as" hydrated-commit link falls back to the dry commit's repo
  when the CRD leaves the hydrated `repoURL` empty, so the link resolves.
- "Dry commit" / "Hydrated commit" tooltips on the drawer SHA links.
- Route every relative-time entry through the custom tooltip instead of the
  native browser `title` attribute.

### Consistency
- Sort commit statuses by key at ingestion, so the checks list renders in a
  stable order across the overview, history matrix, and drawer.

### Results
image

## Summary by CodeRabbit

* **New Features**
  * Expanded History detail drawer with deployed commit metadata and an “app commits” section showing related source commit details.
  * Added environment-to-environment selection from the drawer for “across environments” entries.
  * Enhanced checks presentation (grouped labels and optional per-check “View details” links).

* **Improvements**
  * Replaced “WAS HERE” with “REPLACED” and updated replacement date messaging.
  * Standardized tooltip behavior across timestamps/author lines (no native title tooltips).

* **Chores**
  * Normalized ordering of commit status entries before they’re displayed.


ui/components-lib/src/components/HistoryView/DetailDrawer/DetailDrawer.tsx
```diff
@@ -198,9 +254,10 @@ const DetailDrawer: React.FC<{
           </div>
         )}
 
-        {cell.commitStatuses.length > 0 && cell.kind !== 'failed' && (
+        {cell.commitStatuses.length > 0 && (
```
zachaller, line 257: **Failing checks lose their details link with this change.**

The removed "Why it failed" section was the only place that rendered a failing check's `c.url`:

```tsx
<a href={c.url} target="_blank" rel="noreferrer" ...>View details</a>
```

The unified Checks list this folds into only renders icon + key + description, so after this PR there's no way to click through from a FAILED cell to the failing check (CI run, ArgoCD app, etc.) — that link is often the whole reason someone opens the drawer on a failed promotion.

Failing checks also sort last here (`[...passingChecks, ...pendingChecks, ...failingChecks]`), so on a failed cell the row you care about ends up at the bottom of the list.

Suggestion: render `c.url` as a link on each check row, e.g.

```tsx
{c.description && <span className="hp-drawer__check-desc">{c.description}</span>}
{c.url && (
  <a
    href={c.url}
    target="_blank"
    rel="noreferrer"
    aria-label={`View details for ${c.key}, opens in new tab`}
  >
    View details
  </a>
)}
```

and consider ordering failing checks first (or at least first when `cell.kind === 'failed'`).

ui/components-lib/src/components/TimeAgo.tsx
```diff
@@ -17,7 +18,11 @@ const TimeAgo: React.FC<TimeAgoProps> = ({ date, interval = 60000 }) => {
   }, [interval]);
 
   const dateString = typeof date === 'string' ? date : date.toISOString();
-  return <span title={formatDate(dateString)}>{timeAgo(dateString)}</span>;
+  return (
+    <Tooltip content={formatDate(dateString)}>
```
zachaller: **This introduces invalid DOM nesting (`<div>` inside `<span>`) at existing `TimeAgo` call sites.**

The components-lib `Tooltip` renders a wrapper `<div className="promoter-popover-wrapper">` around its children. `TimeAgo` is used inside inline text wrapped in `<span>`s — e.g. `CommitInfo.tsx`:

```tsx
<span className="commit-date">authored <TimeAgo date={commit.date} /></span>
```

and `Modal.tsx` (`by <strong>{author}</strong> (authored <TimeAgo ... />)`). With this change those become a `div` inside a `span`, which is invalid HTML. It happens to look right today because the wrapper is `display: inline-flex`, but it's fragile (HTML validators, hydration warnings in stricter setups, and any future CSS that assumes block context).

Two easy fixes:
1. Change the `Tooltip` wrapper element from `div` to `span` (it's already `display: inline-flex`, so nothing changes visually), or
2. Use the portal-based `HistoryView/Tooltip` here instead — it clones the child and adds no wrapper element at all, which also nudges us toward one tooltip implementation instead of two.

ui/components-lib/src/components/HistoryView/DetailDrawer/DetailDrawer.tsx
jwinters01: added details links to the status checks 👍 

image

zachaller (approved):

PR #1804, opened 2026-07-23
fix(ui): dashboard card layout, navigation, and unified history entrypoint

jwinters01:
## What this does

Several fixes to the PromotionStrategy overview UI (shared components, dashboard, extension).

### Stacked view behaves correctly

- You can now scroll to the first and last environment cards when the row overflows.
- The arrow between cards always points in the true promotion direction, even when the layout collapses to a stack without a window resize.
- The "Pushing to Active" indicator reads cleanly as a compact badge instead of overlapping the Active cell.

### Other fixes

- **Query params survive navigation** — params like `?mock=true` are no longer lost when you use the Back button, click a tile, or open History. 
- **Cards fill the available height** — the cards area now stretches to fill the viewport instead of leaving dead space below it, and the card's drop shadow is no longer clipped at the bottom edge. 
- **Consistent way into History** — History is now reached from a tab in the tab bar, the same as the ArgoCD extension, rather than a button buried on each environment card. It still opens the dedicated History page, so the view stays linkable and works with the Back button. 
- Also noticed a race condition that seems to happen in the `tilt` check in CI fairly often, added some changes to attempt to fix that

### Results

**Proposed Changes directional indicator fix:**
Before:
image

After:
image

**Incoming changes indicator fix:**
Before:
image

After:
image

## Summary by CodeRabbit

- **New Features**
  - Added a **History** tab to promotion strategy pages.
- **Improvements**
  - In-app navigation now preserves existing URL query parameters across route changes.
  - Enhanced card layout sizing, spacing, and horizontal scrolling for both orientations.
  - Card orientation updates automatically as the container resizes.
  - Refined proposed promotion banner styling, including a pulsing animation and updated label rendering.
  - Improved full-height flex layout for consistent page structure.
- **Removal**
  - Removed promotion history controls from cards; history is accessed via the tab.


ui/dashboard/src/components/PromotionStrategyDetailsView.tsx
```diff
@@ -10,17 +9,9 @@ interface PromotionStrategyDetailsViewProps {
...
   const environments = strategy.status?.environments || [];
-  const ns = strategy.metadata.namespace || '';
-  const name = strategy.metadata.name || '';
-
-  const handleHistoryNavigate = (_branch: string) => {
-    navigate(`/promotion-strategies/${ns}/${name}/history`);
-  };
 
-  return <Card environments={environments} onHistoryNavigate={handleHistoryNavigate} />;
+  return <Card environments={environments} />;
```
zachaller, line 14: This was the only caller passing `onHistoryNavigate` — the extension renders `<Card environments={...} />` bare in both `AppViewExtension.tsx` and `resourceExtension.tsx`. So after this change the prop is dead code: the History button block in `Card.tsx` (the `onHistoryNavigate && history.length > 0` render, the `FaHistory` import, and the `history`/`environment` lookup feeding it), plus the `.history-toggle` / `.history-controls` styles in `Card.scss`, are all unreachable.

Since this PR is unifying the history entrypoint, can we delete those here too? Otherwise the next reader has to rediscover that this path can never render.

ui/dashboard/src/hooks/useNavigateWithParams.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Like useNavigate, but preserves the current query string when navigating to a
+ * string path that doesn't already carry its own. Keeps params such as ?mock=true
+ * intact across in-app navigation (Back button, tile clicks, history links).
+ */
+export function useNavigateWithParams() {
+  const navigate = useNavigate();
+  const { search } = useLocation();
+
+  return useCallback(
+    (to: string, options?: NavigateOptions) => {
```
zachaller, line 14: Nice hook — two navigation paths still drop the query string though:

- `TopBar.tsx`: the brand `<Link to="/">` 
- `App.tsx`: the root redirect `<Navigate to="/promotion-strategies" replace />`

So clicking the logo from a `?mock=true` session loses mock mode (it goes through both). Worth covering those too, e.g. the redirect could preserve search with `<Navigate to={{ pathname: '/promotion-strategies', search }} replace />` — or note the gap if it's intentionally out of scope.

ui/dashboard/src/components/PromotionStrategyDetailsView.tsx
jwinters01: cleaned up all the dead code 👍

ui/dashboard/src/hooks/useNavigateWithParams.ts
jwinters01: fixed this

zachaller:
@jwinters01 can we fix the conflicts then LGTM

jwinters01:
@zachaller ready when you are

zachaller (approved):

PR #1824, opened 2026-07-26
fix: stop no-op hydration enqueue loop

zzxxyyeeqqww:
## Summary

- stop the 15-second enqueue loop for a terminal no-op hydration only when the proposed dry SHA is already active and equals the current batch target
- preserve enqueue behavior for real output changes, missing or partial metadata, and older environments that still need a newer note propagated
- make existing rate-limit fixtures explicitly exercise an active/proposed difference

## Why

A ChangeTransferPolicy can legitimately reach a state where `active.dry.sha == proposed.dry.sha` while `proposed.note.drySha` differs. Re-enqueuing that current-target no-op every 15 seconds cannot produce a promotable output change and creates continuous controller/Event churn.

The additional `proposed == target` guard is intentional: an older settled environment must still be nudged when another environment establishes a newer batch target.

## Compatibility

No CRD, status-writing, Git note, PullRequest, health, or promotion semantics are changed. This only narrows the internal `enqueueOutOfSyncCTPs` decision.

## Verification

- focused `Rate limiting for enqueueOutOfSyncCTPs` context: PASS
- `make lint`: PASS, 0 issues
- `make fuzz-replay`: PASS
- `go mod tidy`: no diff
- `git diff --check`: PASS

Regression coverage includes terminal no-op mismatch, real diff, missing/empty note, empty proposed SHA, mixed batch, 25 simultaneous no-op CTPs, cross-environment note propagation, repeated owner-watch loops, and the next output-changing hydration.

`go test ./...` was also attempted locally; its controller and e2e packages hit their 10-minute suite timeouts outside the changed focused context. The focused controller context passed consistently and CI can run the repository-prescribed parallel suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved promotion synchronization to skip unnecessary processing when hydration metadata differs but the proposed and active dry states already match the target.
  * Prevented repeated no-op updates from being enqueued.
  * Ensured genuine output differences continue to be processed correctly.

* **Tests**
  * Expanded coverage for hydration states, mixed batches, repeated reconciliation cycles, and subsequent real changes.

### DIFF
--- a/internal/controller/promotionstrategy_controller.go
+++ b/internal/controller/promotionstrategy_controller.go
@@ -468,6 +468,19 @@ func (r *PromotionStrategyReconciler) enqueueOutOfSyncCTPs(ctx context.Context, 
 			continue
 		}
 
+		proposedDrySha := ctp.Status.Proposed.Dry.Sha
+		activeDrySha := ctp.Status.Active.Dry.Sha
+		noteDrySha := ""
+		if ctp.Status.Proposed.Note != nil {
+			noteDrySha = ctp.Status.Proposed.Note.DrySha
+		}
+		// A non-empty note mismatch with the proposed dry SHA already active at the
+		// current batch target is a terminal no-op hydration; there is no promotable
+		// metadata diff to retry.
+		if noteDrySha != "" && proposedDrySha != "" && proposedDrySha == activeDrySha && proposedDrySha == targetSha {
+			continue
+		}
+
 		// Add SHA information to context for logging
 		ctxWithLog := log.IntoContext(ctx, log.FromContext(ctx).WithValues(
 			"effectiveSha", effectiveSha,
--- a/internal/controller/promotionstrategy_controller_test.go
+++ b/internal/controller/promotionstrategy_controller_test.go
@@ -5679,6 +5679,11 @@ var _ = Describe("PromotionStrategy Bug Tests", func() {
 							DrySha: "old123", // Git note SHA (out-of-sync with target)
 						},
 					},
+					Active: promoterv1alpha1.CommitBranchState{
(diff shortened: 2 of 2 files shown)


zzxxyyeeqqww:
The sole failing check (`Run Tilt CI`) appears unrelated to this controller-only change. The failing process is `go run ./cmd dashboard --port=8080`, which times out waiting for `PromotionStrategyDetails.view.promoter.argoproj.io` because the Tilt deployment does not have the view APIService available before dashboard startup. The same cache-sync panic is present on upstream `main` runs [30144162101](https://github.com/argoproj-labs/gitops-promoter/actions/runs/30144162101) and [30140445018](https://github.com/argoproj-labs/gitops-promoter/actions/runs/30140445018). All controller CI, lint, codegen, static analysis, and Kubernetes E2E jobs for this PR passed. I cannot rerun the failed job because GitHub requires repository admin rights.

zachaller:
@zzxxyyeeqqww Thanks for pointing this out, there was actually quite a bit of not great logic I wanted to revise after digging into this I have a semi draft up: https://github.com/argoproj-labs/gitops-promoter/pull/1831

zzxxyyeeqqww:
Thanks for digging into this and for putting together #1831. The effective-SHA target selection and bounded disagreement retries look like the better root fix, especially since they also correct the lagging-environment selection issue that #1824 leaves behind. I’ll follow #1831 and am happy to help validate it against the production no-op hydration case that motivated this PR. Once #1831 is ready and merged, I’ll close this PR as superseded. Thanks again!

zzxxyyeeqqww:
Closing this PR as superseded by #1831. Thanks again for taking the broader redesign forward.

PR #1831, opened 2026-07-27
fix(controller): stop re-enqueueing unchanged CTP note disagreements

zachaller:
When a PromotionStrategy reconciles, it nudges out-of-sync ChangeTransferPolicies to refetch git notes — git note pushes don't trigger webhooks, so this nudge is the only sub-requeue propagation path. The previous logic had two defects:

1. **It looped forever on unresolvable disagreements.** The batch target was the raw `Proposed.Dry.Sha` (from `hydrator.metadata`), while the comparison used the note-preferred effective SHA. After a no-op hydration (the hydrator updates the git note to a newer dry SHA without creating a new commit, so the file never changes), those two can never converge — the same CTP was re-enqueued every 15s indefinitely, refetching a note it already had. This is the churn reported in #1824.
2. **It nudged the wrong environment.** When one environment's note had advanced and a sibling's lagged, the file-based target matched the *laggard* (nothing to do) and mismatched the env that had *already fetched* the note — so the caught-up env was hammered while the env with something to gain was never selected.

This PR replaces that with disagreement-aware retry control:

- **Target selection** — the batch target is now the effective dry SHA (note-preferred, else proposed dry SHA) of the CTP with the newest hydrated commit. A no-op hydration moves the target correctly, so a fully converged batch is silent and only environments whose notes actually lag get nudged.
- **Bounded retries per disagreement** — each distinct `(effectiveSha, targetSha)` pair gets one immediate enqueue plus up to 3 delayed retries at 15s intervals. After the budget is exhausted, nothing further is enqueued until the disagreement changes; the periodic CTP requeue (`changeTransferPolicy.workQueue.requeueDuration`) is the long-tail retry path.
- **Budget reset on change** — a fetched note or a moved target produces a new pair, which resets the counter so the CTP is nudged promptly again.
- **Cleanup fix** — the hourly `enqueueStates` sweep now keys off `lastSeenTime` (touched every time the decision considers a CTP, including skips) instead of `lastEnqueueTime`, so CTPs with exhausted budgets aren't evicted while live and accidentally re-armed every hour. Memory stays bounded: one overwritten pair per CTP, never an accumulating set.

### Design notes

**Why retries exist at all (and why not zero).** A nudge can fire *before* the note lands in git — the pair is then unchanged after the refetch, but the note arrives seconds later. A pure "once per disagreement" version of this change failed three existing specs (`should unblock staging when dev receives only a git note`, `should allow production to promote older commit when staging has moved ahead`, GitCommitStatus `Revert Detection Flow`) for exactly this race. The 3 retries (~45s of coverage) span the window in which a hydrator normally finishes writing notes across environments.

**Why not more retries / exponential backoff.** This mechanism cannot *discover* anything: it only reads statuses, and an unseen note in git is invisible until some reconcile fetches it — so the periodic requeue was always the first-fetch path, before and after this PR. The old 15s-forever loop was a post-fetch echo, not a discovery mechanism. Once a few prompt retries have covered the note-landing race, additional polling adds nothing over the requeue.

**Why exhaustion can't strand a newly relevant note.** The budget is keyed to the pair and the pair includes the target. Any note arrival that matters changes somebody's side of it: the env's own effective SHA (after a fetch) or the target (when the newest env learns the new SHA) — either resets the budget. E.g. with 3 envs settled at `A` and a no-op note `B` landing everywhere: whichever env's requeue fetches `B` first moves the target, which hands the other envs a fresh `(A, B)` pair and an immediate nudge. Exhaustion only ever silences retries for a disagreement that already proved unfetchable.

**Mixed fleets.** An environment whose hydrator doesn't write notes falls back to its file SHA and can never match a sibling's note-derived target — previously a permanent 15s live-lock; now bounded to 4 nudges per target.

### Relationship to #1824

#1824 addresses the same churn with a narrow suppression guard (`note != "" && proposed == active == target`). That fixes the terminal-loop symptom but keeps the file-based target, so the wrong-env asymmetry remains and the guard's correctness depends on subtle case analysis. This PR fixes the target semantics at the root instead, which makes the converged case naturally silent and adds the missing laggard-nudging behavior.

### Verification

- Red/green: with the controller reverted to `main` (tests kept), 5 of the 9 unit specs fail, each at its intended assertion — single-env loop, converged-batch silence, laggard-vs-sibling selection, note-less fallback, and the retry cap (main lands a 5th enqueue where the budget stops at 4).
- Focused run: 14/14 including the three note-propagation flow specs above that killed the retry-free variant.
- Full `make test-parallel`: all suites pass, 0 failures.
- `golangci-lint`: 0 issues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved promotion synchronization by selecting the target from the newest hydrated commit’s effective dry SHA, preferring revision notes when available.
  * Made out-of-sync enqueueing disagreement-scoped per CTP, with bounded, time-spaced retry chaining that only re-arms when the disagreement changes and clears once reconciled.
* **Tests**
  * Reworked enqueue and rate-limiting tests into disagreement-based scenarios, covering lagging-environment selection, agreement/silence behavior, retry budgeting, and independent throttling across CTPs.


internal/controller/promotionstrategy_controller.go
```diff
@@ -438,20 +486,20 @@ func (r *PromotionStrategyReconciler) enqueueOutOfSyncCTPs(ctx context.Context,
...
+	// Find the target SHA - the effective dry SHA from the CTP with the newest proposed
+	// hydrated commit. CTPs whose own effective dry SHA doesn't match this target need to
+	// reconcile to fetch the updated git note. Note a single-environment strategy can
+	// never disagree with itself: its own effective SHA is the target.
 	var targetSha string
 	var newestTime metav1.Time
 	for _, ctp := range ctps {
-		proposedDrySha := ctp.Status.Proposed.Dry.Sha
-		if proposedDrySha == "" {
+		effectiveSha := getEffectiveDrySha(ctp)
```
crenshaw-dev: I'd call it `effectiveProposedDrySha` across the board just to be uber clear.

internal/controller/promotionstrategy_controller.go
```diff
@@ -46,9 +46,36 @@ import (
...
 
-// ctpEnqueueState tracks rate limiting state for enqueuing out-of-sync CTPs.
+// enqueuePair identifies one distinct note disagreement: the CTP's own effective dry SHA
+// versus the batch target it was compared against. A reconcile triggered for a pair
+// refetches git state; if the pair is unchanged afterwards, git held nothing the status
+// hadn't already seen, so re-enqueueing the same pair can never accomplish anything.
+type enqueuePair struct {
+	effectiveSha string
+	targetSha    string
+}
```
crenshaw-dev:
```suggestion
type ctpDisagreement struct {
	// currentEffectiveProposedDrySha is the proposed dry sha on the target CTP.
	currentEffectiveProposedDrySha string
	// newestEffectiveProposedDrySha is the newest proposed dry sha observed on any sibling CTP. 
	// So we expect that requeuing the target CTP should eventually help discover this sha.
	newestEffectiveProposedDrySha    string
}
```

crenshaw-dev (approved):

PR #1923, opened 2026-08-26
perf(git): batch commit metadata reads

zachaller:
## Summary

- Replace per-field `git show` subprocess calls for commit metadata with batched reads and an in-memory per-reconcile cache on `EnvironmentOperations`.
- Commit metadata comes from a single `git log --no-walk=unsorted --stdin -z --pretty=format:...`, so git formats the fields and there is no raw commit-object decoding in Go. `cat-file --batch` is used only for `hydrator.metadata` blobs.
- Prefetch commits and blobs during `ChangeTransferPolicy` reconcile (`calculateStatus` and `calculateHistory`) so metadata readers hit the cache instead of forking git repeatedly.
- Cache parsed trailers on `commitObject` alongside the commit message, so repeated `GetTrailers` calls for the same SHA run `git interpret-trailers` at most once per reconcile.
- Skip `calculateHistory` on steady-state requeues when the active hydrated tip is unchanged and persisted history is already populated.

## Problem

CPU profiling on a steady-state controller showed `runCmdWithEnv` at **0s flat / 1.31s cumulative (33%)** of samples. The cost was fork/exec volume, not Go code inside the function:

- `os/exec.(*Cmd).Start` — 0.95s
- `syscall.forkExec` — 0.56s
- `calculateHistory` — 1.05s cumulative

Each `GetShaMetadataFromGit` call previously ran **four** separate `git show -s --format=...` subprocesses (time, author, subject, body). A single CTP reconcile could spawn **50+** git processes for metadata reads alone (status + history).

## Approach

```mermaid
flowchart TD
    Reconcile[CTP Reconcile] --> CS[calculateStatus]
    CS --> GBS["GetBranchShas x2"]
    GBS --> P1["LoadCommitAndMetadataBlobs: active + proposed"]
    P1 --> SCM[setCommitMetadata from cache]
    Reconcile --> Skip{active tip unchanged?}
    Skip -->|yes, history populated| Done[keep persisted History]
    Skip -->|no| CH[calculateHistory]
    CH --> RL["rev-list --first-parent (5)"]
    RL --> P2["LoadCommitAndMetadataBlobs: 5 commits + blobs"]
    P2 --> TR["GetTrailers (cached commit + trailers)"]
    TR --> P3["LoadCommits: proposed SHAs from trailers"]
    P3 --> HE[buildHistoryEntry from cache]
```

**New in `internal/git/catfile.go`:**

- `commitLogBatch` / `parseCommitLogOutput` — one `git log` per batch, with git formatting the fields we need: `%an`, `%cI`, `%s`, `%b`, `%B`. Records are NUL-separated (`-z`) and fields within each record are also NUL-separated, so parsing is a split plus one `time.Parse`.
- `catFileBatch` / `parseCatFileBatch` — one `git cat-file --batch` per batch for `hydrator.metadata` blobs, reading the length-prefixed stream (header line, then exactly *size* bytes).
- `LoadCommits` / `LoadCommitAndMetadataBlobs` — per-reconcile cache on `EnvironmentOperations`, split into a commit cache and a blob cache.
- `getTrailers` — parses trailers via `git interpret-trailers` once per SHA and stores the result on the cached `commitObject`.

**New in `internal/controller/changetransferpolicy_controller.go`:**

- `shouldSkipHistoryRecalculation` — when `Status.Active.Hydrated.Sha` is unchanged from the pre-reconcile snapshot and `Status.History` is non-empty, skip the entire history git path and keep the persisted history from etcd. Empty history is never skipped so a prior best-effort failure can retry.

**Why `git log` rather than decoding commit objects:** `%s` applies git's own subject folding, `%b` its own subject/body split, and `%cI` emits RFC 3339 with the commit's UTC offset. Letting git do that removes hand-written decoding of the header/message separator, the author and committer ident lines, and the numeric timezone offset — including the `gpgsig` continuation-line handling those needed.

**Prefetch is deliberately non-fatal, which keeps `main`'s behavior for an absent SHA.** No behavior change is intended here; this is the one place where the batching mechanism could have quietly introduced one, so it is worth spelling out.

On `main` there is no prefetch: `GetShaMetadataFromGit` ran four `git show` subprocesses per SHA, so an absent SHA failed only its own read. `populateProposedMetadata` logs that at V(4) and stores the zero value, so the cost is one empty `Proposed.Hydrated` field in one history entry — the rest of `Status.History` is still built and written.

Batching, and with it the possibility of a whole-batch failure, arrives in the first commit of this PR. That commit preserved main's behavior for free: `cat-file --batch` reports `<request> missing` per request and still exits zero, so an absent SHA never failed the batch. `git log` is different — it exits 128 on the first absent SHA and emits nothing for any of them (`--ignore-missing` does not help here). Letting that error propagate would have been a regression against `main`, because `calculateHistory` returns early on prefetch failure: one garbage-collected SHA would drop *all* of `Status.History` rather than blanking one field.

That case is not hypothetical. The active SHAs come from `rev-list` and so exist in the clone by construction, but the proposed SHAs are read out of trailers written by earlier commits, so they routinely point at hydrated commits that were garbage collected or never fetched into this clone. A failed commit prefetch therefore logs at V(4) and leaves the cache unpopulated; each SHA is then resolved individually and only the absent one errors, reproducing main's per-SHA outcome at the cost of one wasted process in the rare bad case.

The blob batch stays fatal on error, since `cat-file` reports missing objects per request and still exits zero — an error there means git itself failed, not that an object is absent. In that case `calculateHistory` returns before assigning `ctp.Status.History`, so the previous history is retained rather than overwritten with empty entries. The `calculateStatus` prefetch likewise returns the error up the reconcile, which is what `setCommitMetadata` already did on `main` for the same class of failure.

**Removed:** `GetShaTime`, `GetShaAuthor`, `GetShaSubject`, `GetShaBody` (no remaining callers).

**Subprocess plumbing:** `runCmdWithEnvAndStdin` is now the single place in the package that runs git; `runCmdWithEnv` delegates to it with a nil stdin, so no existing call site changed.

**Unchanged:** ref/network git operations (`fetch`, `merge-tree`, `rev-parse`, `ls-remote`, `git notes show`).

## Expected impact

| Workload | Before (metadata reads) | After |
|----------|-------------------------|-------|
| `setCommitMetadata` | ~11 subprocesses | cache hits after 1–2 batches |
| `calculateHistory` (5 entries, active moved) | ~31–51 subprocesses | 2–3 batch processes + up to 5 `interpret-trailers` (once per SHA) |
| `calculateHistory` (steady-state requeue) | ~31–51 subprocesses | **skipped** (0 git execs) |
| `BenchmarkCatFilePrefetchedReads` (5 SHAs, steady state) | ~179 ms/op | ~0.4 µs/op |

The prefetched benchmark is a map lookup per read now that a cache hit returns already-formatted fields instead of re-decoding a raw commit object. `BenchmarkCatFileIndividualReads` sits at ~199 ms/op, but that figure is dominated by the fresh `CloneRepo` each iteration performs rather than by the metadata reads themselves.

## Test plan

- [x] Ginkgo unit tests for `parseCommitLogOutput` (multiple records keyed by SHA, committer time offset, empty output, bad field count, unparsable time)
- [x] Ginkgo unit tests for `parseCatFileBatch` (mixed hit/miss, zero-size blob, truncated object, output ending early)
- [x] Integration tests: `GetShaMetadataFromGit` matches `git show` field formatting, including multi-line subject folding and a subject-only commit
- [x] Integration test: a SHA absent from the clone does not fail `LoadCommits`, and the other SHAs still resolve
- [x] Integration test: repeated `GetTrailers` for the same SHA returns cached trailers without re-parsing
- [x] Unit tests: `shouldSkipHistoryRecalculation` (unchanged active tip, active moved, empty history, no active tip)
- [x] `go tool ginkgo -r internal/git/` — 48/48 pass
- [x] `make test-parallel` — full suite green (312 pass, 1 pending)
- [x] `golangci-lint run ./internal/git/...` — 0 issues

## Summary by CodeRabbit

* **Performance**
  * Improved history and status responsiveness in larger repositories through batched commit metadata loading and caching.
  * Reduced repeated Git operations when retrieving commit details and file content.

* **Reliability**
  * Improved handling of missing, malformed, truncated, or incomplete commit data.
  * Added clearer error handling when required commit information cannot be loaded.

* **Bug Fixes**
  * Improved consistency when loading history and status information.
  * Improved support for multi-line commit messages, empty content, varied timestamp formats, and uppercase commit identifiers.


internal/git/catfile.go
```diff
@@ -0,0 +1,319 @@
...
+		return commitObject{}, fmt.Errorf("git log did not return a record for commit %q", sha)
+	}
+	return commit, nil
+}
+
+// getTrailers returns git trailers for a commit, parsing at most once per cached commitObject.
+func (g *EnvironmentOperations) getTrailers(ctx context.Context, sha string) (map[string][]string, error) {
+	key := strings.ToLower(sha)
+
+	commit, err := g.getCommit(ctx, sha)
```
crenshaw-dev:
```suggestion
	commit, err := g.getCommit(ctx, key)
```

I think with this change we could tighten the `fullObjectID` regex to expect lowercase. Not sure if that's valuable. But if it's basically the regex for our cache keys, specific is probably best.

internal/controller/changetransferpolicy_controller.go
```diff
@@ -253,6 +294,29 @@ func (r *ChangeTransferPolicyReconciler) calculateHistory(ctx context.Context, c
 	}
 	logger.V(4).Info("Rev-list history for active branch", "shaList", shaListActive)
 
+	if err := gitOperations.LoadCommitAndMetadataBlobs(ctx, ctp.Spec.ActivePath, shaListActive...); err != nil {
```
crenshaw-dev, line 298:
```suggestion
	// We know which active commits we'll need, so pre-load them.
	if err := gitOperations.LoadCommitAndMetadataBlobs(ctx, ctp.Spec.ActivePath, shaListActive...); err != nil {
```

internal/controller/changetransferpolicy_controller.go
```diff
@@ -253,6 +294,29 @@ func (r *ChangeTransferPolicyReconciler) calculateHistory(ctx context.Context, c
...
+		logger.V(4).Info("failed to prefetch history commit objects", "err", err)
+		return
+	}
+
+	var proposedHistoryShas []string
```
crenshaw-dev, line 304:
```suggestion
	// Each active commit has a corresponding proposed commit. Get those shas so we can preload them.
	var proposedHistoryShas []string
```

crenshaw-dev (approved):

PR #1963, opened 2026-09-02
fix(pr): storm and lock contention on list-installations

crenshaw-dev:
## Summary by CodeRabbit

- **Improvements**
  - GitHub App installation lookups are faster and more efficient through caching.
  - Repeated requests for unknown organizations are temporarily cached, reducing unnecessary API calls.
  - Concurrent requests for the same installation information are consolidated.
  - Configured installation IDs continue to bypass installation discovery.

- **Monitoring**
  - Added metrics for GitHub App installation-listing requests, including provider-only call tracking and duration reporting.
  - Updated metric documentation to clarify provider-scoped calls and the new operation label.


internal/scms/github/git_operations.go
```diff
@@ -8,21 +8,27 @@ import (
...
 )
 
 const (
 	// githubAppPrivateKeySecretKey is the key in the secret that contains the private key for the GitHub App.
 	githubAppPrivateKeySecretKey = "githubAppPrivateKey"
+
+	defaultInstallationMissCacheTTL = 1 * time.Minute
 )
 
+// installationMissCacheTTL is how long a missing org+app installation lookup is remembered without re-listing.
```
pjiang-dev, line 29: Not blocking but could we consider a single cache instead of one positive/one negative cache?

Since a single list call actually returns the complete set of orgs installed for this app, as of time T.
Could we consider a single cache struct?
```
type appInstallations struct {
    byOrg   map[string]int64
    fetched time.Time
}
var appCache map[int64]appInstallations // keyed by appID
```

We would probably need separate TTL for positive hit vs negative hit as we don't want to relist often for positive hits. e.g 30min TTL for positive/ 1 min TTL for negative hit

With this we get added benefit if someone uninstalls the app and re-installs it, it prevents using a stale installationID.

PR #2013, opened 2026-09-10
chore: more PS predicates to reduce reconciles

crenshaw-dev:
I went with duplication instead of DRY just for easy reading.

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced unnecessary status recalculations when promotion strategy updates affect only unrelated metadata, status information, or other irrelevant fields.
  * Reconciliation now responds to relevant branch, repository reference, commit, and commit-time changes.

* **Reliability**
  * Ensured relevant promotion strategy changes continue to update Git, scheduled, and timed commit status results.
  * Preserved processing for newly created and deleted promotion strategies.
  * Added coverage to verify that relevant updates are processed while unrelated changes are ignored.

### DIFF
--- a/internal/controller/gitcommitstatus_controller.go
+++ b/internal/controller/gitcommitstatus_controller.go
@@ -425,8 +425,7 @@ func promotionStrategyGitCommitStatusGateRelevantPredicate() predicate.Predicate
 			if !okOld || !okNew {
 				return true
 			}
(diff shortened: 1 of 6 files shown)


crenshaw-dev:
Not worth pursuing for now — cluster reconcile counts look reasonable, so the extra watch filtering isn't justified.

PR #2054, opened 2026-09-20
UI plugins

jwinters01:


ui/dashboard/package.json
```diff
@@ -12,7 +12,7 @@
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "preview": "vite preview",
-    "build:embed": "npm run build && mkdir -p ../web/static && cp -r dist/* ../web/static/",
+    "build:embed": "npm run build && cp ../plugins/plugin*.js dist/ 2>/dev/null; mkdir -p ../web/static && cp -r dist/* ../web/static/",
```
zachaller, line 15: Because of the `;`, this script exits 0 even when `npm run build` fails. The `mkdir -p ../web/static && cp -r dist/* ../web/static/` part still runs and copies whatever old `dist/` is on disk into `web/static`, so the Go binary embeds out-of-date UI assets and CI stays green.

The `;` is presumably there so a missing `ui/plugins/` doesn't stop the build. That can be handled without breaking the chain:

```suggestion
    "build:embed": "npm run build && (cp ../plugins/plugin*.js dist/ 2>/dev/null || true) && mkdir -p ../web/static && cp -r dist/* ../web/static/",
```

internal/webserver/server.go
```diff
@@ -296,6 +322,143 @@ func (ws *WebServer) StartDashboard(ctx context.Context, addr string) error {
...
+			depth[open]--
+		}
+	}
+	for _, count := range depth {
+		if count != 0 {
+			return false
+		}
+	}
+	return true
+}
```
zachaller, line 361: This check counts brackets without skipping strings, regex literals, template literals or comments. As a result it rejects valid JavaScript that is common in minified bundles. I ran the function as written:

| Input | Result |
|---|---|
| `console.log("(")` | rejected |
| `var re=/[a-z(]+/g;` | rejected |
| ``const s=`${a}}`;`` | rejected |
| `// TODO: handle ) case` ↵ `foo()` | rejected |
| `}{` | **accepted** |

A real webpack bundle only needs one unbalanced bracket inside a string or regex to be dropped. The only trace is an Info-level log line, so the author sees their plugin silently missing from `/plugins.js`. It also accepts some invalid input, like `}{`, so it doesn't reliably catch breakage either.

The syntax check is only needed because the plugins are joined into one response: one `SyntaxError` stops the whole script from parsing, and the per-plugin `try/catch` can't catch it. Some alternatives:

1. **Serve each plugin as its own script** (e.g. `/plugins/<name>.js` plus a small manifest or loader), and have `loadPluginBundle` add one `<script>` per file. The browser then limits a syntax error to that one file, and the Go-side check can go away.
2. **Keep the single bundle but only reject empty files**, and document that a malformed plugin breaks `/plugins.js`.
3. At minimum, log skipped files at Warn/Error level, since a skipped plugin is a real problem for the operator.

The docs claim plugins are syntax-checked on all three paths (see the "Loading a plugin for testing" section in `developing-ui-plugins.md`). That's only true for the extension path, where `concat-plugins.mjs` uses `new Function()`. The docs should be updated whichever option is chosen.

ui/components-lib/src/components/HistoryView/DetailDrawer/DetailDrawer.tsx
```diff
@@ -142,7 +83,9 @@ const DetailDrawer: React.FC<{
   const failingChecks = cell.commitStatuses.filter((s) => s.phase === 'failure');
   const passingChecks = cell.commitStatuses.filter((s) => s.phase === 'success');
   const pendingChecks = cell.commitStatuses.filter((s) => s.phase === 'pending');
-  const checks = getChecks([...failingChecks, ...pendingChecks, ...passingChecks], branch);
+  const checks = getChecks([...failingChecks, ...pendingChecks, ...passingChecks], branch, {
+    promotionStrategy,
+  });
```
zachaller, line 88: Plugins get less context here than on the card view. `getEnvDetails` (card) fills in `promotionStrategy`, `environment`, and all four `*Sha` fields, but this call only passes `promotionStrategy`. In the drawer, `check.environment`, `activeDrySha`, `activeHydratedSha`, `proposedDrySha`, and `proposedHydratedSha` are therefore all `undefined`.

This matters because the drawer is the only place `rowContent` ever renders, and `developing-ui-plugins.md` documents these fields as part of the `Check` contract without a caveat. A plugin that uses, for example, `check.activeHydratedSha` or `check.environment` works in the card view and quietly renders nothing here.

Everything needed is already in scope:

```ts
const environment = promotionStrategy?.status?.environments?.find((e) => e.branch === branch);
const shas = cell.isProposed
  ? { proposedDrySha: cell.commit?.sha, proposedHydratedSha: cell.hydrated?.sha }
  : { activeDrySha: cell.commit?.sha, activeHydratedSha: cell.hydrated?.sha };
const checks = getChecks([...failingChecks, ...pendingChecks, ...passingChecks], branch, {
  promotionStrategy,
  environment,
  ...shas,
});
```

A related point: the meaning of these fields is also inconsistent in `getEnvDetails`:
- The same `checkContext` is passed to both active and proposed checks, so an active check also receives `proposed*Sha`.
- When `index > 0`, `proposedDrySha` is always `undefined`, and `proposedHydratedSha` comes from a variable named `proposedDry` that actually holds the hydrated commit.

It would help to define what each field means ("the commit this check was evaluated against" vs. "the environment's current active/proposed state") in the `Check` type docs. That way both surfaces fill them in the same way and plugin authors know what to expect.

docs/integrating-with-argocd/index.md
```diff
@@ -92,6 +92,79 @@ server:
...
+      initContainers:
+        - name: extension-gitops-promoter
+          image: quay.io/argoprojlabs/argocd-extension-installer:v0.0.9@sha256:d2b43c18ac1401f579f6d27878f45e253d1e3f30287471ae74e6a4315ceb0611
+          env:
+            - name: EXTENSION_NAME
+              value: gitops-promoter
+            - name: EXTENSION_URL
+              value: https://github.com/argoproj-labs/gitops-promoter/releases/download/v0.38.1/gitops-promoter-argocd-extension.tar.gz
+            - name: EXTENSION_CHECKSUM_URL
+              value: https://github.com/argoproj-labs/gitops-promoter/releases/download/v0.38.1/gitops-promoter_0.38.1_checksums.txt
```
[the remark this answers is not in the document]
jwinters01: we will have to fix this in the future since this is the PR introducing this API.

ui/shared/src/components/plugins/reactSharing.ts
```diff
@@ -0,0 +1,32 @@
...
+ * A plugin built with `externals: {react: 'React'}` reads `window.React` as
+ * its module-scope code executes when the `<script>` tag loads. If that global
+ * is missing, or a different object than the host's own React, the plugin ends
+ * up running hooks against a different React internals than the host's
+ * component tree uses — the classic "invalid hook call" failure, which gives
+ * no indication that a duplicate React copy is the actual cause. This check
+ * turns that into a clear, immediate error at load time instead.
+ */
+export function assertSharedReactInstance(hostReact: unknown): void {
+  const globalReact = typeof window === 'undefined' ? undefined : window.React;
```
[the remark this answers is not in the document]
jwinters01: `window.React` already type-checks — `@types/react` ships `export as namespace React;`, a UMD global export, which TypeScript resolves through `window.React` automatically. `tsc --noEmit` passes clean on this file today, and the existing `window.React = hostReact` assignments in `reactSharing.test.ts` already compile. No augmentation needed.

PR #2056, opened 2026-09-21
fix(controller): release old Secret finalizer when ScmProvider secretRef changes

### LINKED ISSUE #1766: ScmProvider secret finalizer not removed when secretRef changes
## Problem

When an `ScmProvider`'s `spec.secretRef` is updated to point at a different `Secret`, the controller adds a finalizer to the new secret but does not remove the finalizer from the previously referenced secret.

The orphaned secret can no longer be deleted and may block namespace termination.

## How we hit it

During manual SCM testing we switched providers by re-applying manifests (`apply-provider.sh gitlab` → `apply-provider.sh gitea`) without deleting the namespace. The old GitLab secret (`scm-labels-gitlab-secret`) remained with `scmprovider.promoter.argoproj.io/secret-finalizer` after `ScmProvider/scm-labels-scm` was updated to reference the Gitea secret. Namespace deletion then stuck on that secret.

Secret finalizer removal currently runs only when the `ScmProvider` is deleted, using the current `spec.secretRef` — not when `secretRef` changes on a live provider.

`ClusterScmProvider` appears to follow the same pattern.

somaz94:
When a ScmProvider or ClusterScmProvider changes spec.secretRef, the controller adds its
secret finalizer to the new Secret but never removes it from the old one. The old Secret
can then no longer be deleted and can block namespace termination.

Each reconcile now also lists the Secrets in the provider's namespace (the controller
namespace for ClusterScmProvider) and removes the provider secret finalizer from any Secret
that no provider of that kind still references. Removal reuses removeSecretFinalizerForProvider.
The API is unchanged, and Secrets orphaned before this fix get cleaned up too. Secrets are
already in the informer cache, so the list does not hit the API server.

I added an envtest case for each kind that switches secretRef and checks the finalizer moves.
Both fail on main and pass with this change.

Validation: make lint, make test-parallel, make fuzz-replay and make nilaway-no-test all pass locally.

Fixes #1766

This change was written with help from Claude Code. I reviewed and tested it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Secret cleanup finalizers are now automatically removed when a provider changes its referenced Secret.
  * Finalizers remain on Secrets still referenced by other providers, preventing premature cleanup.
  * Cleanup continues for other eligible Secrets even if removing one finalizer fails.
  * Updated behavior is covered for both cluster-scoped and namespace-scoped providers.


crenshaw-dev:
Thanks, @somaz94! Ready for review, or still needs more work?

somaz94:
Fixed the CodeRabbit point in c4d115a6, so one Secret that can't be updated no longer stops the rest of the cleanup. Ready for review now.

crenshaw-dev (approved):
Thanks, @somaz94!

PR #2061, opened 2026-09-21
test: limit coverage to internal and webrequestsimulator

crenshaw-dev:
Stop instrumenting api, applyconfiguration, cmd, and other module packages in every Ginkgo binary. Codecov should still reflect the packages the suite actually exercises.


crenshaw-dev:
Closing. A cold-GOCACHE `ginkgo build` A/B of the same suites showed no compile-time difference between `-coverpkg` on the whole module vs `internal` + `webrequestsimulator` (~1m46 vs ~1m45). Not worth the Makefile change for speed; Codecov filtering can stay a separate discussion if we still want it.

